PR #12, opened 2015-02-14
Exec /bin/bash by default

CoRfr:
Hi,

Pushing few patches concerning mainly two things:
- There are still some references to old Docker repositories (ulexus/ or ceph/ceph-), that I replaced. I kept the ones for rbd in ulexus/ since they don't exist (yet ?) on ceph/.
- I added an helper script in ceph/base to be able to assist in the execution of this image directly (which can be conveniant to use ceph directly).

If you intend to create ceph/rados ceph/rbd ceph/rbd-* , please create a patch with:
grep -Rl "ulexus/" |grep -v ".git" |xargs sed -i 's^ulexus/^ceph/^g'

Thanks,
Bertrand

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -12,7 +12,11 @@ RUN apt-get update
 RUN apt-get install -y wget
 
 # Install Ceph
-CMD wget -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc' | apt-key add -
+RUN wget -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc' | apt-key add -
 RUN echo deb http://ceph.com/debian-$CEPH_VERSION/ trusty main | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list
 RUN apt-get update
 RUN apt-get install -y --force-yes ceph
+
+ADD entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+
--- a/base/README.md
+++ b/base/README.md
@@ -9,5 +9,6 @@ https://registry.hub.docker.com/u/ceph/base/
 ## Usage (example)
 
 ```bash
-$ docker run -i -t ceph/base /bin/bash
+$ docker run -i -t ceph/base
 ```
+
--- a/base/entrypoint.sh
+++ b/base/entrypoint.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+if [ -z "$1" ]; then
+  /bin/bash
+else
+  $@
+fi


leseb:
@Ulexus LGTM, what do you think?

Ulexus:
@leseb It looks good to me, yes.  I'll also kick off a rebuild on Docker (still no way for me to setup the autobuild... I have insufficient access to the Github account for that, whatever that implies)

@CoRfr Just rebase, and it should be good.  Sorry for the delay... not sure why I didn't see a notification about this PR.

Ulexus:
@CoRfr And thanks!

CoRfr:
@leseb @Ulexus I updated my branch to remove all fix references stuff, as this has already been done (and now ceph/ is an automated repo, awesome :) ).

PS: name is inconsistent for ceph/rbd (CEPH instead of Ceph)

Ulexus:
Thanks; it looks good.  I'm curious why you explicitly name `bash` in the `apt-get install` statement, though.  I'm pretty sure it's in the base image already.

Capitalization fixed, mostly.  Full consistency would have be name them RBD and RADOS instead of rbd and rados... but since they are referring to the cli tools, I think it would be inappropriate.  I guess I could change MON, OSD, MDS, etc., but I'm going to stop here, for now.

base/README.md
```diff
@@ -9,5 +9,6 @@ https://registry.hub.docker.com/u/ceph/base/
 ## Usage (example)
 
 ```bash
-$ docker run -i -t ceph/base /bin/bash
+$ docker run --rm -ti ceph/base
```
Ulexus: I don't think we should default to `--rm` in the example for base.  If people are using base directly, they are probably trying to construct their own images for use.

base/Dockerfile
```diff
@@ -8,11 +8,13 @@ MAINTAINER Sébastien Han "seb@redhat.com"
...
 
 # Install Ceph
-CMD wget -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc' | apt-key add -
+RUN wget -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc' | apt-key add -
 RUN echo deb http://ceph.com/debian-$CEPH_VERSION/ trusty main | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list
-RUN apt-get update
-RUN apt-get install -y --force-yes ceph
+RUN apt-get update && apt-get install -y --force-yes ceph
+
+COPY ./docker-entrypoint.sh /docker-entrypoint.sh
```
Ulexus: I would prefer to keep the entrypoint script names the same, for consistency.  Otherwise, the downstream images will end up having multiple entrypoint scripts, which becomes confusing.

TLDR; rename to `/entrypoint.sh`

CoRfr:
1/ For bash, I though it was not in the ubuntu:14.04 for some reason, while it is (my mistake), so I'm removing that part

2/ For --rm, I use the base image a bit differently, using --volumes-from, as a "base" to communicate with the ceph system, which is why I don't want leftovers. 
Something like: ```docker run --rm -ti --volumes-from=ceph-data ceph/base ceph osd ls```
Though, I understand what you mean and removed the --rm.

3/ I renamed to entrypoint.sh for consistency as you asked.
I guess it would be nice to also have rados images in sync (entrypoint as /usr/local/bin/entrypoint.sh ), and I'm not sure the chmod step is necessary (expect for radosgw since the entrypoint.sh has no +x).

Ulexus:
@CoRfr I'm pretty sure all the others are now in sync (`/entrypoint.sh`).  Let me know if I missed one, though.

Ulexus:
The `entrypoint.sh` file was changed, but the `Dockerfile` still needs to be updated with the new path.

CoRfr:
@Ulexus I'm not sure I understand. There are 2 different path for "entrypoint.sh", one is "/entrypoint.sh" (rbd-unlock, rbd-lock, rbd-volume) and one is "/usr/local/bin/entrypoint.sh" (mon, osd, mds, radosgw, rbd-volume).

Since "core" images where using "/usr/local/bin/entrypoint.sh", I went for that option, should it be "/entrypoint.sh" ?

CoRfr:
My bad, I missed a rebase, I will update to /entrypoint.sh right away.

Ulexus:
Thanks @CoRfr

PR #28, opened 2015-03-01
Sync cluster bootstrap using etcd

BugRoger:
This container allows to automatically bootstrap a cluster or distribute existing ceph config to new hosts. It makes use of etcd to store the inital configuration. It is especially cool when run on CoreOS and bootstrapping a cluster from nothing without manual intervention. It is compatible to the existing configuration mechanisms in the `ceph-mon` containers. 

If this fits in the vision of `ceph-docker` I would be happy to keep on helping to maintain this config mechanism.

I'm using this on my own ceph cluster as can be seen here: https://github.com/BugRoger/d26a/tree/master/units/ceph

### DIFF
--- a/config/Dockerfile
+++ b/config/Dockerfile
@@ -1,0 +1,29 @@
+# DOCKER-VERSION 1.0.0
+# 
+# Ceph Config 
+#
+#  USAGE NOTES:
+#    - Define at least the two following environment variables:
+#      MONHOST - the hostname of this monitor
+#      MONIP   - the (externally visible) IP address of this monitor
+#    - /etc/ceph is set as a volume, so you may use a common configuration directory
+#      among your ceph daemons (this also keeps private keys outside of the image)
+#
+# VERSION 0.0.1
+
+FROM ceph/base
+MAINTAINER Sébastien Han "seb@redhat.com"
+
+RUN wget https://github.com/coreos/etcd/releases/download/v0.4.6/etcd-v0.4.6-linux-amd64.tar.gz
+RUN tar xzvf etcd-v0.4.6-linux-amd64.tar.gz
+RUN cp etcd-v0.4.6-linux-amd64/etcdctl /usr/local/bin
+
+# Add bootstrap script
+ADD entrypoint.sh /entrypoint.sh
+
+# Add volumes for ceph config and monitor data
+VOLUME ["/etc/ceph"]
+
+# Execute monitor as the entrypoint
+ENTRYPOINT ["/entrypoint.sh"]
+
--- a/config/README.md
+++ b/config/README.md
@@ -1,0 +1,47 @@
+ceph-config
+===========
+
+This Dockerfile may be used to bootstrap a cluster or add the cluster configuration
+to a new host. It uses etcd to store the cluster config. It is especially suitable
+to setup ceph on CoreOS.
+
+The following strategy is applied:
+
+  * If a cluster configuration is available, it will be written to `/etc/ceph`
+  * If no cluster configuration is available, it will be created. A lock mechanism 
+    is used to allow concurrent deployment of multiple hosts.
+
+## Usage 
+
+To bootstrap a new cluster run:
+
+`docker run -e MON_IP=192.168.101.50 -e MON_NAME=mymon -e CLUSTER=testing -v /etc/ceph:/etc/ceph ceph/config`
+
+This will generate:
+
+  *  `ceph.conf` 
+  *  `ceph.client.admin.keyring` 
+  *  `ceph.mon.keyring` 
+  *  `monmap` 
+
+Except the `monmap` the config will be stored in etcd under `/ceph-config/${CLUSTER}`. 
+
+In case a configuration for the cluster is found, the configuration will be pulled
+from etcd and written to `/etc/ceph`.
+
+Multiple concurrent invocations will block until the first host finished to generate 
+the configuration.
+
+When run without `MON_IP` and `MON_NAME` it will not generate a config but block and 
+wait until it becomes available. 
+
+## Configuration
+
+The following environment variables can be used to configure the bootstrapping:
+
+  * `CLUSTER` is the name of the ceph cluster (defaults to: "ceph") 
+
+Mandatory Configuration:
+  * `MON_NAME` is the name of the monitor. Usually the short hostname
+  * `MON_IP` is the IP address of the monitor (public)
+  * `ETCDCTL_PEERS` is a comma seperated list of etcd peers (e.g. http://192.168.2.4:4001)
--- a/config/README.md
+++ b/config/README.md
@@ -7,15 +7,16 @@ to setup ceph on CoreOS.
 
 The following strategy is applied:
 
+  * An `/etc/ceph/ceph.conf` is found, do nothing.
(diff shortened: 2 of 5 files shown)


Ulexus:
@BugRoger I quite like this idea, and it is looks very good to me.  My only concern (and this as a CoreOS user, myself) is its dependency on `etcd`.  For my part, I'd just add a note to the README indicating this dependency, but we should probably get @leseb 's opinion, on the matter.

leseb:
@BugRoger Thanks a lot for pushing this. Actually that's part of the vision I have for ceph-docker.
I believe this is the big missing things in this repo, having an distributed entity to store config and keys is nice to have.
This was on my todo list :). Glad to see that this is here.

@Ulexus I understand your concern about the usage of etcd and it's fair. However I believe it has a sort of consensus now. Many big projects like Kubernetes are using it (Red Hat Atomic too).
So yes, somehow this is tied to CoreOS but nothing prevents you to use it.

I'd be really happy to see this in ceph-docker. I'll review that patch soon.

config/entrypoint.sh
```diff
@@ -0,0 +1,67 @@
...
+ 
+CLUSTER=${CLUSTER:-ceph}
+CLUSTER_PATH=/ceph-config/$CLUSTER
+
+if [ -e /etc/ceph/ceph.conf ]; then
+  echo "Found existing config. Done."
+  exit 0
+fi
+ 
+# Aquire lock to not run into race conditions with parallel bootstraps
```
leseb: s/Aquire/Acquire/

config/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Configuration
+
+The following environment variables can be used to configure the bootstrapping:
+
+  * `CLUSTER` is the name of the ceph cluster (defaults to: "ceph") 
+
+Mandatory Configuration:
+  * `MON_NAME` is the name of the monitor. Usually the short hostname
+  * `MON_IP` is the IP address of the monitor (public)
+  * `ETCDCTL_PEERS` is a comma seperated list of etcd peers (e.g. http://192.168.2.4:4001)
```
leseb: Would it be clearer to simply use `ETCD_PEERS` since we refer to an etcd host?

config/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+#      MONIP   - the (externally visible) IP address of this monitor
+#    - /etc/ceph is set as a volume, so you may use a common configuration directory
+#      among your ceph daemons (this also keeps private keys outside of the image)
+#
+# VERSION 0.0.1
+
+FROM ceph/base
+MAINTAINER Sébastien Han "seb@redhat.com"
+
+RUN wget https://github.com/coreos/etcd/releases/download/v0.4.6/etcd-v0.4.6-linux-amd64.tar.gz
```
leseb: Any particular reason to use this version?
Or maybe etcdctl didn't change since this version?

config/ceph-config.service
```diff
@@ -0,0 +1,35 @@
+[Unit]
```
leseb: It looks like we are not using this file here, is it only for Fleet?

leseb:
Just reviewed that, the overall looks really good, I just left a couple of questions.

config/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN tar xzvf etcd-v0.4.6-linux-amd64.tar.gz
```
leseb: @Ulexus what do you think of adding etcdctl to ceph/base?

Ulexus:
I'm pretty sure etcdctl 2 is compatible with etcd 0.4, so we should probably use it instead.  CoreOS is still shipping with 0.4.7, though, so we should probably verify that (I'll do so when I get in to the office in a couple hours)

@leseb I think including etcdctl in base is a great idea.  That would allow me to rewrite osd, Mon, etc, to (optionally) use etcd, as well.  This feature was requested on my old repo, anyway.

Ulexus:
Verified that etcdctl v2.0.4 works just fine with etcd v0.4.6 (current CoreOS beta, 584, and stable, 557.2).  I don't have an alpha box running v0.4.7, but I can't imagine it _not_ working with v0.4.7 if it works with v0.4.6.  v0.4.7 is just a staging update for migration to 2.x.

config/ceph-config.service
Ulexus: This is consistent with the other components.  It's just an example systemd unit file.  I think they _should_ be included, as they serve as good, practical documentation.

It's not useful to include them in the Docker images, of course, but as a part of the Github repo, I think they are invaluable.

config/Dockerfile
Ulexus: This is the version that CoreOS beta and stable are using, presently, which I imagine is the reason.  That said, etcdctl 2.0.x is compatible with etcd 0.4.x, so I say we should use etcdctl 2.0.x.

config/Dockerfile
Ulexus: Absolutely

config/README.md
Ulexus: `ETCDCTL_PEERS` is actually the "official" environment variable for `etcdctl`

Ulexus: That said, it might not be a bad idea to set `ETCDCTL_PEERS` equal to `ETCD_PEERS` if the former is undefined.  I know I have confused the two in the past.

leseb: Yes I think the name doesn't really match what it does.

config/ceph-config.service
leseb: I'm not discussing if they are useful or not, I strongly believer they are.
However since they are directly connected to the subject of this commit I was thinking that another commit might be more appropriate.
But I'm just being fussy here, so we can merge them with this PR.

leseb:
@Ulexus Thanks for the quick tests :)

config/ceph-config.service
Ulexus: Sorry, I think I misread your intent.  It doesn't strike me as necessary to separate the commits:  to me, they seem to be part of the same thought.

Ulexus:
I don't see a reason not to commit this PR then update base with etcdctl v2.0.4, then update this component.  Objections?

Ulexus:
Well, except for the misspelling you found, anyway. :)

config/ceph-config.service
leseb: Np we can go like this.

leseb:
No objections unless @BugRoger wants to take care of that.
Since he didn't response to our comments, I believe we should merge this and rework what needs to be reworked :).

Ulexus:
PR #29 updated etcdctl and moves it to ceph/base

hunter:
A little late to the discussion, we'd started work on something similar using Consul for the KV. Any sense in using confd to make things a little more universal?

leseb:
@hunter indeed confd will probably make the repo more universal.
If you have something really with confd I'd be happy to have a look at it.
Just created #34 so we can continue the discussion.

BugRoger:
Thanks for considering this. And as I see you already fixed the initial comments. You folks are quick. :smile:

Ulexus:
@BugRoger Thanks for this; I just used it to rebuild my broken cluster.  We should definitely integrate the other components to use it effectively.

Issue #31, opened 2015-03-03
Create new images with tags for ceph version


leseb:
We currently ship firefly as latest tag, we should move to giant and add a tag for firefly.

Next steps:
- [x] Create `firefly` branch in Github
- [x] Create `firefly` tag referencing that branch on Docker Hub
- [x] Update each `FROM` statements in `Dockerfile`s for downstream components in the `firefly` branch to reference `ceph/base:firefly`
- [x] Create `firefly` tags for each downstream component on Docker Hub
- [x] Create `giant` tags for each component on Docker Hub

Ulexus:
I've got to run for a few hours; if it's not done before 20:00 UTC, I'll complete the last two steps here.

leseb:
I don't think I'll have time to do this today.

Ulexus:
No problem; I just wanted to make sure we didn't both work on it, since this was assigned to you at the time.

Ulexus:
Giant tags on docker track the master branch, for now.

PR #32, opened 2015-03-03
Use the latest stable version of Ceph

leseb:
Moving away from Firefly and use Giant instead.

Fixes: #31 

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -5,19 +5,18 @@
 FROM ubuntu:14.04
 MAINTAINER Sébastien Han "seb@redhat.com"
 
-ENV CEPH_VERSION firefly
 ENV ETCDCTL_VERSION v2.0.4
 ENV ETCDCTL_ARCH linux-amd64
+ENV CEPH_VERSION giant
 
 # Install prerequisites
-RUN apt-get update
-RUN apt-get install -y wget
+RUN apt-get update &&  apt-get install -y wget
 
 # Install Ceph
 RUN wget -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc' | apt-key add -
 RUN echo deb http://ceph.com/debian-$CEPH_VERSION/ trusty main | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list
-RUN apt-get update
-RUN apt-get install -y --force-yes ceph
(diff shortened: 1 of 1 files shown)


leseb:
@Ulexus Not sure how we can properly create a tag for firefly...

Ulexus:
We now have Docker tags for `firefly` and `giant` (as per #31 ), and the modifications to `ceph/base` seem to work as expected for the `giant` release.

My only question is with respect to the upgrade.  Is it safe for people using the `latest` tag to simply pull fresh containers and run them, or does there need to be additional checking or instructions involved for the upgrade.  

Since I'm one of the users of the `latest` tag, I'd really hate to destroy my ceph cluster by merging this PR. ;)

leseb:
We should probably add a warning, just to remember users to always use tags (and ceph version) while pulling an image form the hub.
For the upgrade you might to follow a sequence.

Ulexus:
Is there any way to query the versions of the other daemons in the Ceph cluster before running the component?  That is, can we have an OSD container check that the MONs are upgraded before running itself?  Is that even a good or necessary action?

The order appears to be fairly simple for the actual upgrade, though:
 1.  Upgrade each monitor
 2.  Upgrade each OSD
 3.  Upgrade each other daemon (MDS, RADOSGW)
 4.  Upgrade the utilities (rbd,rados,etc.)

Ulexus:
The warning is good.  The only problem is that we _didn't_ have tagging before yesterday, so everyone was necessarily using `:latest`.

The plus side there is that according to Docker Hub, there have only been a couple hundred downloads of `ceph/osd`, and fewer still `ceph/mon`.  For all I know, that could just be me (though that is probably not the case).  In any case, there should not be very many clusters affected, if out-of-order upgrading causes issues.  Of course, to those few...  It would be nice if we had a way to issue email notices to container users, but that's not going to happen, I'm afraid.

leseb:
To check the mon version you can:

```ceph daemon mon.$(hostname) version```

However this needs to be run inside the container since this talks to the socket.
For OSDs you can:

```ceph osd metadata $osd_id```

leseb:
Agree on the few users that might be affected, I don't see any good way to email current users.
I believe we will have to merge that like this :(

Ulexus:
In that case, I guess we merge away

PR #39, opened 2015-03-04
docker-registry

lorieri:
rados-backed docker-registry

### DIFF
--- a/docker-registry/README.md
+++ b/docker-registry/README.md
@@ -1,0 +1,27 @@
+Docker-registry
+===============
+
+```
+docker run -d \
+         --name ceph-docker-registry
+         -e AWS_BUCKET=mybucket \
+         -e AWS_KEY=myawskey \
+         -e AWS_SECRET=myawssecret \
+         -e AWS_HOST=myowns3.com \
+         -e AWS_SECURE=true \
+         -e AWS_PORT=80 \
+         -e AWS_CALLING_FORMAT=boto.s3.connection.OrdinaryCallingFormat \
+         -p 5000:5000 \
+         ceph/docker-registry
+```
+
+In the example boto.s3.connection.OrdinaryCallingFormat  makes API calls in the format:
+```
+http://HOST:PORT/BUCKET/OBJECT
+```
+if AWS_CALLING_FORMAT is empty, it calls like:
+```
+http://BUCKET.HOST:PORT/OBJECT
+```
+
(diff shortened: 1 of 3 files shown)


Ulexus:
@lorieri Is this intended to be a request to create this container based on the README, or do you have a Dockerfile already somewhere?

lorieri:
Sébastien Han asked me to integrate my efforts on docker-registry here.
Since those efforts were merged by docker-registry I create only the readme with instructions to use the official image.

Ulexus:
Ahh, okay (and I want trying to blow you off).  So, you are saying the
official docker registry now supports rados natively?

lorieri:
Don't worry, part of the job :)
And yes, it does :)

Ulexus:
Excellent, and thanks.  I'll try to throw a container together for it next week, if someone else doesn't get to it first.

I'll go ahead and merge this, as it won't do any harm; we'll just hold off adding it to the Docker Registry until after it has a container.  If you don't mind, though, just flag the listing in the top-level README with `WIP` or something like that.

lorieri:
If you want I can make a Dockerfile importing the official image and setting some default values

lorieri:
did it

docker-registry/README.md
```diff
@@ -0,0 +1,27 @@
...
+```
+docker run -d \
+         --name ceph-docker-registry
+         -e AWS_BUCKET=mybucket \
+         -e AWS_KEY=myawskey \
+         -e AWS_SECRET=myawssecret \
+         -e AWS_HOST=myowns3.com \
+         -e AWS_SECURE=true \
+         -e AWS_PORT=80 \
+         -e AWS_CALLING_FORMAT=boto.s3.connection.OrdinaryCallingFormat \
```
Ulexus: Is this static for ceph/radosgw?  If so, can we move it into the Dockerfile?

lorieri: default is empty, most cases people configure the wildcard DNS
I put it only because it is worth mention, specially for automations not dependent on DNS

Ulexus: Ah, so it is related not to Ceph, but to the DNS-bucket (or whatever) mapping?

lorieri: it is related to this:
http://ceph.com/docs/master/install/install-ceph-gateway/#add-wildcard-to-dns

when you don't have the wildcard.

common used when you want encryption but you don't have a wildcard cert

PR #52, opened 2015-03-27
rbd-volume: the possibility to mount the rbd image in read-only mode

dmitryint:
~~~
-e RBD_OPTS=ro
~~~
This gives the possibility to mount the snapshots.


rbd-volume/entrypoint.sh
```diff
@@ -7,6 +7,12 @@ set -e
 : ${RBD_FS:=xfs}
 : ${RBD_TARGET:=/mnt/rbd}
 
+MOUNT_OPTIONS=''
+
+if [ "$RBD_OPTS" == "ro" ]; then
```
leseb: don't you want to add more mount options than ro?
why don't you simply check with -z if the variable is populated?

leseb:
The commit makes sense, I just wonder why you don't make it more general? Just a mount option where we can have whatever we want.

PR #54, opened 2015-03-27
rbd-volume: remove modprobe rbd

dmitryint:
When I try to to load the module from Docker container, an error occurs: modprobe: ERROR: ../libkmod/libkmod.c:556 kmod_search_moddep() could not open moddep file '/lib/modules/3.19.0/modules.dep.bin'. 
I think this is due to the lack of the kernel modules inside the container. Better to load rbd module outside Docker container.

### DIFF
--- a/rbd-volume/entrypoint.sh
+++ b/rbd-volume/entrypoint.sh
@@ -10,9 +10,6 @@ set -e
 # Make sure the mountpoint exists
 mkdir -p ${RBD_TARGET}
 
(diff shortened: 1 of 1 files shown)


leseb:
Just hit the same error on a CoreOS VM, the workaround makes sense.
@Ulexus what do you think?

Ulexus:
I certainly agree; the module should not be loaded within the container.  I'm not exactly sure what I was thinking...

PR #57, opened 2015-03-31
rbd-volume: the possibility to provide filesystem independent mount options to mountWait

dmitryint:
for example:
~~~
docker run ... -e RBD_OPTS=ro,noatime,noexec ceph/rbd-volume
~~~


Ulexus:
Thanks, @dmitryint !

PR #58, opened 2015-03-31
rbd-volume: service for mount rbd volume to host

dmitryint:
A convenience container to mount the requested `RBD` image to a host filesystem mountpoint. You can then use this mountpoint as volume from other containers with the `--volume`  Docker `run` option.

Example usage:
* `mountWait -rbddev $(/usr/bin/docker run --rm -v /etc/ceph:/etc/ceph:ro --privileged --net host -v /dev:/dev --entrypoint '/usr/bin/rbd' ceph/rbd-volume map rbd/vol-rbd) -fstype xfs -target /mnt/vol-rbd`
* [systemd service] (https://github.com/dmitryint/ceph-docker/blob/rbd-volume/rbd-volume/rbd-mount%40.service)


Ulexus:
That's a very interesting way to go about it, and it certainly should work pretty well.  I can't help but think there's a more decoupled means of doing this, though.

A few thoughts:
 * namespace the mount target or allow a means to fully express a more generalized mount target (such as with the `RBD_TARGET` variable, which already exists)
 * split the mapping component from the mounting component (or provide a means to multi-call the container cleanly or, even better, the option to multi-call or pipeline the sequence)... obviously, the mount action itself cannot be executed from inside the container, but perhaps there is a better interface to be had
 * use a "standardized" installation method, like as is done with [ibuildthecloud/systemd-docker](http://https://github.com/ibuildthecloud/systemd-docker/blob/master/image/startup.sh)
 * modify `mountWait` to use the pre-defined environment variables in addition to (or in replacement of) the execution flags

dmitryint:
Voila!

mountWait now works from Docker container. The solution with systemd-docker I liked it, but did not come for a number of reasons.
For mountWait it is important to launch the container with the bind mount option: `-v /proc:/host/proc`

Ulexus:
What a simple, elegant fix!  Thanks, @dmitryint

PR #61, opened 2015-04-15
Add Docker AIO support

leseb:
This commit introduces a container that runs all the Ceph service. This
is likely a step toward a common image that gets activated on demand.
Ultimately we want a single container in which we can activate which
service we want.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get update &&  apt-get install -y wget
 # Install Ceph
 RUN wget -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc' | apt-key add -
 RUN echo deb http://ceph.com/debian-$CEPH_VERSION/ trusty main | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list
-RUN apt-get update && apt-get install -y --force-yes ceph && \
+RUN apt-get update && apt-get install -y --force-yes ceph radosgw && \
 apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install etcdctl
--- a/demo/Dockerfile
+++ b/demo/Dockerfile
@@ -1,0 +1,20 @@
+# DOCKER-VERSION 1.0.0
+#
+# Ceph Demo AIO
+#
+# VERSION 0.0.1
+
+FROM ceph/base
+MAINTAINER Sébastien Han "seb@redhat.com"
+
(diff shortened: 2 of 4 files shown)


Ulexus:
My issue with this is that I would call it more a demo.  It's not really a configuration that I envision being deployed anywhere but for a demo/test/dev rig.  It is what it says:  all-in-one; I'm just hesitant to put it forth... un-caveated?

leseb:
The intended goal is mainly for people not familiar with Ceph to quickly bootstrap a complete cluster.
Actually this is something we need for the Kubernetes guys in this patch: https://github.com/GoogleCloudPlatform/kubernetes/pull/6649/files#r28358644

I'd rather prefer this container to live in this repo than anywhere else.

Sorry what do you mean "un-caveated"? :)

Ulexus:
Sorry; I tend to fabricate terms at times.  I'm really just saying that this feels more like a demo bootstrap... something to let people get a full Ceph cluster up to play with.  It's not designed to be used in production.

I guess I'm really just complaining about the name.  I think there should be an inherent warning to keep people from trying to use this as a canonical production component.  Perhaps am I merely being protective about something that doesn't need it.

As instruction, though, (as for what I'm assuming you are wanting to use it for that PR), I think it is excellent, and I'm happy to pull it in.  I just want to call it "demo" or something, rather than "all-in-one", which, if I were looking, I would say:  "Why mess with all the other junk; let me just use this one."

All of that aside, I still think the idea of having a single container image which is configurably-executable (you can run-time select which ceph component(s) run), it great.

TLDR; don't consider my objection to be a blocker, but a request to consider its name and potential misapprehension by users.

Ulexus:
Of course, perhaps I am missing the point, when you say bootstrap.  Is the idea here that you can create a cluster with all the necessary keys and configuration for export into a production cluster grown from it?  In that case, I would call it "bootstrap" (or similar) and provide details of its use-case in its `README.md`.

leseb:
I agree that this is a demo. Something that can help Docker users unfamiliar with Ceph to quickly get started or test a Ceph feature.
Here I'm trying to redruce all the steps required to get a Ceph up and running.
So basically "docker run ..." would give you everything, after this just collect the keys in /etc/ceph and start doing what you want.

I can flag this container as demo if you prefer, I just thought that saying AIO would ring a bell for educated users.
I'll rename it for "demo".
I should definitely add more context in the README.

leseb:
Just sent a new version.

Ulexus:
Thanks; I'm much more sanguine about this, now.

PR #62, opened 2015-04-16
Container's Dinamic ip as MON_IP

isserrano:
Shouldn't MON_IP be the private ip of the container? If try to use any other ip I cannot get external connectivity.

By doing that nat works straight away and I can connect from other hosts to the monitor.

### DIFF
--- a/mon/entrypoint.sh
+++ b/mon/entrypoint.sh
@@ -12,10 +12,9 @@ if [ ! -n "$MON_NAME" ]; then
    exit 1
 fi
 
-if [ ! -n "$MON_IP" ]; then
-   echo "ERROR- MON_IP must be defined as the IP address of the monitor"
-   exit 1
-fi
+# MON_IP is the private container ip
+MON_IP=`ifconfig |grep "inet addr"|cut -d":" -f2|cut -d" " -f1|grep -v "127.0.0.1"`
+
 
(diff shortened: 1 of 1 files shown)


leseb:
Hum, how do you run your container?
Don't you use -net=host?

isserrano:
No, I am happy with an isolated network for the monitors container,

should I use the same network as the host?

this is how I am running it
 docker run -d --name cephpocmon1 -h cephpocmon1 -e MON_NAME=cephpocmon1
-p 192.168.1.122:6789:6789 -t ceph/mon

Ulexus:
The public IP in this sense is not necessarily the same as that of a SOHO network.  This is referring to the public side of the ceph cluster:  the service side.

There are typically two networks in a ceph cluster:  the internal data exchange and synchronization network, and the public service network.

The "public" network need not be an IP routable on the public internet, but it refers to the IP address which is open to the public users of this cluster.

Ulexus:
I see what you mean, though... if you do not use `--net=host`, this is confusing.

This patch, however, is reliably functional only on systems which have a single IP address, so it's not going to work, generally.

Instead, if you'd like to have `MON_IP` set to the first available IP address to the container with a flag... such as `MON_IP="container"` or something else which doesn't break current deployments, I'd be fine with that.

isserrano:
done now, take a look and merge it if you are happy with that

PR #63, opened 2015-04-16
Remove radosgw package installation

leseb:
The radosgw package is part of ceph/base now, I should have fixed that with #61...


leseb:
@Ulexus I'm trying to apply a simple rule here.
I always try to not merge my own PRs.

So if this PR looks good to you, please merge it :).
Thanks!

Ulexus:
Understood; that seems like a good policy.  I'll try to remember to do the same.

PR #64, opened 2015-04-16
Changes for when using containers network

isserrano:
MON_IP to use container private ip when using containers network

### DIFF
--- a/mon/entrypoint.sh
+++ b/mon/entrypoint.sh
@@ -15,17 +15,9 @@ fi
 if [ ! -n "$MON_IP" ]; then
    echo "ERROR- MON_IP must be defined as the IP address of the monitor"
    exit 1
-fi
-
-# When using contanier network stack we use the first available ip
-if [ "$MON_IP" == "container" ]; then
+# When using contanier network stack we use container's ip
+elif [ "$MON_IP" == "container" ]; then
    MON_IP=$(hostname --ip-address)
-fi
-
-
-if [ "$MON_IP" ]; then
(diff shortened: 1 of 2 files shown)


mon/entrypoint.sh
```diff
@@ -17,6 +17,17 @@ if [ ! -n "$MON_IP" ]; then
    exit 1
 fi
 
+# When using contanier network stack we use the first available ip
+if [ "$MON_IP" == "container" ]; then
+   MON_IP=$(hostname --ip-address)
```
leseb: not sure about the "hostname --ip-address"
on my system it returns 2 addresses.

might I suggest this to get the IP?

ip -4 -o a | awk '/eth0/ { sub ("./..", "", $4); print $4 }'

mon/entrypoint.sh
```diff
@@ -17,6 +17,17 @@ if [ ! -n "$MON_IP" ]; then
...
+fi
+
+
+if [ "$MON_IP" ]; then
```
leseb: use elif maybe instead of 2 if?

mon/entrypoint.sh
isserrano: It looks like inside a container it returns the ip you have in /etc/hosts for the container, the way the container is build, it returns only one ip.
In other systems , yes it can return multiple ips

mon/entrypoint.sh
isserrano: I didnt want to modidy those lines but yes you are right, elif will look better

PR #65, opened 2015-04-16
MON_IP to use container private ip when using containers network

isserrano:


Ulexus:
That looks pretty reasonable, thanks.

I'd like to know, though, under what circumstances this actually works.  I can see using an overlay network (à la flannel) to get away from `--net=host`, but all the testing I've ever done relies on monitors (and OSDs) not changing their IPs.  Otherwise, you end up with consensus problem.  How does this use-case avoid that problem?

leseb:
+1 last time I tried putting monitors on Docker they were complaining about the NAT.

isserrano:
The host where the container is will have a fixed ip, and is the one I will
use for the clients and the osd connecting to the monitors.

Ulexus:
I'm still failing to see how this actually works.  The mon will be thinking it is a different IP than it is.  Are you saying you have this in operation?

isserrano:
No. I am building a new cluster and I would like to use containers, I would
like to use nat for the monitors inside the container, that is why  I am
doing that change.

Ulexus:
I'm pretty sure no one has a functional cluster with NAT'd monitors.  If you get this working, though, I'd be more than interested.

leseb:
+1 NAT'd monitors can't work...

Ulexus:
I'm going to close this for now, @isserrano .  If you have differing results, please reopen it.

PR #69, opened 2015-06-05
added ceph osd create

mingfang:


osd/entrypoint.sh
```diff
@@ -18,6 +18,7 @@ do
 
    # Check to see if our OSD has been initialized
    if [ ! -e /var/lib/ceph/osd/${CLUSTER}-${OSD_ID}/keyring ]; then
+      ceph osd create
```
leseb: Not sure about this one, if we have to use it, this should be after the client.admin.keyring check.
I also think you should catch the OSD_ID at some point.
@Ulexus what do you think? Did we have any good reason to not use the command in the entrypoint?

mingfang: I stole this from the cep/demo script.
My osd would not work without it.

Ulexus:
Like @leseb said, if you create an OSD, you need to capture the created id and use it.  Also, though, the way the OSD container works, you have to know the id before hand.  Either the `OSD_ID` must be set or the directory must be appropriately named.  So far as I know, you cannot tell `ceph osd create` which id to create, which makes it unusable in this container, as is.

That said, I do think it is torturous to have to know the id in advance, and I do want to design a more automated approach.  Unfortunately, this PR alone will not do that, and it doors not work at all, except in very special circumstances (strict, linear additions of OSDs with no skips in the lineup).  Moreover, it has the potential to create a huge number of unused OSDs, without bound.

mingfang:
Just a disclaimer, I'm a complete newbie with ceph and was just trying a bunch of random things to get the various Docker Ceph containers to work.
I was finally able to get ceph/mon, osd, and mds to work and was able to mount cephfs on a client.
And I was not able to get it to work with this line in the script.

Please feel free to delete this PR if it's not the right solution.
But please also update the documentation on how to get the stack to work properly.

leseb:
Thanks for your understanding.
The OSD doc says:

To create your OSDs simply run the following command:
`docker exec <mon-container-id> ceph osd create.`

This is the way to setup it.

Ulexus:
@mingfang Sorry if I was overly harsh-sounding.  I have a tendency toward succinctness, particularly when mobile.

We _do_ appreciate the feedback, input, and help, and I don't want to dissuade you from providing it.

I can certainly understand the confusion, and I definitely _do_ want to modify the container such that it will work a bit more intuitively.

mingfang:
no worries.
thanks for your help.

PR #70, opened 2015-06-05
create pools needed for cephfs

mingfang:


mds/entrypoint.sh
```diff
@@ -18,6 +18,9 @@ fi
 
 # Check to see if we are a new MDS
 if [ ! -e /var/lib/ceph/mds/ceph-$MDS_NAME/keyring ]; then
+   ceph osd pool create cephfs_data 8
```
leseb: :-1: 
I believe pool pg and pgp should be configurable.

leseb:
I just thought about it but this is only necessary for release > 0.87.
I'll need to rework a bit the demo container as well.

Ulexus:
Since the master branch is currently based on Giant, this should be "safe," but one of the advantages of the change is to allow flexibility of the data and metadata pools (even more importantly, to allow for multiple metadata services, for multiple CephFS systems on the same cluster...down the road).  At a minimum, whatever solution we implement here must be able to support custom naming of the data and metadata pools.

I think I'd like to take this as more of a wait-and-see thing, though, since this area of Ceph appears to be in a bit of flux at the moment.

leseb:
What about 4 variables then?

* $cephfs_data_pool
* $cephfs_data_pool_pg
* $cephfs_metadata_pool
* $cephfs_metadata_pool_pg

Ulexus:
Works for me

leseb:
Cool, let's wait for @mingfang now. :)

mingfang:
That works for me too. 
Thanks.

Ulexus:
PR #74 Implements this, mostly as described.

PR #71, opened 2015-06-05
remove /etc/environment

rubycut:
This directory of file doesn't exist on coreos.

### DIFF
--- a/mon/ceph-mon.service
+++ b/mon/ceph-mon.service
@@ -3,7 +3,6 @@ Description=Ceph Monitor
 After=docker.service
 
 [Service]
-EnvironmentFile=/etc/environment
 Environment=MON_IP=${COREOS_PUBLIC_IPV4}
 Environment=MON_NAME=%H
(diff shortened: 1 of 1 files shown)


Ulexus:
Is there a need or reason for this?  I use this statement on all my nodes.

rubycut:
Well, this line breaks service on coreos. It won't start at all because
/etc/environment doesn't exist.

Ulexus:
That's curious; all my hosts are CoreOS.  What version are you running that `/etc/environment` doesn't exist?

Anyway, I wonder if `EnvironmentFile` will take the `-` prefix... something like:
```
EnvironmentFile=-/etc/environment
```

Ulexus:
According to the [documentation](http://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=), it does.  If you wouldn't mind modifying your PR to simply prefix `-` to `/etc/environment`, it should be clean and good.

Ulexus:
PR #73 fixes this issue

PR #72, opened 2015-06-08
WIP: multi-OS container support

leseb:
This is the first bits of a multi-OS support implementation. I don't see
any clear way to specify a base image unless than creating sub
directories. So yes, this duplicates some part of the code but given
that we use automated builds I don't see how we can use Docker's build
system to search for different 'base'.
My idea was to have something like:

FROM %%CEPH_NAMESPACE%%/%%CEPH_PREFIX%%base:%%CEPH_TAG%%

It looks like we can achieve this only by using an external build system
that will push our images to the hub. However that means having an infra
somewhere, maintain it and all the pain that comes with it. I don't
believe @Ulexus or I (@leseb) are willing to invest in such system.

This requires heavy review before merging as this will break our
building system. We need to modify the hub to reflect this change.


Ulexus:
@leseb Would you be averse to reorganizing things such that this _doesn't_ break the build system?  Basically, a structure something like:
```
.
├── base -> root/ubuntu/base
├── mds -> root/ubuntu/mds
├── mon -> root/ubuntu/mon
├── osd -> root/ubuntu/osd
├── rgw -> root/ubuntu/rgw
└── root
    ├── centos
    │   ├── base
    │   ├── mds
    │   ├── mon
    │   ├── osd
    │   └── rgw
    └── ubuntu
        ├── base
        ├── mds
        ├── mon
        ├── osd
        └── rgw
```

Note:  I'm not actually _certain_ this won't break the build system (it depends on the Docker builder honouring symlinks), but it has a better chance.  

It _also_ means that, for the common use cases, people can easily select their build root simply by changing the symlinks.  Of course, we could also prebuild alternate roots as something like `ceph/centos-base`, `ceph/centos-mon`, etc., through the same mechanism.  This leaves open the door to, with minimal impact, swap default roots very easily, and with much testing, should such a thing be desired, in the future.

As a post-script, I'm guessing there is some clear and present desire to offer other build-roots?

leseb:
This is really cleaver :).
Just made the necessary changes.
Just changing the symlink to get another OS is pretty good and flexible.
We could potentially add more OS in the future.

Ulexus:
I'll test the symlink adherence with another project, first.  In the mean time, I just broke the PR; can you rebase, when you get the chance?

leseb:
Yes reviewing #74 at the moment, going to rebase now.

leseb:
The symlinks did their jobs pretty well :), the rebase easy.

Ulexus:
Excellent; and I just tested the symlink dereferencing on the Docker Hub autobuild system.  They work just fine.

docker-registry/Dockerfile
```diff
@@ -4,7 +4,7 @@ ENV SETTINGS_FLAVOR s3
 ENV STORAGE_PATH /registry
 ENV SEARCH_BACKEND sqlalchemy
 ENV AWS_ENCRYPT false
-ENV AWS_DEBUG true 
+ENV AWS_DEBUG true
```
Ulexus: Is this necessary, or is is just an accidental addition?

leseb: Well I'm simply cleaning up the trailing space.

Ulexus:
I think, just for safety, I'd like to separate master from hammer on Docker Hub before merging this.  That should minimize any problems, in case something in this breaks stuff (making this change only apply to the `latest` tag, instead of also applying to `hammer`).  I probably should have done that already, anyway.

Ulexus:
Also, do you intend to include the `rbd*` packages in the build-root alternation?  I don't know if they do anything that is build-root-specific or not.

docker-registry/Dockerfile
Ulexus: Never mind; I just noticed the difference was not boolean, but a stray trailing space

leseb:
I can, this was just a first attempt, will push the change for the remaining directories.

Ulexus:
It might be a good idea to version the bases, too....  `ubuntu-14.04`, for instance.

leseb:
Ubuntu base already have that, for centos we don't.
For simplicity we could start with centos:centos7 tag. What do you think?

Ulexus:
Oh, I was meaning for the directory names, (so that we can offer multi-root with versions, since that's almost as important... and allows testable upgrades).  The main intention was to match the FROM statements to _their_ tags.  Since the `ubuntu/base` explicitly references ubuntu:14.04, I think the directory should be named accordingly.

leseb:
Like this?

ps: we have an issue with centos and runit for the OSDs... I need to find a rpm somewhere.

leseb:
Looks like this:

```
├── base -> root/ubuntu/base
├── config
├── demo -> root/ubuntu/demo
├── docker-registry
├── mds -> root/ubuntu/mds
├── mon -> root/ubuntu/mon
├── osd -> root/ubuntu/osd
├── rados -> root/ubuntu/rados
├── radosgw -> root/ubuntu/radosgw
├── rbd -> root/ubuntu/rbd
├── rbd-lock -> root/ubuntu/rbd-lock
├── rbd-unlock -> root/ubuntu/rbd-unlock
├── rbd-volume -> root/ubuntu/rbd-volume
└── root
    ├── centos
    │   ├── base -> bases/centos7
    │   ├── bases
    │   │   └── centos7
    │   ├── demo
    │   ├── mds
    │   │   ├── fuse
    │   │   └── systemd-unit
    │   ├── mon
    │   │   └── systemd-unit
    │   ├── osd
    │   │   ├── my_init
    │   │   └── systemd-unit
    │   ├── rados
    │   ├── radosgw
    │   ├── rbd
    │   ├── rbd-lock
    │   ├── rbd-unlock
    │   └── rbd-volume
    │       └── mountWait
    └── ubuntu
        ├── base -> bases/ubuntu-14.04
        ├── bases
        │   └── ubuntu-14.04
        ├── demo
        ├── mds
        │   ├── fuse
        │   └── systemd-unit
        ├── mon
        │   └── systemd-unit
        ├── osd
        │   ├── my_init
        │   └── systemd-unit
        ├── rados
        ├── radosgw
        ├── rbd
        ├── rbd-lock
        ├── rbd-unlock
        └── rbd-volume
            └── mountWait
```

leseb:
Found the runit package.

Ulexus:
Yeah, that works.

Ulexus:
The only thing of note I can think of is that I'm pretty sure using symlinks means that people cannot build this on Windows systems.  Of course, I'm not even sure you can do anything with Docker on Windows, so I don't know that that's much of a concern.

Ulexus:
Actually, we do still have a problem.  Can we break this into two stages?

The first stage would be exactly like this PR _except_ that the ubuntu Dockerfiles would reference `ceph/base` instead of `ceph/ubuntu-base`.  The next stage (after we _create_ `ceph/ubuntu-base`) can be to change those back to `ceph/ubuntu-base`.  Otherwise, we have a chicken-and-egg problem.

Ulexus:
I'm also thinking that it _may_ be better to do the build-root selection with branches and tags instead of directory trees and symlinks.  On the one hand, it means maintaining different branches, but since _most_ changes are in the base itself, it shouldn't really be that bad.  (and we'd still have to maintain the separate Dockerfiles and support files in the different directories, anyway... using branches would actually make that easier).

So, for instance, we could simply have branches:
`hammer/ubuntu` build to Docker Hub's `ceph/base:hammer-ubuntu` and `ceph/base:hammer`
`hammer/centos` builds to Docker Hub's `ceph/base:hammer-centos`
`trunk` builds to Docker Hub's `ceph/base:latest` and `ceph/base:latest-ubuntu`
`trunk/centos` builds to Docker Hub's `ceph/base:latest-centos`

This would seem to actually use git in a more natural and advantageous way, since the branches would already maintain the branch-specific modifications, and it would be easy to merge in non-conflicting updates just using git.

Sorry, I should have thought of this at first (though I'm not sold on the nomenclature).

leseb:
I believe using branches and tags make more sense as well.
Moreover users won't get confused with the symlinks, it's more natural to switches branches instead of changing symlinks. Symlinks can be messed up really easily too. Branches will avoid confusions.

So what's the action plan?

I'm going to rollback "bases" and keep ceph/base.

leseb:
Last version? :)

I also changed the rbd-lock and tbd-unlock, back to initial FROM

Ulexus:
If the goal is to go with the branches and tags approach, you should probably just push a 'centos' branch forked from `master` to be the HEAD for the centos-rooted containers, modifying directly the top-level daemons `mon`, `osd`, etc.  That will be a perpetual branch that you should be able to maintain sync with master easily.

Of course, then we'll need to finalize the naming scheme for the Docker tags so that you can set the proper FROM statements for them.

PR #73, opened 2015-06-08
Make EnvironmentFile includes optional

Ulexus:
This makes the EnvironmentFile includes from each of the sample unit files optional, meaning that the units will not fail if the file does not exist.


leseb:
@Ulexus If you feel confident with this one, I'll let you merge it.

Ulexus:
Yes, I've tested it; the `-` prefix works as expected (ignores the failure to load the file).

PR #74, opened 2015-06-08
Add support for cephfs and pool creation to MDS

Ulexus:
This PR adds support for _optional_ (and disabled by default) CephFS creation to the MDS.  

Even if CephFS creation is enabled, it should check, first, to make sure that the pools and filesystem don't already exist before creating them.

This should be non-breaking to existing installations.

### DIFF
--- a/mds/entrypoint.sh
+++ b/mds/entrypoint.sh
@@ -3,8 +3,21 @@ set -e
 
 # Expected environment variables:
 #   MDS_NAME - (name of metadata server)
+# Optional environment variables:
+#   CEPHFS_NAME (defaults to 'cephfs')
+#   CEPHFS_DATA_POOL (defaults to ${CEPHFS_NAME}_data)
+#   CEPHFS_DATA_POOL_PG (defaults to 8)
+#   CEPHFS_METADATA_POOL (defaults to ${CEPHFS_NAME}_metadata)
+#   CEPHFS_METADATA_POOL_PG (defaults to 8)
 # Usage:
 #   docker run -e MDS_NAME=mymds ceph/mds
+
+: ${CEPHFS_CREATE:=0}
+: ${CEPHFS_NAME:=cephfs}
+: ${CEPHFS_DATA_POOL:=${CEPHFS_NAME}_data}
+: ${CEPHFS_DATA_POOL_PG:=8}
+: ${CEPHFS_METADATA_POOL:=${CEPHFS_NAME}_metadata}
+: ${CEPHFS_METADATA_POOL_PG:=8}
 
 if [ ! -n "$MDS_NAME" ]; then
    echo "ERROR- MDS_NAME must be defined as the name of the metadata server"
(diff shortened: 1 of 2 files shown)


mds/entrypoint.sh
```diff
@@ -34,6 +47,27 @@ if [ ! -e /var/lib/ceph/mds/ceph-$MDS_NAME/keyring ]; then
...
    fi
+
+fi
+
+# Create the Ceph filesystem, if necessary
+if [ $CEPHFS_CREATE -gt 0 ]; then
+   FS_EXISTS=$(ceph fs ls | grep -c name:.${CEPHFS_NAME},)
+   if [ $FS_EXISTS -eq 0 ]; then
+      # Make sure the specified data pool exists
+      ceph osd pool stats ${CEPHFS_DATA_POOL} >/dev/null
```
leseb: s/exists/does not exist/  ?

mds/entrypoint.sh
```diff
@@ -34,6 +47,27 @@ if [ ! -e /var/lib/ceph/mds/ceph-$MDS_NAME/keyring ]; then
...
+      if [ $? -ne 0 ]; then
+         ceph osd pool create ${CEPHFS_DATA_POOL} ${CEPHFS_DATA_POOL_PG}
+      fi
+
+      # Make sure the specified metadata pool exists
```
leseb: s/exists/does not exist/  ?

mds/entrypoint.sh
```diff
@@ -34,6 +47,27 @@ if [ ! -e /var/lib/ceph/mds/ceph-$MDS_NAME/keyring ]; then
       # Generate the new MDS key
       ceph auth get-or-create mds.$MDS_NAME mds 'allow' osd 'allow *' mon 'allow profile mds' > /var/lib/ceph/mds/ceph-${MDS_NAME}/keyring
    fi
+
+fi
+
+# Create the Ceph filesystem, if necessary
+if [ $CEPHFS_CREATE -gt 0 ]; then
```
leseb: if we want to stick with the recommendation of the README, should we check if CEPHFS_CREATE=1 instead?

mds/entrypoint.sh
Ulexus: It shouldn't be... `grep -c` should return the count of matches, so 0 means it does not exist, >0 means it _does_ exist.  Unless I'm cognitively broken; it is still morning, here.

mds/entrypoint.sh
Ulexus: Yes, we should.  Updating...

mds/entrypoint.sh
leseb: yes but at this point the pool shouldn't exist (the data one).
the way you check that the data pool exists or not is with the `$?`where the `grep` checks for the existence of the ceph filesystem.
so to me you're checking that the data pool doesn't exist and depending on the output you do the necessary things.

Ulexus: Yeah, the reason I check for the filesystem with grep is that there is no way to query for a particular filesystem (that I know of).   Checking the return value of the `ceph osd pool stats` command is vastly more reliable that trying to grep the stdout of the `ceph fs ls`.  Unfortunately, I couldn't find a similar command to check for the existence of the filesystem.

Ulexus: Also, the order is intentional, here.  I don't want to create the specified pools if the filesystem already exists; they would just be wasted.  Hence, I'm _only_ creating the pools if the filesystem doesn't already exist.  It's not intuitive, but I think it's safer than creating arbitrary pools just because a user told the mds container to create the filesystem without reading the docs on it.

Ulexus: So, ultimately:
1) don't do _anything_ unless we are told to create the filesystem
2) don't do anything if the filesystem already exists
3) Otherwise, create each pool if _they_ don't exist
4) Create the filesystem

Ulexus: The external wrapper for the create is _also_ intentional, so that the user doesn't have to have the admin client key installed on the MDS server.

leseb: I understood the logic, I think it's more a phrasing issue on my side. I'd have phrased differently. :)

Ulexus: Sorry for any pedantry; it's hard to quit doing.

Anyway, you're right.  I would prefer to have been more parallel in the evaluations, but I didn't see a way to do it without using double negatives.

PR #78, opened 2015-06-10
WIP: first iteration of a unique daemon image

leseb:
The idea is pretty straighforward we simply pass a new env var and boot
a monitor like this:

`sudo docker run -d --net=host -v /etc/ceph:/etc/ceph -e CEPH_DAEMON=MON
-e MON_IP=192.168.0.20 -e CEPH_NETWORK=192.168.0.0/24 ceph/daemon`

So far, I've been able to successfully bootstrap MON, MDS and RGW.
I had to fix some MDS issue. Because we use `set  -e` we can not really
trap a command return code in a variable, the script will exist before
that since the command potentially returns something different than 0.

For the OSD, it's probably me...

I couldn't find a better name than "daemon" for now.
Let's first discuss the implementation.
I also added some meaningful log messages for the OSD.


leseb:
This probably needs a bit more documentation...

Ulexus:
Thanks @leseb .  "daemon" works for me.

A few ideas:

  *  `CEPH_DAEMON` should accept a space-delimited list of daemons to run.  Yes, recommended practice is to not combine them, but (among other things), this would make your demo package easier.
  *  For the two machine-bound services (`mon` and `osd`), I wonder if we could build on the OSD's autodetection concept by checking to see if directories for mon and osd services exist in standard locations.  In this way, we could have the daemons start automatically based on directory structure, allowing a fully generic host service definition (i.e., "start Ceph on this server").

The bootstrapping, to be offers an interesting quandary.  If we are not instructed to start a `mon`, but we are not bootstrapped (no ceph config or keys), should the container simply die?  Should it attempt to create a mon and bootstrap?  Perhaps if there is no daemon specified, it could bootstrap a mon?

leseb:
I was initially thinking of allowing multiple daemons running into a single container. Later I remembered that it is not what we recommend. The implementation might be tricky as well lead to undesired behaviours.
Moreover I don't think this will help the demo container as the entrypoint is really unique. The way services are configured is intended for a demo nothing else. (low pg/pgp, hardcode pool name etc...)
I don't really want to mix what we consider for production usage and what we recommend as a sandbox.
This is why I'd like to keep both 'daemon' and 'demo' separated. This will probably avoid confusions too.

So yes, in the end I'm more incline to force users to run micro service containers instead of running multiple daemons. Given that containers are really lightweight I don't any reason why someone would want to run more than one daemons inside a container.

There is definitely room for improvement for the OSD part as I believe the current state is too complex and not user-friendly at all. We should think of a new design I guess.

Regarding the bootstrapping, we could let daemons die but a check to verify that a communication can be established with a monitor is probably better. This will avoid unpleasant debugging. I can work on something.
Finally, should we bootstrap a mon if non exist, I'd say why not. Generally I'd rather return an error if the user didn't follow the proper steps. So the error provides some guidance on how to do thing properly. Doing thing under the hood by workarounding users mistake is not really a good idea. If it fails the user will learn why and will set it up properly.

Ulexus:
> Doing things under the hood by workarounding users mistake is not really a good idea

That's a good way of thinking about it.  Yes, I would have to agree.  

You're also correct about the multiple daemons (aside from the OSD and workarounds required there...but that doesn't have any bearing on the daemon selection).

Another thing to consider with the single-container thing:  we don't have to worry about backward compatibility, so we _should_ probably reorganize the osd entrypoint, since it's rather messy.  Maybe the mon, too.  I've been thinking of the various ways to integrate these with etcd/confd (in a flexible and not-mandatory way).  

I'll definitely say that I am presently caught up with the concept of directory-described execution... and I think we could definitely provide means of pulling down configs and keys (from etcd, consult, S3, URL, etc.).

Ulexus:
(perhaps?) Configuration and key extraction procedure:

1.  Check for local `<file>` in `/etc/ceph/`
2.  Check for `CONFIG_METHOD` (defaults to none)
    1. Attempt to pull `<file>` via `CONFIG_METHOD` handler
    2. (For `<file>` other than `ceph.conf` and `ceph.client.admin.keyring`) Call `ceph auth get-or-create`
3.  Otherwise fail or bootstrap, as appropriate

Ulexus:
I think the current OSD detection routine is actually fairly good; it should just be simplified and better documented.  (referring to the directory detection).

Bootstrapping OSDs, though, is a bit painful, at the moment.  Specifically, the need to create the OSD outside of the container is counter-intuitive.  If the client.admin keyring is available, though, we should be able to have the script fully bootstrap an OSD (including creation of that OSD).  This would come close to @hookenz addition concept:  create and mount the directory, and the container takes care of everything else.

leseb:
Just did some cleanup and added some functions to check several things (`ceph.conf` and admin key).

If we focus on the `CONFIG_METHOD`, I agree that we should be compliant with several config stores.
It looks like you want to use these stores to 'store' keys and `ceph.conf`.
I also think that we could use this for configuration options too, that is more difficult to do actually.

leseb:
Can we maybe first merge the daemon prototype and then in another PR I'll work on the config store backends?
I think it's just too much to do in one shoot.

Ulexus:
Oh, certainly.

On Wed, Jun 10, 2015, 12:48 Leseb <notifications@github.com> wrote:

> Can we maybe first merge the daemon prototype and then in another PR I'll
> work on the config store backends?
> I think it's just too much to do in one shoot.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ceph/ceph-docker/pull/78#issuecomment-110829959>.
>
-- 
Seán C McCord
CyCore Systems, Inc
888-240-0308

hookenz:
This looks interesting.

By the way, why do you pass CEPH_DAEMON= as an environment variable and not as a command to run?
And why are you using environment variables for CEPH_NETWORK when you could read this from ceph.conf?

hunter:
Looks like a nice approach.

Is the assumption that multiple OSDs will be run from a single container to get around the issues with inter-host communication?

leseb:
@hookenz mainly because we need to configure the container upfront, if we simply run the command we need to know the id of the daemon. Does it answer your question because I'm not sure if I got it correctly.
We use the CEPH_NETWORK var only for the monitors while building the first monitor and its `ceph.conf`. Since monitors can't run without the `--net=host` flag we need to specify the host network to the container.

@hunter to be honest I'm not sure what's the plan. However what I'm sure is that I'd like to refactor the OSD part and restart something from scratch. For example, I think we should start using ceph-disk to bootstrap them.

hunter:
Agreed. Since the OSDs are such a critical part of running a scalable Ceph cluster it's important that we build something that's going to be flexible for mounting, hot swapping, journals, etc. ceph-disk will probably help with a number of those things. I'll test out a few ideas when I get a spare minute.

leseb:
I started to play with ceph-disk inside a container, results seem to be rather random. From what I observed privileged mode is needed.

leseb:
@Ulexus I just added the support for bootstrap keys since it's part of the best practices to used them instead of always requiring the admin key.

leseb:
Just added an experimental scenario for the OSDs

hookenz:
I don't think you should default to starting any daemon if not specified.  That might lead to rogue or orphaned mon's if your not careful. Better to just exit with error and some help text.

leseb:
@hookenz no worries I don't :)

leseb:
@Ulexus can you do a last round on this one?
I think we are good to merge it.

daemon/README.md
```diff
@@ -0,0 +1,176 @@
...
+#### Multiple OSDs ####
+
+There is a problem when attempting run run multiple OSD containers on a single docker host.  See issue #19.
+
+There are two workarounds, at present:
+* Run each OSD with a separate IP address (e.g., use the new Docker 1.5 IPv6 support)
+* Run multiple OSDs within the same container
+
+To run multiple OSDs within the same container, simply bind-mount each OSD datastore directory:
+* `docker run -v /osds/1:/var/lib/ceph/osd/ceph-1 -v /osds/2:/var/lib/ceph/osd/ceph-2`
```
Ulexus: Or, since we are now providing the OSD directory option as a top-level feature, the example here should probably reflect that:  export the osd directory

daemon/README.md
```diff
@@ -0,0 +1,176 @@
...
+ -  `WEIGHT` is the of the OSD when it is added to the CRUSH map (default is `1.0`)
+ -  `JOURNAL` is the location of the journal (default is the `journal` file inside the OSD data directory)
+ -  `HOSTNAME` is the name of the host; it is used as a flag when adding the OSD to the CRUSH map
+
+The old option `OSD_ID` is now unused.  Instead, the script will scan for each directory in `/var/lib/ceph/osd` of the form `<cluster>-<osd-id>`.
+
+To create your OSDs simply run the following command:
+
+`docker exec <mon-container-id> ceph osd create`.
+
```
Ulexus: Might as well continue with the instructions:

*  Create a filesystem on your OSD(s)
*  Mount your OSD(s) in to `/var/lib/ceph/osd/<cluster>-<osd-id>`
*  Execute the OSD container:  ```docker sudo docker run -d --net=host ...`

We could just merge this as is, and I can update the documentation accordingly, though.

daemon/ceph-osd.service
```diff
@@ -0,0 +1,20 @@
...
+Description=Ceph OSD
+After=docker.service
+RequiresMountsFor=/var/lib/ceph/osd/ceph-%i
+
+[Service]
+EnvironmentFile=-/etc/environment
+Environment=OSD_ID=%i
+Environment=HOSTNAME=%H
+ExecStartPre=/usr/bin/docker run --rm -v /opt/bin:/opt/bin ibuildthecloud/systemd-docker
+ExecStart=/opt/bin/systemd-docker --env run --rm --name %p-%i --net=host --privileged -v /etc/ceph:/etc/ceph -v /dev:/dev -v /var/lib/ceph/osd/ceph-%i:/var/lib/ceph/osd/ceph-%i ceph/osd
```
Ulexus: I'll update this to follow the new pattern after merge

daemon/entrypoint.sh
```diff
@@ -0,0 +1,327 @@
...
+
+  if [[ "$(sudo parted --script ${OSD_DEVICE} print | egrep '^ 1.*ceph data')" ]]; then
+    echo "ERROR- It looks like this device is an OSD, use OSD_FORCE_ZAP to use this device"
+    exit 1
+  elif [[ "$(sudo parted --script ${OSD_DEVICE} print | egrep '^ 1.*ceph data')" && ${OSD_FORCE_ZAP} -eq "1" ]]; then
+    ceph-disk -v zap ${OSD_DEVICE}
+  fi
+
+  if [[ ! -z "${OSD_JOURNAL}" ]]; then
+    ceph-disk -v prepare ${OSD_DEVICE}:${OSD_JOURNAL}
```
hunter: What does ceph-disk do in this case? I can't recall what it supports. Do partitions need to be precreated? Will it format the journal if it doesn't exist?

Ulexus:
Architecturally, I don't know that we particularly need to separate the execution commands for `OSD_DEVICE` and `OSD_DIRECTORY`.  We should be able to simply check the /var/lib/ceph/osd tree on start to see if there exist OSDs in that directory, then check for `OSD_DEVICE`, then fail.  That also means that you can bootstrap an OSD using `--privileged`, but on subsequent runs, have the host mount the OSD and run without `--privileged`.

daemon/entrypoint.sh
leseb: `ceph-disk` will create partitions if they don't exist. In this case it will create a small partition on the OSD_JOURNAL device. This makes me think that we should make the journal size configurable. Adding this.

On the other case it will create partitions, fs, create osd id, populate osd dir, mount etc

daemon/README.md
```diff
@@ -0,0 +1,176 @@
...
+### Ceph disk ###
+
+```
+$ sudo docker run -d --net=host \
+--privileged=true \
+-v /etc/ceph:/etc/ceph \
+-v /var/lib/ceph/:/var/lib/ceph/ \
+-v /dev/:/dev/ \
+-e CEPH_DAEMON=OSD \
+-e OSD_DEVICE=/dev/vdd \
```
hunter: This doesn't support multiple devices?

What about the docker run 'devices' option? Redundant with mounting /dev but any point in also using?

Ulexus:
One other thing:  is there any particular reason to require the execution be an environment variable instead of a parameter?  If we are calling this the entrypoint script, docker automatically passes the arguments to its execution on.  Hence, we could simply execute `docker run ceph/daemon osd` instead of `docker run -e CEPH_DAEMON=CEPH_OSD_DEVICE ceph/daemon`.  The former seems more natural, to me.

daemon/README.md
leseb: I've tried the `--device` option without luck. Even doing `-v /dev/vdb` didn't work. The only way for me to get it working was to use `-v /dev:/dev`...

leseb:
I agree it makes more sense to run `docker run ceph/daemon osd` unfortunately I don't know how to do this :). Assistance required on this. :)

Ulexus:
Ah, no problem.  I say we go ahead and merge this but not publish it to Docker Hub yet.  Then we can work on it... WIP, after all.

daemon/README.md
```diff
@@ -0,0 +1,176 @@
...
+Deploy an OSD
+-------------
+
+There are two available options:
+
+* use `OSD_CEPH_DISK` where you only specify a block device
+* use `OSD_DIRECTORY` where you specify an OSD mount point to your container
+
+
+### Ceph disk ###
```
hunter: It seems Ceph disk is for single use only? Looks like it checks the partition and exits (unless a zap is needed). Should it be documented that the directory mode should be used on following runs?

daemon/README.md
hunter: I'm guessing ceph-disk needs access to the other parts of /dev? /dev/disk/by-*

leseb: Correct. `sgdisk` uses disk's uuid.

daemon/README.md
leseb: Yes we should clarify that preparation steps are needed before running the container and exposing the OSD directory.

PR #80, opened 2015-06-12
Convenience wrappers I call "ceph-cli"

hookenz:
I've created a few convenience wrappers which I hope you may find useful.
Currently I have ceph and ceph-disk.

They are simply shell scripts around docker run and can be installed in much the same way as systemd-docker.  Which is where I got the idea.

If you think this is a good way to use ceph from the host going forward, feel free to add some more, or I'll do so.


cli/ceph-disk
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+/usr/bin/docker run --rm --privileged=true -v /etc/ceph:/etc/ceph -v /dev:/dev ceph/base ceph-disk "$@"
```
leseb: I think you should also expose `var/lib/ceph/bootstrap-osd` so `ceph-disk` can add new OSDs using the bootstrap key.
I did some (unsuccessful) experiment trying to get `ceph-disk activate` I'll be happy to hear if you got it working :).

hookenz: I'll take a look.

leseb: I just managed to create OSDs with `ceph-disk` somehow using `-v /dev:/dev` instead of `-v /dev/sdb:/dev/sdb worked.`

hookenz:
Yeah. After that I'm temp mounting the disk and putting the osd is into the whoami file.  Then id you start the ceph osd it'll be initialised.  That process is what ceph-disk activate does. I'm thinking of emulating that process when you call activate   with ceph disk because the real activate starts the things differently than our server.  I've got an idea on how to do that

hookenz:
Excuse the spelling and grammar. Wrote that in on my phone.  Hope it makes sense.

Ulexus:
I like this; I use something similar (but simpler) on my workstation:

```
#!/bin/bash

ARGS=""
if [ $# -gt 0 ]; then
   ARGS="ceph $*"
else
   ARGS="bash"
fi
ssh -t $CEPH_MON docker exec -it ceph-mon $ARGS
```

I don't know that I would call it `cli`, though.  My hunch would be to create an `/examples/coreos/` directory and file it there as `tools`.

hookenz:
Had thought it calling it tools later.  We could do that.

I also was thinking of making it start a container that would use docker exec.  That way it would work on any monitor or ceph node.  As some tools like ceph-disk you want on a node.  but if we remove the admin key then ceph status won't work on them.

hookenz:
I've rearranged everything as suggested.  I've also added my mount-osds script which you may find useful.  It's how I'm mounting them under CoreOS.

Ulexus:
LGTM; @leseb ?

leseb:
Thanks @hookenz!

PR #82, opened 2015-06-15
Add a k8s pod file

leseb:
I recently gave a first and quick try to Kubernetes.
This is a first step toward running Ceph in Docker managed by
Kubernetes.

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -1,0 +1,18 @@
+Usage
+=====
+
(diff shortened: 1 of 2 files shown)


Ulexus:
Perhaps we should have an `/examples` top-level directory which can house these types of things (along with the systemd unit files and the coreos/fleet examples).

leseb:
Yes I think it's a good idea.

PR #83, opened 2015-06-15
Add argument parsing option for daemon selection

Ulexus:
Allows the specification of the daemon to run by command-line argument instead of environment parameter.

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -20,19 +20,46 @@ set -e
 : ${RGW_REMOTE_CGI_PORT:=9000}
 : ${RGW_REMOTE_CGI_HOST:=0.0.0.0}
 
-
-if [ ! -n "$CEPH_DAEMON" ]; then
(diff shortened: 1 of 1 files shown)


daemon/entrypoint.sh
```diff
@@ -34,6 +27,40 @@ if [[ ! -e /etc/ceph/${CLUSTER}.conf ]]; then
...
+# If we are given a valid first argument, set the
+# CEPH_DAEMON variable from it
+case "$1" in
+   mds)
+      CEPH_DAEMON=MDS
+      ;;
+   mon)
+      CEPH_DAEMON=MON
+      ;;
+   osd)
```
leseb: Currently we don't have an `OSD` variable but only `OSD_DIRECTORY` and `OSD_CEPH_DISK`.

Ulexus: Sneaky me... that's my next PR #84

PR #84, opened 2015-06-15
Example common OSD differentiator

Ulexus:
Add a differentiator to allow for common OSD execution (automatically figure out what to do).

Note:  better would be to do this, _then_ try ceph-disk, _then_ try manual bootstrapping.  That, however, would require a bit more rearrangement.


leseb:
LGTM :+1:

PR #90, opened 2015-06-16
Necessary changes for k8s to work

leseb:
We can't use `--net=host` with k8s. Thus if we manually set `MON_IP` as
an environement variable we will get duplicate content when using a
replication controller.
Introducing a new variable to auto detect the IP address of the
container.
This will generally work for every use case that don't want to use
`--net=host`.

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -76,7 +76,7 @@ if [[ "$CEPH_DAEMON" = "MON" ]]; then
 
   if [ ${MON_IP_AUTO_DETECT} -eq 1 ]; then
     MON_IP=$(ip -6 -o a | grep scope.global | awk '/eth/ { sub ("/..", "", $4); print $4 }' | head -n1)
-    if [ -z $MON_IP ]; then
+    if [ -z "$MON_IP" ]; then
       MON_IP=$(ip -4 -o a | awk '/eth/ { sub ("/..", "", $4); print $4 }')
     fi
   elif [ ${MON_IP_AUTO_DETECT} -eq 4 ]; then
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -4,6 +4,7 @@ set -e
 : ${CLUSTER:=ceph}
 : ${CEPH_CLUSTER_NETWORK:=${CEPH_PUBLIC_NETWORK}}
 : ${MON_NAME:=$(hostname -s)}
+: ${MON_IP_AUTO_DETECT:=0}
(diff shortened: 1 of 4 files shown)


daemon/entrypoint.sh
```diff
@@ -73,6 +74,10 @@ if [[ "$CEPH_DAEMON" = "MON" ]]; then
     exit 1
   fi
 
+  if [ ${MON_IP_AUTO_DETECT} -eq 1 ]; then
```
Ulexus: I'm going to be a pain about this:  I propose that `MON_IP_AUTO_DETECT` can be set to various options:
*  0 = disabled
*  1 = enabled, automatic (IPv6, falling back to IPv4)
*  4 = enabled, IPv4
*  6 = enabled, IPv6

hunter:
Have we thought about what to do when the containers are moved around by k8s, fleet, etc (node failure, etc)? The monmap is going to get pretty weird. Do we need some code to clean out old mons?

We have some pretty gnarly ansible tasks for handling IP/Port changes, we might be able to repurpose if it makes sense.

Ulexus:
@hunter It's definitely a concern, but it's not actually all that different with and without kubernetes.  We have the same problem on any system where you don't have statically-defined mon's.

A plus of kubernetes is that we can define static IPs and ports; they just have to be defined outside of the execution environment.  This patch is really about allowing that orientation of IP derivation.  That is, the container internally determines its IP rather than being passed as a parameter of execution.  It's not as robust, but it should be adequate for most purposes.... and this does _not_ preclude passing in the mon's IP.

Ulexus:
@leseb  Also, I wonder if we ought not set the autodetect to 1 if it is not defined _and_ if no `MON_IP` was passed... rather than simply failing.  On the one hand, that _might_ be more dangerous; on the other, it makes the container operate in a more useful default action.

leseb:
If someone could try the ipv6, looks like on my system it detects ipv6 (and it's enabled) but fail to bind.
We might need to stick with either 4 or 6, if the user is sure about its setup.

Ulexus:
This works just fine, now, on my IPv6 setup.  If your IPv6 setup is (the default) link-local, it's not going to work... but then, a link-local setup isn't going to work, anyway.  The only concern is if some distribution out there enables IPv6 by default, but doesn't properly configure it otherwise.

Ulexus:
Now, this will skip link-local IPv6 addresses, so it should properly fall back to IPv4 if IPv6 is enabled in Docker but unconfigured.

PR #94, opened 2015-06-17
Remove all the roles in favour of the daemon container

leseb:
Since #78, we introduced the daemon container that centralizes all the
Ceph daemons instead of having separate container images. We now only
have one.


Ulexus:
I think the best way to do this is to kill the build triggers to the others first, merge this PR, then create the build trigger for it.  Being a large change, though, I'd really like to take the opportunity to roll in the other major changes at the same time:  etcd/consult/confd, more fine-grained detection, etc.

Also, the documentation, most everywhere, needs to be updated.

leseb:
Ok let's wait a bit and suspend this PR for now.

hookenz:
I think that should go for the convenience wrappers too.  I think the wrappers are too heavyweight and don't merit a container for each one.  You could just do what I did with ceph-tools, so instead of:

docker run -v /etc/ceph:/etc/ceph ceph/rados lspools

Have a script like this:
---- rados ----
#!/bin/sh
docker exec ceph-daemon rados "$*"
----------------

I've been experimenting with:
#!/bin/bash

# Find local ceph daemon
NAME=$(docker inspect --format {{.Name}} ceph-mon 2>/dev/null) || \
     $(docker inspect --format {{.Name}} ceph-mds 2>/dev/null) || \
     $(docker inspect --format {{.Name}} ceph-osd 2>/dev/null)

# strip leading '/' 
NAME=${NAME:1}

echo "Name: $NAME"
if [ -n "${NAME}" ]; then
  # Run the command
  docker exec ${NAME} rados "$@"
else
  echo "Error: No local ceph daemon is running, rados can not be run right now"
fi

Ulexus:
@hookenz If you are running a docker daemon on the host in question, then yes, you might as well just `docker exec` within that daemon's container space.  The convenience wrappers were designed for execution on machines which are not otherwise running a docker daemon.  Still, they are not at all heavyweight, considering they are all built on `ceph/base`.  While `ceph/base` itself is large, the layered approach of Docker images means there is very little extra disk space for the wrappers (mostly just the script itself), and the running overhead of a container itself is not much more than a normal process.

That said, I do think the convenience wrappers can be rolled in to the `ceph/daemon` (whose name would not make much sense at that point, I admit) simply by adding handlers for each of them as the first commandline parameter.

leseb:
I think we need to make clearer documentation if we don't merge this. I keep getting emails about people using ceph/mon/osd and so forth.

@Ulexus I'm not entirely sure how to proceed but here are a couple of ideas:

* remove everything and keep the current work on ceph/daemon. This will be clearer for users but will likely put people using the old images in a bad situation.
* add warning and depreciation messages both on github and docker hub

Can we somehow keep the images if we remove the repos?

Ulexus:
@leseb  My thought was to make the removal at the next release fork, but yes, we should probably note the deprecation in the README, and I think I would add a WARNING log entry to the executions, too.

klausenbusk:
Shouldn't this be merged?

Ulexus:
@klausenbusk We are waiting to do this when we move to the next major stable release

Ulexus:
At this point, a few of the external daemons (`ceph/radosgw` and `ceph/mds`) are failing to build their related containers.  I think we can probably remove the external roles and just keep `ceph/base` and `ceph/daemon` at this point.  The existing Docker containers will remain available, but we can remove the code from the Github repo.

Ulexus:
I have now shut down the Docker automatic builds for the external daemons, so we should be clear to merge this PR.

leseb:
Finally :). Merging this!

Issue #34, opened 2015-03-03
Usage of confd to make the repo more universal


leseb:
Proposed by @hunter in #28 .
Discussion can continue here.

Ulexus:
This sounds good to me; confd also, conveniently, supports environment variables as a backend.

We should standardize the configuration schema based on the work from @BugRoger 's `ceph/config`, I'm guessing.

hunter:
Great! confd should give us a little more flexibility. I'd started with direct consul-template integration but I'll see if I can switch it over to pure confd.

One issue I was dealing with was the initial config. @BugRoger's looks to take a good approach. Any thoughts though on how to handle large numbers of configuration options? We currently use ceph-ansible which has a good set of defaults but I hadn't found an clean way to get a large number of 'default' variables into the KV.

leseb:
I'm not really familiar with confd but can we set a number of values in etcd and then iterate over them to build the ceph.conf? Or is it pure tempting with only static options?

Another solution might be to inject args, however not everything can be injected.
In which value are you interested in?

hunter:
The two options I'd been pondering were bulk insertion into the KV before launch or setting a bunch of defaults as part of the entrypoint script. I think the first option should give a little more flexibility for a larger number of defaults but doesn't make it self contained.

For the initial untested experimentation I'd just added a function for adding defaults in the entrypoint (borrowed from the deis store) - https://github.com/AcalephStorage/ceph-docker/tree/wip-confd

Ulexus:
@leseb Yes, confd does support iterators, at least with etcd and consul... I think we're out of luck for environment variables.

@hunter I wonder if we can have confd mutate the names of the keys, converting `_` into spaces.  That would provide a structured but somewhat free-form way of storing configuration values.

hunter:
@Ulexus I was thinking of doing the same thing. I'd started with almost a straight copy/paste of the ceph-ansible ceph.conf but makes more sense to let the KV store provide the structure.

Ulexus:
Interestingly, etcd does support spaces in keynames just fine, so we may not actually need to adapt it at all.  I don't know what potential problems that opens up, though.  It does sound a bit dangerous.
```
etcdctl set /ceph/osd/5/"osd use stale snap" true
etcdctl get /ceph/osd/5/"osd use stale snap"  # returns true
```
works just fine

hunter:
It seems that consul also does something similar (although needs to be url encoded)

```
[...]

PR #100, opened 2015-06-18
fix broken OSD_DIRECTORY mode

dmitryint:


leseb:
Thanks dmitry for fixing this.  Perhaps we can add runit to base. Since it's not really tight to this container. 

@ulexus? Thoughts?

hunter:
What's the purpose of using runit/my_init over something a little more simple (like forego)?

Ulexus:
Nothing; it's just the tool at hand.  If you have a better runit alternative, I'm pretty sure we are -happy to use it... as long as we are consistent across all of the packages (though the "all" here will _soon_ be "one").

Anyway, @dmitryint is certainly correct; we missed the actual execution of the daemon at start.  I think we can go ahead and merge this, and if @hunter has a better init option, that is most welcome, but it should go in a separate PR.

leseb:
I'm good with that as well, merging :).

Ulexus:
@leseb  I think @hunter is talking about _replacing_ it with a simpler alternative, not so much moving it to outside of `ceph/base`.

hunter:
Sorry, should have been more clear. Just a simpler replacement for myinit and runit. We've used forego successfully in the past which is just a single Go binary+config for running multiple processes.

I'm just working through some code for #34 to make it work with the daemon image. Might take a look at this after I'm done with the confd stuff.

Ulexus:
@hunter I hadn't heard of [forego](https://github.com/ddollar/forego) before (and it's not exactly high in the Google search listings), but yes, that looks like a much cleaner, simpler option.

leseb:
@hunter it was clear :), I was simply wondering despite of what we use that we should put it on ceph/base :). That's it.

PR #103, opened 2015-06-23
KV Backend

hunter:
Following from the discussion kicked off in #34 I have started working on a KV storage for config storage. I thought I'd open a PR early so that we can discuss and make improvements as we go along.

The initial work has added confd and a simple tool - kviator to manage keys in different backends (consul and etcd). Populating the KV is currently handled by a shell script but could be moved to a config parameter or image.

The daemon entrypoint has basic support for mon configs (although is lacking the latest bootstrap keyring changes). Still needs some cleanup and testing in a container. Following that the next step is the OSD config.


dmitryint:
Why instead kviator not to use native tools from etcd and consul? Kviator is a very young project, it has not yet gathered a large community. As I understand it, it does not support many features from those that are in native clients, such as encryption, authentication and acl. And, in the future, there is no guarantee that there will be support latest features of consul and etcd.

hunter:
Kviator is a brand new project, we created it specifically for managing the different KV stores from the command line. It does use libkv in the backend which has been created for Docker so it's robust enough for their needs. We can certainly look at reducing the dependency but it does mean adding more complexity to the entrypoint and including etcdctl/consul/zk in the base image.

daemon/entrypoint.sh
```diff
@@ -28,6 +32,89 @@ if [[ ! -e /etc/ceph/${CLUSTER}.conf ]]; then
...
 
+function create_mon_ceph_config_from_kv {
+
+  CLUSTER_PATH=ceph-config/${CLUSTER}
+
+  echo "Adding Mon Host - ${MON_NAME}"
+  kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} cas mon_host/${MON_NAME} ${MON_IP} > /dev/null 2>&1
+
+  # Acquire lock to not run into race conditions with parallel bootstraps
+  until kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} cas ${CLUSTER_PATH}/lock $MON_NAME > /dev/null 2>&1 ; do
```
leseb: s/$MON_NAME/${MON_NAME}

daemon/entrypoint.sh
```diff
@@ -28,6 +32,89 @@ if [[ ! -e /etc/ceph/${CLUSTER}.conf ]]; then
...
+    until ./confd -onetime -backend ${KV_TYPE} -node ${KV_IP}:${KV_PORT} -prefix="/${CLUSTER_PATH}/" -confdir="./conf.d" ; do
+      echo "Waiting for confd to update templates..."
+      sleep 1
+    done
+
+    echo "Adding Keyrings"
+    kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} get ${CLUSTER_PATH}/monKeyring > /etc/ceph/ceph.mon.keyring
+    kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} get ${CLUSTER_PATH}/adminKeyring > /etc/ceph/ceph.client.admin.keyring
+
+    if [ ! -f /etc/ceph/monmap ]; then
```
leseb: Is this check really necessary?
I mean, don't we always fall under this condition?

If monSetupComplete exists, I think it's natural to assume that `/etc/ceph/monmap` doesn't exist.

daemon/entrypoint.sh
```diff
@@ -28,6 +32,89 @@ if [[ ! -e /etc/ceph/${CLUSTER}.conf ]]; then
...
+      echo "Monmap is missing. Adding initial monmap..."
+      kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} get ${CLUSTER_PATH}/monmap > /etc/ceph/monmap
```
leseb: Why do you do both?
Shouldn't kviator return the same value as the ceph command?

daemon/entrypoint.sh
```diff
@@ -28,6 +32,89 @@ if [[ ! -e /etc/ceph/${CLUSTER}.conf ]]; then
...
+
+    echo "Creating Keyrings"
+    ceph-authtool /etc/ceph/ceph.client.admin.keyring --create-keyring --gen-key -n client.admin --set-uid=0 --cap mon 'allow *' --cap osd 'allow *' --cap mds 'allow'
+    ceph-authtool /etc/ceph/ceph.mon.keyring --create-keyring --gen-key -n mon. --cap mon 'allow *'
+
+    echo "Creating Monmap"
+    monmaptool --create --add ${MON_NAME} ${MON_IP} --fsid ${FSID} /etc/ceph/monmap
+
+    echo "Importing Keyrings and Monmap to KV"
+    MONKEYRING=$(cat /etc/ceph/ceph.mon.keyring)
```
leseb: Why do you need this variable?
Doesn't kviator support files directly?

leseb:
@hunter thanks a lot for this PR. I'm glad to see that we are moving forward with the KV stuff :).

daemon/entrypoint.sh
hunter: That monSetupComplete condition is still run after initial set up on the same host to keep the ceph.conf up to date so there are cases where monmap can exist. The check for the monmap was to avoid overwriting if the monmap has been modified (new hosts added, etc).

I probably should spend a little more time getting familiar with the ceph monmap code since I'm not entirely clear on what happens in cases of old monmaps or new hosts with changing hostnames ;)

daemon/entrypoint.sh
hunter: Not yet. We're just about to add support for stdin

hunter:
Quite a few commits but we've managed to get a working KV mon cluster on plain old Docker. Now testing on K8s.

With the autodetection that we've already got for OSD directories, I'm wondering if there is really any needed for the KV store for OSDs (after the conf/keys are set up). @leseb, @Ulexus can you think of any cases where it would be of use?

leseb:
@hunter sorry for the late reply. I can't think of anything that would require it (at least for now).
I'd like to move forward with this, I think we should merge it (as soon as you rebase your work) and then.
@Ulexus any objections?

Ulexus:
A couple (related) concerns:
  * It looks like the system will break if not _all_ the configuration keys for each component are defined
  * It looks like you can _only_ define configuration for values which have been predetermined and coded by the script

I like that the system is well-structured, but I'm concerned about its potential fragility and lack of flexibility.

I haven't looked at the code very thoroughly, though... perhaps I am missing something?

hunter:
Yes, that's the one issue I've found with confd is that it requires all keys to exist in the KV store if they are specified in the templates. The only alternative I can think of at the moment is looping through all keys and parsing out the key and value. Might make the config a little more complex but would likely mean less hands on changes.

hunter:
@Ulexus, I've made some changes to allow dynamic configuration of the ceph.conf.

Not pretty since go's text/template doesn't handle whitespace well but it does mean that the ceph.conf will only have keys that are added to the KV.

daemon/confd/templates/ceph.conf.tmpl
```diff
@@ -0,0 +1,51 @@
...
+  {{else}}
+  cephx_service_require_signatures = false
+  {{end}}
+{{else}}
+  auth cluster required = none
+  auth service required = none
+  auth client required = none
+  auth supported = none
+{{end}}
+{{range gets "/global/*"}}
```
Ulexus: These ranges need to be closed (with `{{end}}`), I believe.

Ulexus: Never mind, I'm blind.  Strange to have the end tacked on the line

Ulexus:
@hunter Yes, thanks!  I'm much happier with that.

@leseb ?  Anything else?

leseb:
@leseb not really, I'm good with that. Feel free to merge.

daemon/confd/templates/ceph.conf.tmpl
hunter: Yeah, I'm not a fan either but since text/template's whitespace handling is rough in some places the `{{end}}` on the same line avoids excess line breaks.

PR #112, opened 2015-07-09
Introduce a ceph_disk_activate

leseb:
In 'some' cases the operators might not want to run ceph-disk prepare
inside the container but might want it outside.
In this scenario, we only activate the OSD and don't prepare it.

Signed-off-by: leseb <seb@redhat.com>

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -121,7 +121,7 @@ function create_mon_ceph_config_from_kv {
     kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} put ${CLUSTER_PATH}/bootstrapOsdKeyring - < /var/lib/ceph/bootstrap-osd/ceph.keyring
     kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} put ${CLUSTER_PATH}/bootstrapMdsKeyring - < /var/lib/ceph/bootstrap-mds/ceph.keyring
     kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} put ${CLUSTER_PATH}/bootstrapRgwKeyring - < /var/lib/ceph/bootstrap-rgw/ceph.keyring
-    
+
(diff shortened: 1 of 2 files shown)


daemon/entrypoint.sh
```diff
@@ -415,6 +418,29 @@ elif [[ "$CEPH_DAEMON" = "OSD_CEPH_DISK" ]]; then
...
+  ceph_config_check
+
+  if [[ -z "${OSD_DEVICE}" ]];then
+    echo "ERROR- You must provide a device to build your OSD ie: /dev/sdb"
+    exit 1
+  fi
+
+  mkdir -p /var/lib/ceph/osd
+
+  ceph-disk -v activate ${OSD_DEVICE}1
```
Ulexus: Is this trailing 1 intentional?

daemon/entrypoint.sh
```diff
@@ -415,6 +418,29 @@ elif [[ "$CEPH_DAEMON" = "OSD_CEPH_DISK" ]]; then
...
+  OSD_ID=$(cat /var/lib/ceph/osd/$(ls -ltr /var/lib/ceph/osd/ | tail -n1 | awk -v pattern="$CLUSTER" '$0 ~ pattern {print $9}')/whoami)
+  OSD_WEIGHT=$(df -P -k /var/lib/ceph/osd/${CLUSTER}-$OSD_ID/ | tail -1 | awk '{ d= $2/1073741824 ; r = sprintf("%.2f", d); print r }')
```
Ulexus: Is there something magic about the number 1073741824?

daemon/entrypoint.sh
leseb: Not really. Removing it.

daemon/entrypoint.sh
leseb: Actually this is the formula used in the ceph init script to set the weight of an OSD, see here: https://github.com/ceph/ceph/blob/697a20a0c542576a0e4ecf0e06414affeb64f1e0/src/init-ceph.in#L368

Basically 1073741824 bytes = 1GB.

PR #115, opened 2015-07-10
Fix KV backend implementation

leseb:
Prior to this patch confd was looking for the auth directory, which
doesn't exist since the fsid key gets stored in common.
I tried to set a different path in confd but somehow failed, so I simply
changed the way the key is stored. I know it's not the most elegant but
at least it works. @hunter if you have a better option please let me
know.

Adding some content to the kv populate script as well since confd complains that no mds
path can be found.
Add some docs too.

Use the kv function to generate a configuration file for other nodes and
don't anything if the mon setup is not complete.
Remove snippet for mount option, it fails mounting the partition.

With this patch I successfully bootstrapped 3 mons on 3 separate machines, one OSD, MDS and RGW on another machine :)

Signed-off-by: leseb <seb@redhat.com>


hunter:
Thanks for the fixes @leseb! We were working through some similar things but hadn't got to the other services.

Issue #116, opened 2015-07-12
ceph/daemon totally broken


dmitryint:
for example, if run MDS:
```
sudo docker run -d --net=host \
-v /var/lib/ceph/:/var/lib/ceph/ \
-v /etc/ceph:/etc/ceph \
-e CEPHFS_CREATE=1 \
ceph/daemon mds
```

in entrypoint.sh kviator runs with `--kvstore=none` argument, what causes error:  `Backend storage not supported yet, please choose another one`

Ulexus:
That's not good.  I'm guessing these `KV_TYPE` checks are failing... looking into it

Ulexus:
Nope; leaky `KV_TYPE` walling... line 502 is the problem.  I'm going to hack-fix this for now, but this script needs to be refactored.

dmitryint:
Sorry, I was wrong not all broken, but only MDS and RESTAPI.

dmitryint:
line 572?

Ulexus:
Yeah, I see.  I'm working on a cleanup right now that will get that, too.

Ulexus:
Feel free to PR a quick-fix for that one, too, though.  This cleanup will take me a while, but I'll be happy to merge in a hack-fix

Ulexus:
I believe that should do it; reopen, if necessary

PR #118, opened 2015-07-12
Refactor entrypoint script

Ulexus:
The entrypoint script has gotten a bit ragged; this is an attempt to refactor it, and make it a bit more flexible, with less repetition.

In particular, I have broken out the "static" and "kv" backends, which both adds separation and leaves open the option of others providing their own configuration backends.  More thought and work is probably in order.


Ulexus:
Reference #116

PR #120, opened 2015-07-12
Refactor entrypoint script

Ulexus:
The entrypoint script has gotten a bit ragged; this is an attempt to refactor it and make it a bit more flexible, with less repetition.

In particular, I have broken out the "static" and "kv" backends, which both adds separation and leaves open the option of others providing their own configuration backends. More thought and work is probably in order.

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -441,8 +441,8 @@ case "$CEPH_DAEMON" in
       if [ ! -n "$CEPH_DAEMON" ]; then
           echo "ERROR- One of CEPH_DAEMON or a daemon parameter must be defined as the name "
           echo "of the daemon you want to deploy."
-          echo "Valid values for CEPH_DAEMON are MON, OSD_DIRECTORY, OSD_CEPH_DISK, OSD_CEPH_DISK_ACTIVATE, MDS, RGW, RESTAPI"
-          echo "Valid values for the daemon parameter are mon, osd_directory, osd_ceph_disk, osd_ceph_disk_activate, mds, rgw, restapi"
+          echo "Valid values for CEPH_DAEMON are MON, OSD, OSD_DIRECTORY, OSD_CEPH_DISK, OSD_CEPH_DISK_ACTIVATE, MDS, RGW, RESTAPI"
+          echo "Valid values for the daemon parameter are mon, osd, osd_directory, osd_ceph_disk, osd_ceph_disk_activate, mds, rgw, restapi"
           exit 1
       fi
       ;;
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -383,7 +383,7 @@ function start_restapi {
   get_admin_key
   check_admin_key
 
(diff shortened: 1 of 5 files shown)


daemon/entrypoint.sh
```diff
@@ -304,47 +123,54 @@ fi
...
+         osd_directory
+         ;;
+      disk)
+         osd_disk
+         ;;
+      activate)
+         osd_activate
+         ;;
+      *)
+         if [ -n "$(find /var/lib/ceph/osd -prine -empty)" ]; then
```
hunter: Should the arg here but `-prune`?

hunter:
So much cleaner! A big +1 from me :)

daemon/entrypoint.sh
Ulexus: Thanks!

leseb:
I like this approach too, I believe the same path should be followed for internal daemons.
So could also stripe this to monitor, osd, mds, radosgw and restapi.
This can be part of another PR though.

@Ulexus stupid question maybe, but did you get the chance to test this out? If so and if everything is fine I'm good for merging this work. If it's not the case I can find some time to test this.

Ulexus:
@leseb No, it's only rudimentarily tested.  I'll see if I can throw up a cluster and test it more throughly; I'm on holiday for the next couple of weeks, so it might not happen quickly.

leseb:
@Ulexus Ok thanks for letting me know, I'll then try to find some time to test this and eventually merge it.
Enjoy your holiday!

daemon/entrypoint.sh
```diff
@@ -462,34 +277,34 @@ elif [[ "$CEPH_DAEMON" = "OSD_CEPH_DISK_ACTIVATE" ]]; then
...
+       KEYRING_OPT="--keyring /etc/ceph/${CLUSTER}.client.admin.keyring"
+    elif [ -e /var/lib/ceph/bootstrap-mds/ceph.keyring ]; then
+       KEYRING_OPT="--keyring /var/lib/ceph/bootstrap-mds/ceph.keyring"
+    else
+      echo "ERROR- Failed to bootstrap MDS: could not find admin or bootstrap-mds keyring.  You can extract it from your current monitor by running 'ceph auth get client.bootstrap-mds -o /var/lib/ceph/bootstrap-mds/ceph.keyring'"
       exit 1
     fi
 
-    timeout 10 ceph --cluster ${CLUSTER} --name client.bootstrap-mds --keyring /var/lib/ceph/bootstrap-mds/ceph.keyring health || exit 1
+    timeout 10 ceph --cluster ${CLUSTER} --name client.bootstrap-mds $KEYRING_OPT || exit 1
```
leseb: You forgot the "health", the command is broken.

Ulexus: _That's_ why that command didn't make sense!  Thanks

leseb:
I tested all the daemons, everything works with the last mds fix, merging this now.

Ulexus:
Thanks @leseb

PR #122, opened 2015-07-15
Adding Mon Host: key is already set

h0tbird:
Running `entrypoint.sh mon` more than once results in `key is already set`.
I think it should be safe to rewrite this key as many times as needed.
This is what I get otherwise:
```
+ echo 'Adding Mon Host - core-1.cell-0.dc-1.demo.lan'
Adding Mon Host - core-1.cell-0.dc-1.demo.lan
+ kviator --kvstore=etcd --client=127.0.0.1:4001 cas ceph-config/ceph-cell-0/mon_host/core-1.cell-0.dc-1.demo.lan 172.16.0.81
root@core-1:/# kviator --kvstore=etcd --client=127.0.0.1:4001 cas ceph-config/ceph-cell-0/mon_host/core-1.cell-0.dc-1.demo.lan 172.16.0.81
key is already set
```

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -49,7 +49,7 @@ function create_mon_ceph_config_from_kv {
   CLUSTER_PATH=ceph-config/${CLUSTER}
 
   echo "Adding Mon Host - ${MON_NAME}"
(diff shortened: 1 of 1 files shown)


Ulexus:
I agree; this looks reasonable to me.  @leseb or @hunter?

leseb:
Looks good to me as well, I'd just want to merge #120 before (still under testing), so you don't need to rebase. Just prioritising things here.

Ulexus:
This is trivial enough; I'll rebase.

PR #123, opened 2015-07-15
Be more ${CLUSTER} aware

h0tbird:
Using a cluster name other than `ceph` is not working.
This PR is to fix a few lines of code I have spotted.

Unfortunately I can't set `--cluster` when I call `ceph-disk activate`:

````
+ ceph-disk -v activate /dev/sda1
INFO:ceph-disk:Running command: /sbin/blkid -p -s TYPE -ovalue -- /dev/sda1
INFO:ceph-disk:Running command: /usr/bin/ceph-conf --cluster=ceph --name=osd. --lookup osd_mount_options_xfs
2015-07-15 15:18:53.769457 7ffabab197c0 -1 did not load config file, using default settings.
INFO:ceph-disk:Running command: /usr/bin/ceph-conf --cluster=ceph --name=osd. --lookup osd_fs_mount_options_xfs
````

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -32,7 +32,7 @@ set -e
 
 function ceph_config_check {
 if [[ ! -e /etc/ceph/${CLUSTER}.conf ]]; then
-  echo "ERROR- /etc/ceph/ceph.conf must exist; get it from your existing mon"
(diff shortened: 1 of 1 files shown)


Ulexus:
I'm pretty sure we have more of these cluster oversights.  Thanks for catching this one!

PR #125, opened 2015-07-15
Crush location expert and multi cluster support

### LINKED ISSUE #121: Support CRUSH locations
Spawned by discussion from #113 

CRUSH location types [http://ceph.com/docs/master/rados/operations/crush-map/](http://ceph.com/docs/master/rados/operations/crush-map/):
```
type 0 osd
type 1 host
type 2 chassis
type 3 rack
type 4 row
type 5 pdu
type 6 pod
type 7 room
type 8 datacenter
type 9 region
type 10 root
```

I think we should offer a free-form CRUSH_LOCATION for "experts," but we should also offer each of the types explicitly (ex. CRUSH_RACK), whose purpose and use should be more easily-grasped, though such may be more verbose.

leseb:
Fixes #121 and more.

### DIFF
--- a/daemon/config.kv.sh
+++ b/daemon/config.kv.sh
@@ -32,11 +32,11 @@ function get_mon_config {
     mkdir -p /var/lib/ceph/bootstrap-{osd,mds,rgw}
 
     echo "Adding Keyrings"
-    kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} get ${CLUSTER_PATH}/monKeyring > /etc/ceph/ceph.mon.keyring
-    kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} get ${CLUSTER_PATH}/adminKeyring > /etc/ceph/ceph.client.admin.keyring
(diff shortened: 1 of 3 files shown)


Ulexus:
The PR description is a bit misleading, since most of this if actually multi-cluster stuff, but it looks good to me.

Ulexus:
Also, the `CEPH_OPTS` is a great idea.

leseb:
You're right, I'll test this tomorrow, just to make sure I haven't broke anything.

Ulexus:
I do think it should be overridable, though... same as the others at the top of the file; set the default, allowing users to override it, should they wish.

Ulexus:
Fixes #121

leseb:
I think it's a bit risky to let users modify the behaviour of the entrypoint as this can completely mess up the bootstrap of a container.
Moreover I can't think of anything that the users would wish to modify. Can you?

Ulexus:
I have no idea; I don't have any need to do so, but I can certainly see that it might theoretically be useful for something.  On the other hand, people who want that much control can just use base and handle everything themselves.  You're probably right.

leseb:
Was testing this and everything was working accept the MDS.
The MDS will be fixed by #127 

Merging this work first.

PR #127, opened 2015-07-16
entrypoint.sh: mds auth error w/ admin keyring

mbessler:
mds won't start for me with the following error:

2015-07-16 16:20:14.168684 7f95b6028700  0 librados: client.bootstrap-mds authentication error (1) Operation not permitted
Error connecting to cluster: PermissionError

when running 
   ceph --cluster ${CLUSTER} --name client.bootstrap-mds $KEYRING_OPT health
with $KEYRING_OPT set to
   --keyring /etc/ceph/${CLUSTER}.client.admin.keyring

It appears that the admin keyring (/var/lib/ceph/bootstrap-mds/ceph.keyring) is not accepted, however, it does work with the bootstrap-mds keyring in /var/lib/ceph/bootstrap-mds/ceph.keyring 

Is this the right approach? I've got 7 machines running each a container with mon and osd so far.

### DIFF
--- a/daemon/config.kv.sh
+++ b/daemon/config.kv.sh
@@ -32,11 +32,11 @@ function get_mon_config {
     mkdir -p /var/lib/ceph/bootstrap-{osd,mds,rgw}
 
     echo "Adding Keyrings"
-    kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} get ${CLUSTER_PATH}/monKeyring > /etc/ceph/ceph.mon.keyring
-    kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} get ${CLUSTER_PATH}/adminKeyring > /etc/ceph/ceph.client.admin.keyring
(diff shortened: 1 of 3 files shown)


leseb:
I've been experiencing the same issue and was troubleshooting this :).
Actually this is normal, we are using the admin key with the bootstrap-mds user.

The fix should be:

```
    if [ -e /etc/ceph/${CLUSTER}.client.admin.keyring ]; then
       KEYRING_OPT="--name client.admin --keyring /etc/ceph/${CLUSTER}.client.admin.keyring"
    elif [ -e /var/lib/ceph/bootstrap-mds/${CLUSTER}.keyring ]; then
       KEYRING_OPT="--name client.bootstrap-mds --keyring /var/lib/ceph/bootstrap-mds/${CLUSTER}.keyring"
    else
```

And then remove the `--name` from the ceph health call.

leseb:
I just merged #125 please rebase your work :).

mbessler:
closing in favor of #128 (note to self: practice PR rebase flow)

PR #128, opened 2015-07-16
entrypoint.sh: mds auth error w/ admin keyring

mbessler:
mds won't start for me with the following error:

2015-07-16 16:20:14.168684 7f95b6028700 0 librados: client.bootstrap-mds authentication error (1) Operation not permitted
Error connecting to cluster: PermissionError

when running
ceph --cluster ${CLUSTER} --name client.bootstrap-mds $KEYRING_OPT health
with $KEYRING_OPT set to
--keyring /etc/ceph/${CLUSTER}.client.admin.keyring

It appears that the admin keyring (/var/lib/ceph/bootstrap-mds/ceph.keyring) is not accepted, however, it does work with the bootstrap-mds keyring in /var/lib/ceph/bootstrap-mds/ceph.keyring

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -296,18 +296,18 @@ function start_mds {
      mkdir -p /var/lib/ceph/mds/${CLUSTER}-${MDS_NAME}
 
     if [ -e /etc/ceph/${CLUSTER}.client.admin.keyring ]; then
-       KEYRING_OPT="--keyring /etc/ceph/${CLUSTER}.client.admin.keyring"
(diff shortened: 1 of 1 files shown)


mbessler:
This is a rebased version with a more proper name from #127

mbessler:
I've applied your 'should be' fix from #127. Let me know if I overlooked something.

Ulexus:
This is curious; you're saying the authentication fails if you use the client.admin keyring, but it does not if you use the bootstrap-mds keyring?  That sounds very odd.

I don't see any problem with explicitly defining the credential names, though.

Ulexus:
Thanks @mbessler

PR #135, opened 2015-07-24
Scripts to mount rbd devices directly CoreOS host without ceph-common

hookenz:
These scripts allow mounting rbd devices directly on the CoreOS host without requiring ceph-common.
Mounted filesystems may then be bind-mounted to docker containers without special privileges.

They may be copied directly or dropping via cloud-init or some other means.


leseb:
Thanks @hookenz.

PR #137, opened 2015-07-28
Updated docs and fixed service file so that it actually works

hookenz:
The service file wasn't quite working, and the docs were lacking some vital details.


leseb:
Thanks @hookenz

PR #138, opened 2015-07-28
Default to osd_activate if OSD_DEVICE is provided

h0tbird:
This patch is to dynamically choose between `osd_directory` and `osd_activate`.
The currently implementation forces `osd_directory` always.

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -143,7 +143,13 @@ function start_osd {
          osd_directory
          ;;
       disk)
-         osd_disk
+         if [ -n "$(find /var/lib/ceph/osd -prune -empty)" ]; then
+           echo "No bootstrapped OSDs found; trying ceph-disk on ${OSD_DEVICE}"
+           osd_disk
+         else
+           echo "Bootstrapped OSD(s) found; ${OSD_DEVICE} won't be bootstrapped this time"
+           osd_activate
+         fi
          ;;
       activate)
          osd_activate
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -143,13 +143,7 @@ function start_osd {
          osd_directory
          ;;
       disk)
-         if [ -n "$(find /var/lib/ceph/osd -prune -empty)" ]; then
-           echo "No bootstrapped OSDs found; trying ceph-disk on ${OSD_DEVICE}"
-           osd_disk
-         else
-           echo "Bootstrapped OSD(s) found; ${OSD_DEVICE} won't be bootstrapped this time"
-           osd_activate
-         fi
+         osd_disk
          ;;
       activate)
          osd_activate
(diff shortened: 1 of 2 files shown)


daemon/entrypoint.sh
```diff
@@ -153,8 +153,13 @@ function start_osd {
             echo "No bootstrapped OSDs found; trying ceph-disk"
             osd_disk
          else
-            echo "Bootstrapped OSD(s) found; using OSD directory"
-            osd_directory
+            if [ -z "${OSD_DEVICE}" ]; then
+                echo "Bootstrapped OSD(s) found; using OSD directory"
+                osd_directory
+            else
+                echo "Bootstrapped OSD(s) found; using ${OSD_DEVICE}"
```
leseb: hum if an OSD device is found why don't we try ceph-disk instead of osd_activate?

h0tbird:
Because ceph-disk will wipe out the device and I want to re-use the data.
I plan to run the container like this:

````
ExecStart=/usr/bin/sh -c "/usr/bin/docker run \
  --privileged=true \
  --net=host \
  --name=ceph_osd \
  --volume /var/lib/ceph:/var/lib/ceph \
  --volume /dev:/dev \
  --env CLUSTER='ceph' \
  --env OSD_DEVICE=/dev/sda \
  --env OSD_FORCE_ZAP=1 \
  --env KV_TYPE=etcd \
  --env KV_IP=127.0.0.1 \
  ceph/daemon osd"
````

And my understanding is that:
- The very first time the container runs it will execute `osd_disk` (because `if` condition will be `true`)
- From now on every time the container is restarted the `if` condition will be `false`
- So now we have to choose whether we run `osd_directory` or `osd_activate`
- If the user provides `OSD_DEVICE` I assume `osd_activate` is prefered

Let me know if that makes sense. I am just learning how Ceph works.

leseb:
Ok I didn't quite catch that this was a reboot scenario.

Since you use `OSD_FORCE_ZAP=1`, yes `ceph-disk` will wipe the disk. If you don't set this there is a check here https://github.com/ceph/ceph-docker/blob/master/daemon/entrypoint.sh#L243. I created this variable mainly for convenience during my tests (useful when you want to re-use a device and start from scratch).
But then I understand that you need this flag in order to not exit the entrypoint. Unfortunately I don't see how we can detect that the container is issuing a stop/start or a reboot. So I guess your approach is valid, if a device was passed then we can assume `activate` is desired.

It is correct to assume that the entrypoint doesn't handle restart really well. And I'm happy to see that someone is working with us on it :), so thanks for that.
Would you mind updating the readme to give some instruction on how to issue a reboot. Simply highlight that  `OSD_FORCE_ZAP=1` is required.

h0tbird:
I have updated the README with a fourth case. What do you think @leseb?

leseb:
Thanks @h0tbird!

PR #139, opened 2015-07-28
Remove health warn with PGs

leseb:
When Ceph detects that the number of PGs is not sufficient it shows a
warn on its status. This is quiet anoying, so removing this warning.
I just wonder if you should make this configurable or not... For me it's
just a pain but would like feedback from others.

### DIFF
--- a/daemon/config.static.sh
+++ b/daemon/config.static.sh
@@ -20,6 +20,7 @@ auth client required = cephx
 public network = ${CEPH_PUBLIC_NETWORK}
 cluster network = ${CEPH_CLUSTER_NETWORK}
 osd journal size = ${OSD_JOURNAL_SIZE}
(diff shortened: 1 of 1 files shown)


leseb:
@hunter @Ulexus @hookenz any cons about this? It's minor, but I just want to make sure that this sounds good to you.

hookenz:
Optional sounds good to me. 
One way to deal with these things and future proof it is allowing an additional conf to be merged in with the generated one.  So maybe we can pass that file in during init.

leseb:
@hookenz yeah just thought about that this morning as well.
I think we should be able to provide an external file where we can put config options inside.
Then append its content during the `ceph.conf` generation.

I'll work on something along those lines.

PR #140, opened 2015-07-29
Check if the directory does not exist or is empty

h0tbird:
Because right now if the directory does not exist the `if` statement returns `false` and it must return `true`.


leseb:
Thanks @h0tbird

Issue #147, opened 2015-08-21
How can I configure two networks for ceph-docker cluster?


tangzhankun:
Hi,
I want to enable two networks using ceph-docker as production environment.
But current "entrypoint.sh" only create a very simple ceph.conf also no interface for adding cluster network configuration.
How can we support public and cluster network in ceph-docker?

I am not sure if there is a need for IPtable settings, but refer to this if you want:
http://ceph.com/docs/master/rados/configuration/network-config-ref/

Ulexus:
You can specify the cluster network by passing the environment variable `${CEPH_CLUSTER_NETWORK}`.  Unfortunately, you are correct; it is not handled by anything for the `ceph.conf` generation.

To get this working with the KV-store, you would add it to the `globals` section... thought this is undocumented and likely to change.  See issue #146

leseb:
Hum which image do you use?
If you ceph/daemon this done here: https://github.com/ceph/ceph-docker/blob/master/daemon/config.static.sh#L20-L21

So passing both `CEPH_CLUSTER_NETWORK` and `CEPH_PUBLIC_NETWORK` should be good.

As for the kv store, just populate the proper subnets just as referred here: https://github.com/ceph/ceph-docker/blob/master/examples/confd/populate.sh#L36-L37

Am I missing something?

hookenz:
@leseb - I think the addition of a feature that can suck in a file that contains additional options and is merged with the one built during initialisation is needed.  It'll then cater for most peoples needs and future options.

Unfortunately it's not entirely simple as it needs to come from somewhere that the init process can read from like a bind-mount drive or read from stdin or something.

tangzhankun:
hi @leseb,
thanks. I see. But I am using ceph/mon now.
adding these two lines in ceph.conf is enough? I saw the document of ceph. Seems ceph.conf should have "host" in it like this:
[osd.0]
        host = {hostname}

And IPtables described there, don't know if we should set IPtables. any idea?

BTW, Will ceph/mon, ceph/osd and ceph/radosgw be removed in the futue? If so, I'll change to ceph/daemon

tangzhankun:
@hookenz 
Is it a must for ceph.conf to be a full version when bootstrapping the initial mon? I mean, could the daemons have different ceph.conf?
For instance, the first generated ceph.conf may have simplest configurations in it. But daemons launched later could insert additional configurations(maybe specific configurations for one OSD) to it? 
[...]

PR #149, opened 2015-09-04
Fix key addition for osd_directory scenario

leseb:
Fixes: #147


Ulexus:
That makes more sense.  Thanks!

Issue #152, opened 2015-09-10
API error (400): client is too old, minimum supported API version is 1.12


hookenz:
It's a good thing I'm not yet using my storage cluster with live data.  CoreOS decided to do an update overnight and now the ceph-mon's aren't starting.

Here is the error in the journal

```
Sep 10 02:24:14 service2-2 systemd[1]: Starting Ceph Monitor...
Sep 10 02:24:14 service2-2 systemd-docker[9264]: 2015/09/10 02:24:14 API error (400): client is too old, minimum supported API version is 1.12, please upgrade your client to a newer version
```
If I google that, this problem is something to do with an upgrade to docker 1.7 from 1.6.  What's it all about?  

I'm using a pretty standard systemd unit file to start this which is:
```
[Unit]
Description=Ceph Monitor
After=docker.service

[Service]
EnvironmentFile=/etc/environment
Environment=MON_IP=${COREOS_PRIVATE_IPV4}
Environment=MON_NAME=%H
ExecStartPre=/bin/mkdir -p /etc/ceph /var/lib/ceph
ExecStartPre=/usr/bin/docker run --rm -v /opt/bin:/opt/bin ibuildthecloud/systemd-docker
ExecStart=/opt/bin/systemd-docker --env run --rm --name %p --net=host -p 6789:6789 -v /var/lib/ceph:/var/lib/ceph -v /etc/ceph:/etc/ceph ceph/mon
Restart=always
RestartSec=10s
Type=notify
NotifyAccess=all
TimeoutStartSec=120
TimeoutStopSec=15

[Install]
WantedBy=multi-user.target
```

hookenz:
Oh, sorry guys I just realised I need to put this against systemd-docker not ceph.  But to be honest I experimented with systemd docker and I actually felt it to be more trouble that it's worth.  Since I never start/stop containers with the docker client in production normally it's not a problem.

hookenz:
Actually problem/solution here... but not been officially fixed.  This is yuck.

Ulexus:
I had forgotten my service examples used systemd-docker.  That really should be an issue in and of itself, to remove those references.

I use it, personally, because it reduces a lot of boilerplate, and because it uses control group-based process control so that when you _think_ you kill a container (by systemd/fleet), it really is killed.

[...]

PR #153, opened 2015-09-10
Update systemd/CoreOS unit file examples

Ulexus:
Remove references to unmaintained systemd-docker and provide more
examples for using ceph/daemon.


Ulexus:
Fixes #152

PR #154, opened 2015-09-15
add missing cluster option in demo cmds and config.

rochaporto:
enable multiple clusters in same box with host networking.

add missing --cluster in ceph osd/fs/etc commands. add cluster aware
monmap creation and reference (allows different port in the mon).

### DIFF
--- a/demo/entrypoint.sh
+++ b/demo/entrypoint.sh
@@ -6,6 +6,7 @@ set -e
 : ${MON_NAME:=$(hostname -s)}
 : ${RGW_CIVETWEB_PORT:=80}
 
+CEPH_OPTS="--cluster ${CLUSTER}"
 
 #######
 # MON #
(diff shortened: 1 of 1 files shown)


demo/entrypoint.sh
```diff
@@ -77,17 +77,17 @@ if [ ! -e /var/lib/ceph/mon/${CLUSTER}-${MON_NAME}/keyring ]; then
...
-   ceph-mon --mkfs -i ${MON_NAME} --monmap /etc/ceph/monmap --keyring /tmp/${CLUSTER}.mon.keyring
+   ceph-mon --mkfs -i ${MON_NAME} --monmap /etc/ceph/${CLUSTER}.monmap --keyring /tmp/${CLUSTER}.mon.keyring
 
    # Clean up the temporary key
    rm /tmp/${CLUSTER}.mon.keyring
 fi
 
 # start MON
-ceph-mon -i ${MON_NAME} --public-addr ${MON_IP}:6789
+ceph-mon -i ${MON_NAME} --public-addr ${MON_IP}
```
rochaporto: i missed it here, will repatch after any comments.

demo/entrypoint.sh
```diff
@@ -77,17 +77,17 @@ if [ ! -e /var/lib/ceph/mon/${CLUSTER}-${MON_NAME}/keyring ]; then
    mkdir -p /var/lib/ceph/mon/${CLUSTER}-${MON_NAME}
 
    # Prepare the monitor daemon's directory with the map and keyring
-   ceph-mon --mkfs -i ${MON_NAME} --monmap /etc/ceph/monmap --keyring /tmp/${CLUSTER}.mon.keyring
+   ceph-mon --mkfs -i ${MON_NAME} --monmap /etc/ceph/${CLUSTER}.monmap --keyring /tmp/${CLUSTER}.mon.keyring
```
rochaporto: i missed it here, will repatch after any comments.

leseb:
Thanks @rochaporto, for this commit, if I wanted to be picky I could say that you should have re-used a similar fashion as https://github.com/ceph/ceph-docker/blob/master/daemon/entrypoint.sh#L34 but I'm fine with this.
@Ulexus thoughts?

rochaporto:
I'm happy to do that, and resubmit if you want (will do it anyway for the missing --cluster that i commented on).

leseb:
If you don't mind I think it will be more consistent with what we have :).

leseb:
Thanks @rochaporto

PR #159, opened 2015-09-30
Added OSD_GET_ADMIN_KEY option

h0tbird:
If OSD_GET_ADMIN_KEY=0 (default) nothing happens.
If OSD_GET_ADMIN_KEY=1 and a KV backend is present it gets the admin keyring and stores it in `/etc/ceph/${CLUSTER}.client.admin.keyring`

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -10,6 +10,7 @@ set -e
 : ${MDS_NAME:=mds-${HOSTNAME}}
 : ${OSD_FORCE_ZAP:=0}
 : ${OSD_JOURNAL_SIZE:=100}
+: ${OSD_GET_ADMIN_KEY:=0}
 : ${CRUSH_LOCATION:=root=default host=${HOSTNAME}}
 : ${CEPHFS_CREATE:=0}
 : ${CEPHFS_NAME:=cephfs}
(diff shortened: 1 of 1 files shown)


leseb:
Shouldn't you also call `check_admin_key` for non kv backends if `OSD_GET_ADMIN_KEY=1`?
I know it might sound stupid since the user has to explicitly enable `OSD_GET_ADMIN_KEY` so they also have to provide the admin key... Just another safety check I guess...

Ulexus:
@leseb I think that is a good idea.  By declaring `OSD_GET_ADMIN_KEY=1`, you are implying that the admin key is necessary for the operation of your OSD, so a check for it is a good idea (regardless of the KV setup).

Also, though, is there a reason to not generalize the option?  That is:  `OSD_GET_ADMIN_KEY` -> `CEPH_GET_ADMIN_KEY`.  I see it as a shoehorn, anyway.  The KV method should assure minimum operation, anyway, and get the admin key where required.

leseb:
@Ulexus +1 for generalising the name :+1:

h0tbird:
Great! Let me fix the PR with the new stuff. Let's see if I understood...

h0tbird:
I have added it to OSD and RGW.
I am not sure about the other ones (MDS, MON, RESTAPI), I think the key is added anyway.

leseb:
Thanks @h0tbird

PR #164, opened 2015-11-01
Fix docker caching: explicit ceph version

MikaelCluseau:
Hi,

Either docker caching or registry not being triggered implies that ceph version get outdated. For instance, the current versions of ceph/base:hammer is 0.94.1 while 0.94.5 is out.

This patch fixes this situation by explicitly stating the ceph package's version to use.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -8,14 +8,15 @@ MAINTAINER Sébastien Han "seb@redhat.com"
 ENV ETCDCTL_VERSION v2.0.9
 ENV ETCDCTL_ARCH linux-amd64
 ENV CEPH_VERSION hammer
+ENV CEPH_VERSION_DETAIL 0.94.5-1trusty
 
 # Install prerequisites
-RUN apt-get update &&  apt-get install -y wget
+RUN apt-get update &&  apt-get install -y wget && apt-get clean
 
 # Install Ceph
 RUN wget -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_plain;f=keys/release.asc' | apt-key add -
 RUN echo deb http://ceph.com/debian-$CEPH_VERSION/ trusty main | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list
-RUN apt-get update && apt-get install -y --force-yes ceph && \
+RUN apt-get update && apt-get install -y --force-yes ceph=${CEPH_VERSION_DETAIL} && \
 apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install etcdctl


Ulexus:
I like the explicit versioning, but where is the canonical list of available versions?

MikaelCluseau:
```
apt-cache policy ceph
```
for instance :-) With the apt source loaded, though, you can still consult things like `http://download.ceph.com/debian-hammer/dists/trusty/main/binary-amd64/Packages`. I'm pretty there's something to automate with a script like this one:
```
$ for release in firefly hammer testing; do curl -s http://download.ceph.com/debian-$release/dists/trusty/main/binary-amd64/Packages |grep ^Version |head -n1 |awk "{print \"$release\", \$2}"; done
firefly 0.80.10-1trusty
hammer 0.94.5-1trusty
testing 9.1.0-1trusty
```

MikaelCluseau:
If think this should work:
```
for release in firefly hammer; do
  git checkout $release &&
  v=$(curl -s http://download.ceph.com/debian-$release/dists/trusty/main/binary-amd64/Packages |awk '/^Package:/ {package=$2} /^Version:/ {if (package == "ceph") { print $2; exit}}') &&
  sed -i "/^ENV CEPH_VERSION_DETAIL/s/ [^ ]*\$/ $v/" base/Dockerfile &&
  git commit -am "Version bump to $v" &&
  git push
done
```

Ulexus:
So, while I like the versioned approach, and I think it would be really good to use it for explicit tagged releases, I don't see that _effectively_, we are not going to be at the latest version in the present setup.    The autobuilds on Docker Hub don't cache, so far as I have ever been able to discover, so the package lists should always be reasonably up-to-date. 

Indeed, in checking the present version, we are in sync with the latest available hammer package.  @MikaelCluseau Can you give me a counter-example of a version miss?

MikaelCluseau:
Weird. Here's what I have:
```
CoreOS stable (766.4.0)
ceph-1 ~ # docker run --rm ceph/base:hammer ceph -v
ceph version 0.94.1 (e4bfad3a3c51054df7e537a724c8d0bf9be972ff)
ceph-1 ~ # docker pull ceph/base:hammer
Pulling repository ceph/base
8e1e5444f5d4: Download complete 
e118faab2e16: Download complete 
7e2c5c55ef2c: Download complete 
e04c66a223c4: Download complete 
fa81ed084842: Download complete 
70e8f3fe0f56: Download complete 
0ea2cad8d48a: Download complete 
bf0ae85114ef: Download complete 
dd1f204e5acd: Download complete 
20d62160b5e3: Download complete 
1521d4b74583: Download complete 
e11a0a4d0a83: Download complete 
a2b623befb04: Download complete 
fbcfa17cd925: Download complete 
58271642c4b2: Download complete 
cb371b94bbdb: Download complete 
Status: Image is up to date for ceph/base:hammer
ceph-1 ~ # docker run --rm ceph/base:hammer ceph -v
ceph version 0.94.1 (e4bfad3a3c51054df7e537a724c8d0bf9be972ff)
```

Ulexus:
Ahh.  That is because Docker Hub finally fixed their "build all on any trigger" issue.  That means we'll have to actually push to the 'hammer' branch when an update is required.

Ulexus:
I just kicked off a build of the 'hammer' branch; it should be updating soon.

MikaelCluseau:
Thanks :-) But it's manual work, do want me to propose a script to check versions on every branch, including the explicit versionning you seemed to find interesting? Something that would tag the individual version + bump the branches.

Ulexus:
It's a good idea;  I certainly wouldn't turn it down.

MikaelCluseau:
ahhh good news :) thanks

PR #169, opened 2015-12-03
Add deprecation notices for discrete daemons

Ulexus:
Add deprecation notices for:
 - `ceph/osd`
 - `ceph/mon`
 - `ceph/mds`
 - `ceph/radosgw`

in favor of `ceph/daemon`

This doesn't _remove_ them.  It simply updates the documentation to mark the discrete daemon containers as deprecated.


Ulexus:
Prerequisite for #94

PR #178, opened 2015-12-07
Typo in MDS keyring path

nazar-pc:
I've got:
```
chown: cannot access '/var/lib/ceph/mds/mds-4ecf90c279f5/keyring': No such file or directory
```
When started MDS and I believe this is a typo that left after some previous paths refactoring.


streamnsight:
you might want to check out my fork as i went through lots of issues similar to yours. See PR just submitted.

Ulexus:
Thanks!

PR #179, opened 2015-12-08
fix osd_directory init, create osd in daemon + populate default values with KV store

streamnsight:
fix osd_directory init, create osd in daemon + populate default values with KV store

Please comment if you find any obvious issue. 
It seems to work well for me this way starting from a black slate, but I still have to work on the next steps.

population of defaults is done every time; may not be necessary, but it uses 'cas' command so it only writes if the key is not present.

the osd_directory assumes nothing in the `/var/lib/ceph/osd` dir, and creates 1 OSD per host/folder, since creating multiple OSD on the same physical drive leads to over estimation of the size and whatever other consequences.
if there is no OSD, it is created, and initialized.
if there is an OSD it checks if it is initialized and init if not.
or just run if it's good to go.

### DIFF
--- a/daemon/README.md
+++ b/daemon/README.md
@@ -22,6 +22,16 @@ You can use this container to bootstrap any Ceph daemon.
 * `HOSTNAME` is the hostname of the machine  (DEFAULT: $(hostname))
 
 
+SELinux
+-------
+
+If SELinux is enabled, run the following commands:
+
+```
+$ sudo chcon -Rt svirt_sandbox_file_t /etc/ceph
+$ sudo chcon -Rt svirt_sandbox_file_t /var/lib/ceph
+```
+
 KV backends
 -----------
 
@@ -105,6 +115,7 @@ If the operator does not specify an `OSD_TYPE` autodetection happens:
 Without KV backend:
 ```
 $ sudo docker run -d --net=host \
+--pid=host \
 --privileged=true \
 -v /etc/ceph:/etc/ceph \
 -v /var/lib/ceph/:/var/lib/ceph/ \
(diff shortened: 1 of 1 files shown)


nazar-pc:
My impression is that there are too many things in single PR. It would be easier to discuss if it is more granular.

streamnsight:
sorry if that seems like a lot of commits. I didn't want to squash the history but i mostly work offline and push to test. check the changed files.

nazar-pc:
I do not only mean number of commits, it touches various things: it fixes some bugs and introduces various changes, including backward incompatible with current image.

streamnsight:
Understood. 
it's aiming at more automation, vs. having to do several steps outside of the container.
what do you suggest?

nazar-pc:
A good way of doing PRs is to separate things that are not connected into several PRs.
For instance, purely bug fixes might be in one PR. One feature in second PR. Second feature in third PR and so on.
Otherwise we'll stuck discussing everything together and if maintainers don't like any specific change - they'll decline the entire PR.

streamnsight:
OK. I'm all about making it easier. 
How would I do this now I pushed my changes altogether? 

It's not like one split the tasks when coding: i fix bugs as I see them, as I move on to making things easier. I can go back and aggregate the changes into categories (bugs, improvements..) but sounds like a lot of re-testing. 
Is there an easier way?

nazar-pc:
I don't think there is an easy way now.
Basically you can try to copy branch into new one, then run rebase against upstream master and during rebase skip unnecessary commits, but you'll need to test it afterwards anyway.
Eventually you can either submit new PRs and close this one or update branch of this PR by doing `git push --force` into it (be careful here).

PR #180, opened 2015-12-08
path bug fixes for ceph-osd and ceph mds

streamnsight:


daemon/config.kv.sh
```diff
@@ -9,6 +9,16 @@ function get_mon_config {
 
   CLUSTER_PATH=ceph-config/${CLUSTER}
 
+  # making sure the root dirs are present for the confd to work
+  kviator --kvstore=${KV_TYPE} --client=${KV_IP}:${KV_PORT} put ${CLUSTER_PATH}/auth/_ _ > /dev/null 2>&1
```
Ulexus: Are these `_` characters used as placeholders?  Is that actually required?

streamnsight:
Is it required?: I believe so: when the KV store is not populated, the confd script fails. Since on starting the first monitor there is nothing on the KV store, it fails all the time. Even after populating with the KV store data generated by the monitor, it still fails because osd, mds and rgw are missing...
So, can you think of a better solution? I have tried to use 'tricks' to get the confd script to check for empty keys but it still fails; etcd has no simple way to test for empty dirs or missing path.

daemon/config.kv.sh
streamnsight: PS: _ doesn't actually show as a key in the store when listing them. it does come up on a get call but does not show in the confd file

I tried this trick: https://github.com/kelseyhightower/confd/issues/273#issuecomment-149900330 but that didn't help: it still confused confd when path did not exist.

streamnsight:
PS: this might be a etcd issue only, I don't use consul, but the error happens with confd though, which is not able to generate the ceph.conf because top directory don't exist and it's scanning through those to build the different [blocks] 

Even if the KV store is pre-populated, the `mon_host` key is still missing for confd to build the config.
One should still be able to run a monitor without pre-populating default settings, although I am not sure how the settings are applied (on start, or dynamically?)
If it's on start then NOT having those keys would be a way to prevent bootstrapping without essential defaults, but I would think Ceph uses `defaults` of its own when not provided any, right?

Ulexus:
Yeah, confd is pretty picky about existence, and while it has separate checks, it is tedious to use.  This is why I think we would be better off using a custom "confd"-type implementation.  I was sure I had seen a multi-backend KV interface library for Go, but I can't seem to find it, now.

streamnsight:
kviator is the multi backend KV library... confd is the config tool to build dynamic config files from the KV store. confd seems to be multi KV but etcd for example does nto support checking for empty dir, so you can't do a {{ if exist /dir }} on confd; it doesn't work for etcd.

Ulexus:
No, kviator is not a library; it just the CLI tool.   However, it does use the library I was thinking of: [libkv](https://github.com/docker/libkv).

streamnsight:
nice.

streamnsight:
btw: do you know how these defaults are used? the ceph.conf is updated by confd on KV store update, but what does Ceph do then? Does it update on ceph.conf update or does that only happen on starting the related daemon?

Ulexus:
I'm fairly certain it only reads `ceph.conf` on daemon startup.

streamnsight:
ok.. so what do you want me to do with this PR?

Ulexus:
If you're up to that task, that would be fantastic.  I know I won't have time any time soon.

Ulexus:
I guess the only thing I'm quibbling over are the `_` pieces.  It seems, to me, that it's a hacky fix which takes advantage of an oddity of operation.  I'm afraid, for instance, that even if the `_` fields are ignored by etcd, they may break things with consul or zookeeper.

streamnsight:
Hi,

So i spend a bunch of time trying to see how to fix the `confd` call without pre-populating the directories for `etcd`, but nothing does: `confd` first fails on reading the TOML file because the paths don't exists, and doesn't even get to parsing the template.
I actually don't understand why it works with consul since the typical response for a dir not found should be an error... 

So, I thought maybe I could use the `mkdir` command with kviator to pre-populate instead of placeholders, but `kviator` does not support it. 

So, since this seems to be only an issue with `etcd` as @Ulexus says it works on Consul without this change, I decided the best to do was to use `etcd` directly when the KV store is `etcd`. We have `etcdctl in the image`.

I hope it's a reasonable compromise.
Cheers

Does that sound reasonable? No more 'hack' in the sense there is no _ keys as placeholders that may fails on other KV stores, and it is only implemented for etcd KV store.


Note, during my research here I found out that confd was using a deprecated etcd client, so it may cause troubles in the future. Someone says they're working on porting to the new client, but that was back

Ulexus:
Yeah, I got the feeling from Kelsey last month that he didn't really want to maintain the tool  (confd).  Someone else may pick it up, but I would actually prefer to just phase its use out and generate the configurations and drop the keys in place more directly.

I'm fine with the separate `etcd` path.  `confd` and `kviator` are convenient abstraction tools, but I'm fine with breaking out support into different pathways, too.

streamnsight:
@Ulexus @leseb 
Can we merge this?

leseb:
@streamnsight sorry for the lag, I'm good with this one, will let @Ulexus merge this.

Ulexus:
Thanks, @streamnsight

PR #181, opened 2015-12-08
Populate kv

streamnsight:
populate KV store on start of daemon: only add key if not present. 
also allow to use a file in /etc/ceph/ as populate script


streamnsight:
1st 2 commits are part of my first PR (bug fixes)

PR #182, opened 2015-12-08
Populate KV entrypoint. Only add key if not present + allow loadi…

streamnsight:
Populate KV entrypointt. Only add key if not present + allow loading defaults config file from /etc/ceph/ folder if needed


daemon/entrypoint.sh
```diff
@@ -65,6 +65,11 @@ chown ceph. /var/run/ceph
 # based on the data store
 case "$KV_TYPE" in
    etcd|consul)
+      if [[ ! -e "/etc/ceph/populate.sh" ]]; then
```
Ulexus: This seems like a strange way to phrase this.  I would prefer a direct check for the file to be sourced.  If the intent is to allow an override of the populate script, it should be:
  - commented as such in the script
  - phrased directly (if file exists, versus if file does _not_ exist)
  - documented in the associated README

daemon/Dockerfile
```diff
@@ -21,6 +21,9 @@ RUN mkdir -p /etc/container_environment
 ADD 2.sh /etc/runit/2
 RUN chmod +x /etc/runit/2
 
+ADD populate.sh /populate.sh
```
Ulexus: populate.sh doesn't exist here, so this ADD statement will fail

daemon/entrypoint.sh
```diff
@@ -65,6 +65,11 @@ chown ceph. /var/run/ceph
...
+        source /populate.sh
```
Ulexus: I'm a bit hesitant to just directly execute the populate script without any modifications.  Perhaps it is safe, but I seem to recall it not being suitable for general use, with various assumptions made which may not be good practice.

Ulexus:
@streamnsight I don't meant to be dismissive, despite the comments.  I appreciate the PRs

daemon/Dockerfile
streamnsight: added the script file. I had merged everything in one PR originally and have been reworking it all to split bug fixes and features.

streamnsight:
Ideally all params should be ENV variables, no script: that way it is safe.
The script can then translate to KV store with Kviator, but 

1) I don't know what syntax Ceph uses for those ENV, is there any doc on all these 'defaults' values and how they are set when there is no KV backend?
2) this is a lot of tedious manual rewrite and that was much easier, yes: i just need things to work; to me better work but look hacky than not work; there is always later to make it look nice.
3) It is needed to make things work properly, and it is not straight forward to run the populate script on coreOS unless it is ran in a kviator image.

I added the missing populate.sh script.

PS: I had sent a full PR with all my changes, including setting up a osd_directory in one shot with no external manual commands, but I was told it is too much and confusing. I reversed everything and split the changes into bug fixes and features. I didn't push all of it; there is more to come as the features build on the bug fixes, and I can't do another PR with the features without the bug fixes.

Seems like this project is a WIP. I'm just trying to make things work. It's not perfect but better than not working.

daemon/entrypoint.sh
```diff
@@ -65,6 +65,13 @@ chown ceph. /var/run/ceph
 # based on the data store
 case "$KV_TYPE" in
    etcd|consul)
+      # if populate.sh script found in /etc/ceph/ use that
+      if [[ -e "/etc/ceph/populate.sh" ]]; then
```
h0tbird: What about having a `KV_POPULATE=(true|false)` option in order to enable/disable this feature.
I already populate my KV store outside the container and I don't want to add extra pressure to `etcd` by scanning all the keys again.

streamnsight:
is that better? 
now a specific entrypoint so it is backward compatible.
use a text file for config without all the bash clutter and safer since it's not a script

Ulexus:
@streamnsight I quite like this, thank you.  I'd like another set of eyes on it, just in case.

@leseb ?

streamnsight:
@leseb on peut merger? ;-)

daemon/entrypoint.sh
```diff
@@ -72,6 +72,38 @@ case "$KV_TYPE" in
...
+       etcd|consul)
+          # if ceph.defaults found in /etc/ceph/ use that
+          if [[ -e "/etc/ceph/ceph.defaults" ]]; then
+            DEFAULTS_PATH="/etc/ceph/ceph.defaults"
+          else
+          # else use defaults
+            DEFAULTS_PATH="/ceph.defaults"
+          fi
+          # read defaults file, grab line with key<space>value without comment #
+          cat "$DEFAULTS_PATH" | grep '^.* .*' | grep -v '#' | while read line; do
```
leseb: Cosmetic option :)

`cat "$DEFAULTS_PATH" | awk '/.* .*/ && !/#/' | while read line; do`

leseb:
Looks really good to me, just left one cosmetic comment but we can easily go without it :). The rest is good! Thanks @streamnsight

PR #184, opened 2015-12-08
create osd, init and run in one shot, without the need to run ceph os…

streamnsight:
this changes the logic on osd_directory endpoint

instead of having to run `ceph osd create` on then monitor, then create a directory manually on the host, chown it etc...  before running the daemon, this is done in the entrypoint to start an OSD in osd_directory mode

the original logic was looking for osd ids in the `/var/lib/ceph/osd/` folder however it doesn't make sense to have multiple OSD directories on the same host/drive as each will report full disk size as their own.
Here we automate creation of the OSD (no more manual steps) and prevent creation of multiple osd_directory at the same time.

this may be a breaking change if anyone was actually running multiple osd_directories on the same host, but my tests show it doesn't work well, so it must not be used that way.

PS: this PR goes along with the other PRs to populate KV store and bug fixes. it does not work well without those fixes.

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -188,10 +188,24 @@ function start_osd {
 #################
 
 function osd_directory {
-  if [[ ! -d /var/lib/ceph/osd || -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then
-    echo "ERROR- could not find any OSD, did you bind mount the OSD data directory?"
-    echo "ERROR- use -v <host_osd_data_dir>:<container_osd_data_dir>"
+  if [[ ! -d /var/lib/ceph/osd ]]; then
+    echo "ERROR- could not find the osd directory, did you bind mount the OSD data directory?"
+    echo "ERROR- use -v <host_osd_data_dir>:/var/lib/ceph/osd"
     exit 1
+  fi
+
+  # make sure ceph owns the directory
+  chown ceph. /var/lib/ceph/osd
+
+  # check if anything is there, if not create an osd with directory
+  if [[ -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then
+    echo "Creating osd with ceph osd create"
+    OSD_ID=$(ceph osd create)
+    echo "OSD created with ID: ${OSD_ID}"
+    # create the folder and own it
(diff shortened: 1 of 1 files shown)


daemon/entrypoint.sh
```diff
@@ -188,12 +188,26 @@ function start_osd {
...
-    echo "ERROR- could not find any OSD, did you bind mount the OSD data directory?"
-    echo "ERROR- use -v <host_osd_data_dir>:<container_osd_data_dir>"
+  if [[ ! -d /var/lib/ceph/osd ]]; then
+    echo "ERROR- could not find the osd directory, did you bind mount the OSD data directory?"
+    echo "ERROR- use -v <host_osd_data_dir>:/var/lib/ceph/osd"
     exit 1
   fi
 
+  # make sure ceph owns the directory
+  chown ceph. /var/lib/ceph/osd
```
Ulexus: This should probably be a recursive chown, and I don't see a reason not to set the group, as well (though perhaps that is too invasive):  `chown -R ceph:ceph /var/lib/ceph/osd`

daemon/entrypoint.sh
```diff
@@ -188,12 +188,26 @@ function start_osd {
...
+
+  # check if anything is there, if not create an osd with directory
+  if [[ -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then
+    echo "Creating osd with ceph osd create"
+    OSD_ID=$(ceph osd create)
```
Ulexus: What happens here if the bootstrap or admin key isn't found?  We probably need a failure check here.

daemon/entrypoint.sh
leseb: actually `ceph.` = `ceph:ceph`

Ulexus: Interesting;  I've never seen that abbreviation... I see it mentioned in the man page, though.  It's smarter than that, even; it implies the primary group of that user's login, not necessarily the same name.  Nice.

leseb: :)

daemon/entrypoint.sh
streamnsight: added the below to catch if OSD_ID is not a number (and some other error was returned)

Ulexus:
lgtm

leseb:
LGTM

PR #185, opened 2015-12-08
mon-data added to ceph-mon mkfs.

a-nldisr:
When you configure the clustername with a non-default value (default: ceph) the current ceph-mon mkfs creates a directory under /var/lib/ceph/mon/ceph-${hostname}. When the exec (line 141) for ceph-mon starts it looks inside the directory: /var/lib/ceph/mon/${CLUSTER}-${HOSTNAME}. Result is that the cluster will never start with a non-default clustername

The mkfs populates the directory correct when you use --mon-data.

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -131,7 +131,7 @@ function start_mon {
     create_socket_dir
 
     # Prepare the monitor daemon's directory with the map and keyring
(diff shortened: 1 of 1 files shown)


Ulexus:
@a-nldisr Thanks; I was not aware of this.

PR #187, opened 2015-12-16
Fix to run ceph/daemon osd on Centos.

Deepthidharwar:
From entrypoint.sh, 'ceph-disk activate' call is made to 
activate the disk and later start the ceph-osd process
as a daemon. Currently as systemd is not initalised
as part of base Centos image, this seizes to run and errors out.

Also, another call is made from entrypoint.sh after the 
initialization to kick start ceph osd process in the 
foreground for the ceph-osd docker container to continuously run.
As a result, there are two calls made to run ceph-osd and
this will result in the container erroring out and exiting.
This patch aims to fix the same.

Signed-off-by: Deepthi Dharwar <ddharwar@redhat.com>

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -265,11 +265,13 @@ function osd_disk {
     ceph-disk -v prepare ${OSD_DEVICE}
   fi
 
-  ceph-disk -v activate ${OSD_DEVICE}1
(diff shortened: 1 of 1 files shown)


leseb:
LGTM :+1:

Deepthidharwar:
My bad .. Missed out on the second call. Shall send out new patch.

PR #202, opened 2016-01-20
since ceph-disk activate already starts ceph-osd, don't rerun it; just wait till it exits

rootfs:
@leseb

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -354,19 +354,15 @@ function osd_disk {
     chown ceph. ${OSD_DEVICE}2
   fi
 
-  ceph-disk -v activate ${OSD_DEVICE}1
+  ceph-disk -v activate --mark-init none ${OSD_DEVICE}1
   OSD_ID=$(cat /var/lib/ceph/osd/$(ls -ltr /var/lib/ceph/osd/ | tail -n1 | awk -v pattern="$CLUSTER" '$0 ~ pattern {print $9}')/whoami)
   OSD_WEIGHT=$(df -P -k /var/lib/ceph/osd/${CLUSTER}-$OSD_ID/ | tail -1 | awk '{ d= $2/1073741824 ; r = sprintf("%.2f", d); print r }')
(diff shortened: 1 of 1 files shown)


leseb:
@rootfs we had a similar fix for the centos branch, what do you think about it? https://github.com/ceph/ceph-docker/pull/188/files

rootfs:
that's brutal, but works :)

leseb:
Yes so we don't have to wait and using mark_init none helps with containers too.

leseb:
If you are ok, do you mind using the same code so we keep things consistent? :)

rootfs:
done, thanks.

leseb:
@rootfs thanks! @Ulexus any objection?

Ulexus:
I agree with @rootfs that looks brutal... I don't know the problem, but that seems aggressive

rootfs:
@Ulexus. The problem is that `ceph-disk activate`  spawns `ceph-osd`. So when `entrypoint.sh` continues to `exec ceph-osd`, `ceph-osd` just fails and complains there is another osd daemon running. 

So the solution space left for us is either kill the ceph-osd that `ceph-disk activate` spawns, or keep that osd daemon and watch it.

Ulexus:
Yeah, I see.  I wish I had a better option.  We could monitor that process and exit coincidentally, but that doesn't really seem any better.  `ceph-disk` will not will any accolades from me, but if this is what it takes and if the `ceph-osd` process dies gracefully, then we might as well do it.  (There should have been plenty of Eeyore tones in there, if you didn't pick them up.)

leseb:
@Ulexus so let's go with the "brutal" way? :)

Ulexus:
Hah.  Yes, let's! :)

rootfs:
sorry late to the part. Yes, `killall` is brutal. I just found a couple of issues
- `killall` is not installed by default on centos container (might be the same for ubuntu)
- worse, `killall -9 -w` also kills other osd.

So the watch and wait (the first commit) seems a nicer/safer way to me.

leseb:
Oops... yes  indeed, let me revert this change.

leseb:
@rootfs do you mind re-submitting your patch? :) Thanks!

rootfs:
okie dokie

PR #203, opened 2016-01-25
Fix for issue #171

fmeppo:
This pull provides a patch to ceph-disk, allowing the permissions to be overridden.  A disk-type OSD will then run with perms ceph:disk, so it can get raw block dev access for the journal file.

### DIFF
--- a/daemon/Dockerfile
+++ b/daemon/Dockerfile
@@ -7,7 +7,7 @@
 FROM ceph/base
 MAINTAINER Sébastien Han "seb@redhat.com"
 
-# patch is only needed until PR #7531 is merged in ceph/ceph (and backported
+# patch is only needed until PR #7351 is merged in ceph/ceph (and backported
 # to the version used in ceph/ceph-docker)
 RUN apt-get update && apt-get -y install runit && \
     apt-get install patch && \
--- a/daemon/Dockerfile
+++ b/daemon/Dockerfile
@@ -8,7 +8,8 @@ FROM ceph/base
 MAINTAINER Sébastien Han "seb@redhat.com"
 
 RUN apt-get update && apt-get -y install runit && \
-apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    apt-get install patch && \
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Add bootstrap script
 ADD entrypoint.sh /entrypoint.sh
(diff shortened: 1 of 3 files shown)


Ulexus:
This seems like a good idea, and I appreciate the effort.  I'd rather see this patch go to the upstream ceph-disk rather than maintaining a local patch.   Have you tried there yet?

fmeppo:
The upstream project will get a pull request just as soon as I'm done
fixing the man page for ceph-disk. :-)

Figured I'd at least get something to ceph-docker in the meantime, since
anything I push upstream may not make it to infernalis...

--
Mike Shuey

Ulexus:
Sounds good.  I'm a bit concerned with the local patch adding brittleness, but at least that should fail at image-build time, if it doesn't patch cleanly.

@leseb , thoughts?

leseb:
Same concern here, I'd prefer to have this getting merged in `ceph-disk` upstream first. However I understand that this might need some time to get through.
In the meantime, please link the PR on ceph upstream to fix this and add a message to explain that the patch needs to go away as soon as it's merged and backported :)

Thanks!

fmeppo:
This patch has been rebased into upstream, and there's a pull request (#7351) in queue.  Also added a few comments about that, indicating it can be removed after the PR is done.

daemon/Dockerfile
```diff
@@ -7,8 +7,11 @@
 FROM ceph/base
 MAINTAINER Sébastien Han "seb@redhat.com"
 
+# patch is only needed until PR #7531 is merged in ceph/ceph (and backported
```
leseb: s/7531/7351/

leseb:
@fmeppo thanks for getting back to us on this. Just fix the PR ID in the code and we should be good to go. Right @Ulexus?

fmeppo:
Yeah...sorry about that.  Comment fixed.

daemon/Dockerfile
```diff
@@ -7,8 +7,11 @@
 FROM ceph/base
 MAINTAINER Sébastien Han "seb@redhat.com"
 
+# patch is only needed until PR #7351 is merged in ceph/ceph (and backported
+# to the version used in ceph/ceph-docker)
 RUN apt-get update && apt-get -y install runit && \
-apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    apt-get install patch && \
```
leseb: just to be sure, shouldn't we use `-y` in apt?

fmeppo: It works as-is, but....you're probably right.  Committed a fix; it's in the
pull request now.



--
Mike Shuey

leseb:
Thanks @fmeppo!

PR #205, opened 2016-01-29
Fix READMEs for 'centos' branch

dpkshetty:
'centos' branch READMEs still carry Ubuntu as the OS name
causing confusing for readers visiting github, thus fixing
it here

Signed-off-by: Deepak C Shetty <deepakcs@redhat.com>


leseb:
Thanks @dpkshetty

PR #207, opened 2016-02-01
since ceph-disk activate already starts ceph-osd, don't rerun it; just wait till it exists

rootfs:
@leseb

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -359,7 +359,14 @@ function osd_disk {
   OSD_WEIGHT=$(df -P -k /var/lib/ceph/osd/${CLUSTER}-$OSD_ID/ | tail -1 | awk '{ d= $2/1073741824 ; r = sprintf("%.2f", d); print r }')
   ceph ${CEPH_OPTS} --name=osd.${OSD_ID} --keyring=/var/lib/ceph/osd/${CLUSTER}-${OSD_ID}/keyring osd crush create-or-move -- ${OSD_ID} ${OSD_WEIGHT} ${CRUSH_LOCATION}
 
-  exec /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID} --setuser ceph --setgroup ceph
+  # ceph-disk activiate has exec'ed /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID}
+  # wait till ceph-osd exit
+  OSD_PID=$(ps -ef |grep ceph-osd |grep osd.${OSD_ID} |awk '{print $2}')
+  if [ ! -z ${OSD_ID} ]; then
+      while [ -e /proc/${OSD_PID} ]; do sleep 1;done
+  else
+      exec /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID} --setuser ceph --setgroup ceph
+  fi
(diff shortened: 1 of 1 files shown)


daemon/entrypoint.sh
```diff
@@ -359,7 +359,14 @@ function osd_disk {
   OSD_WEIGHT=$(df -P -k /var/lib/ceph/osd/${CLUSTER}-$OSD_ID/ | tail -1 | awk '{ d= $2/1073741824 ; r = sprintf("%.2f", d); print r }')
   ceph ${CEPH_OPTS} --name=osd.${OSD_ID} --keyring=/var/lib/ceph/osd/${CLUSTER}-${OSD_ID}/keyring osd crush create-or-move -- ${OSD_ID} ${OSD_WEIGHT} ${CRUSH_LOCATION}
 
-  exec /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID} --setuser ceph --setgroup ceph
+  # ceph-disk activiate has exec'ed /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID}
+  # wait till ceph-osd exit
+  OSD_PID=$(ps -ef |grep ceph-osd |grep osd.${OSD_ID} |awk '{print $2}')
+  if [ ! -z ${OSD_ID} ]; then
+      while [ -e /proc/${OSD_PID} ]; do sleep 1;done
```
leseb: I'm a little bit sad that we won't have any output of the OSD...

rootfs: what do you want to see :)

leseb: Stdout of the OSD process? :)
Just checked  `/proc/${OSD_PID}/fd/0` and it redirects to `/dev/null` :(

daemon/entrypoint.sh
```diff
@@ -359,7 +359,15 @@ function osd_disk {
   OSD_WEIGHT=$(df -P -k /var/lib/ceph/osd/${CLUSTER}-$OSD_ID/ | tail -1 | awk '{ d= $2/1073741824 ; r = sprintf("%.2f", d); print r }')
   ceph ${CEPH_OPTS} --name=osd.${OSD_ID} --keyring=/var/lib/ceph/osd/${CLUSTER}-${OSD_ID}/keyring osd crush create-or-move -- ${OSD_ID} ${OSD_WEIGHT} ${CRUSH_LOCATION}
 
-  exec /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID} --setuser ceph --setgroup ceph
+  # ceph-disk activiate has exec'ed /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID}
+  # wait till ceph-osd exit
```
leseb: Are you sure that the process will ever exit? If it actually does then the container will exit as well, so it has to stay alive :). Or when it dies run a new one. wdyt?

rootfs: Right. The idea is to stay sleep and exit, when either the process is killed or container is stopped.

leseb: Ok I got mislead by the `wait till ceph-osd exit` statement, this is not the `ceph-osd` but the container itself.
Can you make this clearer please?

Then we should be good to go :)

rootfs:
cool, done,

PR #215, opened 2016-02-10
Add an appropriate default option for BTRFS

fmeppo:
As-is, BTRFS filesystems can't always be created for OSDs - pre-existing data won't be appropriately wiped.  Turns out -f was specified for XFS filesystems, so do the same thing for BTRFS ones.


leseb:
Seems legit! Thanks @fmeppo

PR #216, opened 2016-02-12
split osd disk to disk prepare and disk activate

rootfs:
@leseb


leseb:
@rootfs even though we discussed that on irc, a little commit message is welcome :)

rootfs:
The idea is that when prepare and activate are split, it is possible to prepare the osd first, leaving osd activate to systemd. Then systemd activates osd without having to prepare again.

leseb:
Thanks @rootfs 
I'm good with this one, @Ulexus?

Ulexus:
I like this, yes.  It's a good way to split out the process so that fixes (like permissions) can be injected.  LGTM.

Ulexus:
It should probably have some documentation added to the README also.

rootfs:
@Ulexus thanks, just updated `README.md`

Ulexus:
Thanks, @rootfs

PR #217, opened 2016-02-12
add systemd unit files for OSes like REHL Atomic

rootfs:
@leseb @Ulexus


Ulexus:
I do usually use `docker kill` for this instead of `docker stop`, but `stop` is probably better for general use.

Issue #220, opened 2016-02-17
kviator doesn't store/retrieve binary data safely


fmeppo:
kviator is used to store a bunch of useful pieces in various KV stores - but it seems to only handle ascii files (at least, that's the case with the etcd version I'm using).  Keyrings are fine, but the monmap is a binary file.  config.kv.sh stores monmap via kviator, and a corrupt version will be retrieved later.  Looks like binaries like monmap must be uuencoded first (and uudecoded on extraction).

Ulexus:
Wow, you're right.  kviator is using string values instead of byte slices.  (The underlying libkv does use byte slices, though, which should be binary safe.)  This is, indeed, a problem for binary storage like the monmap.

fmeppo:
Yup.  See pull request #221 for a fix.  Tested on my local kit, and it seems to work (in that my replicated MONs are now coming up, using a monmap from etcd).

CoRfr:
I think there is an issue with this patch for existing installations.
One of my monitors restarted an pulled this image, it was not able to use the existing monmap and therefore crashed, also leaving the configuration locked.

I'm seeing the following error:
```
Feb 18 17:46:44 carmd-ev-lxcim2 docker[9861]: 2016-02-18T17:46:44Z carmd-ev-lxcim2 confd[108]: INFO Target config /etc/ceph/ceph.conf out of sync
Feb 18 17:46:44 carmd-ev-lxcim2 docker[9861]: 2016-02-18T17:46:44Z carmd-ev-lxcim2 confd[108]: INFO Target config /etc/ceph/ceph.conf has been updated
Feb 18 17:46:44 carmd-ev-lxcim2 docker[9861]: Adding Keyrings
Feb 18 17:46:44 carmd-ev-lxcim2 docker[9861]: Monmap is missing. Adding initial monmap...
Feb 18 17:46:44 carmd-ev-lxcim2 docker[9861]: uudecode fatal error:
Feb 18 17:46:44 carmd-ev-lxcim2 systemd[1]: Stopped docker container f92bfe8222290811b4777abf0cfbee2b809251c75cc49459a3b25d96411fba76.
```

I think the monmap with this new format could potentially be stored elsewhere, like monmap.enc or something alike, and there should be some code for the transition.
Something like:
```
Get monmap.enc
- OK => uudecode => done
- Not OK => Get monmap => Convert monmap to monmap.enc => done
```

CoRfr:
[...]

PR #221, opened 2016-02-17
properly store binaries with kviator

fmeppo:
kviator can't store binary data (see ceph/ceph-docker #220 ).  This works around that limit via sharutils and uuencoding/decoding the monmap.

### DIFF
--- a/daemon/Dockerfile
+++ b/daemon/Dockerfile
@@ -9,8 +9,7 @@ MAINTAINER Sébastien Han "seb@redhat.com"
 
 # patch is only needed until PR #7351 is merged in ceph/ceph (and backported
 # to the version used in ceph/ceph-docker)
-RUN apt-get update && apt-get -y install runit && \
-    apt-get -y install patch sharutils && \
+RUN apt-get update && apt-get -y install runit patch sharutils && \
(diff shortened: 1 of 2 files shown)


daemon/Dockerfile
```diff
@@ -10,7 +10,7 @@ MAINTAINER Sébastien Han "seb@redhat.com"
 # patch is only needed until PR #7351 is merged in ceph/ceph (and backported
 # to the version used in ceph/ceph-docker)
 RUN apt-get update && apt-get -y install runit && \
-    apt-get -y install patch && \
+    apt-get -y install patch sharutils && \
```
leseb: can you integrate the package in the previous so we don't call`apt-get` twice? (unless you have a good reason).

leseb:
Just a minor comment other than that LGTM.
Thanks @fmeppo!

daemon/Dockerfile
leseb: oh wait, why isn't this part of base instead?

fmeppo:
Cleaned up, should be good.

leseb:
I know it's related to your patch and we can fix that later but I still wonder why we have installation step into the entrypoint.
We can fix that another PR though.

Thanks!

PR #224, opened 2016-02-25
README.md update for monitors

fmeppo:
Explain why monitors should be started with a volume for /var/lib/ceph.


Ulexus:
Thanks!  I don't know how we missed this... Even in the systemd unit.

Issue #226, opened 2016-02-26
Can't use NVMe device as OSD


fmeppo:
It doesn't look like an NVMe device can be used as an OSD.  When I try to initialize /dev/nvme0n1, I get back a crash from ceph-disk (usually).  When I run the entry point manually (in an interactive shell), I'll occasionally get the ceph-disk crash.  Digging into this a bit, it looks like ceph-disk may have a race condition between probing for newly-created partitions and actually using them - but that's very preliminary.

If ceph-disk manages to not crash after creating partitions, something (probably a later command in the entry point) fails to compute the proper partition device for a subsequent chown.  By default, journals are created on the second partition on a device; in this case, that's /dev/nvme0n1p2, but chown tries to operate on /dev/nvme0n12 (which is wrong).

Ulexus:
This certainly sounds like an upstream bug of ceph-disk.  

I've actually been meaning to try out one of my new nvme disks as an OSD... not that I would have used ceph-disk for it, though.

fmeppo:
You'll get another PR soon - I'm finding some other bugs in the entrypoint around partition device naming (which may be causing ceph-disk to misbehave).

If you have test cases you'd like run on 3x Intel P3700 drives, let me know.  I'd be interested in your thoughts on ceph-docker's suitability for large production deployments (and would be happy to arrange a concall/skype/whatever if appropriate).

fmeppo:
See PR #227 .  Looks like what I initially thought was a race condition in ceph-disk may have been caused by calling ceph-disk with an incorrect block dev name (due to improperly computed partition block dev names).  This PR should clear up the issue.

PR #227, opened 2016-02-26
Compute names for partition block devices correctly

fmeppo:
Add a dev_part function to properly compute the numbered partition on the named block device, and use it throughout.  This seems to fix issue #226 .


leseb:
Looks good to me. Thanks!

PR #228, opened 2016-03-06
Clean up tools README

thecanadianbaker:
Added backtick for code highlighting
Added clarified directions on running from container

### DIFF
--- a/examples/coreos/tools/README.md
+++ b/examples/coreos/tools/README.md
@@ -1,20 +1,23 @@
 ceph-tools
 ==========
 
-These are wrapper files for ceph cli tools to make working with ceph a little
-easier and directly from the host OS.
+These are wrapper files for ceph CLI tools to make working with ceph a little
+easier and allow direct usage from the host OS.
 
 Installation
 ============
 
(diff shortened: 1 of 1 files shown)


Ulexus:
I believe this should refer to `ceph/base` instead of `ceph/tools`.  @hookenz ?

Ulexus:
Oh, I see.  This is an installation container for the binaries in question.  There is a Dockerfile also in that directory.

@hookenz let's change the name to `ceph/install-utils` or something a bit more descriptive of what the container actually does.  Then I can add it to the Docker Hub repository.

thecanadianbaker:
I will hop on that. Is there any reason not to include RBD in the utils as well?

Ulexus:
None of which I am aware, no.

hookenz:
@Ulexus - Thanks for that.  Sorry I haven't contributed for quite a long time.  We're unfortunately no longer maintaining our own hardware and are moving to Amazon so I'm not using ceph anymore.  But I'm still interested in ceph, it's a fantastic storage platform.

PR #230, opened 2016-03-07
Update CoreOS tools

thecanadianbaker:
* Cleaned up tools README
* Added rbd to loaded tools (See #228)
* Changed container name from `tools` to `install-utils` (See #228)
* Changed `FROM` for install-utils from `ceph/base` to `busybox`
* Moved startup.sh to use a `for` loop to load wrappers


hookenz:
Thanks @thecanadianbaker - long story but I will no longer be using ceph in my environment which is why I haven't contributed for some time.  I'm still interested in the project, but unlikely to maintain.  Feel free to change or remove the MAINTAINER name/email in the Dockerfile.

thecanadianbaker:
@hookenz Thanks for letting me know :) This one is simple enough I would feel comfortable taking it on.  Thanks for the contribs

hookenz:
@thecanadianbaker - You're welcome.  It was fun.  I'm curious to know if anyone is using this script that I wrote.  https://github.com/ceph/ceph-docker/blob/master/examples/common/mount-osds

There is no readme attached to it and may not be clear what it does, but some may find it useful.  It basically takes an OSD, grabs the ID and mounts it to /osd/clusterid-id.  E.g. for the default ceph cluster, osd1 would be /osd/ceph-1  The idea was being able to move OSD disks to different nodes if the node was to fail.

thecanadianbaker:
Sounds interesting. I will try to take a look and add some doc's

Ulexus:
Thanks @thecanadianbaker 

I've added the build to Docker Hub, so the image should come online shortly.

thecanadianbaker:
It is already built and up. Thanks Seán!

PR #233, opened 2016-03-10
Properly handle OSD_JOURNAL for OSD_TYPE=disk

fmeppo:
Properly handle specifying an OSD_JOURNAL when OSD_TYPE=disk.  If we specify an OSD_DEVICE and an OSD_JOURNAL with the disk type, don't attempt to partition the OSD_DEVICE - just go with what the user has specified for block devices.

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -361,12 +361,13 @@ function osd_disk_prepare {
   if [[ ! -z "${OSD_JOURNAL}" ]]; then
     ceph-disk -v prepare ${OSD_DEVICE} ${OSD_JOURNAL}
     chown ceph. ${OSD_JOURNAL}
+    ceph-disk -v --setuser ceph --setgroup disk activate ${OSD_DEVICE}
   else
     ceph-disk -v prepare ${OSD_DEVICE}
     chown ceph. $(dev_part ${OSD_DEVICE} 2)
+    ceph-disk -v --setuser ceph --setgroup disk activate $(dev_part ${OSD_DEVICE} 1)
   fi
 
-  ceph-disk -v --setuser ceph --setgroup disk activate $(dev_part ${OSD_DEVICE} 1)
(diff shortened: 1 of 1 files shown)


Ulexus:
That sounds reasonable and shouldn't break anything.  I don't use ceph-disk, though, so if prefer to have another set of eyes.  @leseb

leseb:
This looks good but I'm curious why are we still activating osds in the prepare function. We should first call prepare with a container and then activate the device with another container.

fmeppo:
So where are we with merging this?  Just curious if there's something else I should fix here....

rootfs:
@fmeppo just found out in your `osd_disk_prepare`, the osd is actviated. This is not expected to happen, since we use `osd_activate` to activate the osd.

fmeppo:
Yeah, that always seemed strange to me.  The activate call has been a part
of prepare for longer than I've been working with this project - I merely
made it work with separately named journal files, and preserved the
original behavior.


--
Mike Shuey

leseb:
That's what I said earlier as well, I think we need to rework that a little bit. Just need to find to time to get this repo right on track with tags for instance...

fmeppo:
I'm happy to help with the tagging, if I can - I just need a point in the
right direction (as I've got no idea how containers are actually built, or
how you'd want to integrate tags & IDs).


--
Mike Shuey

mainboy:
Could I place data in a whole disk and place journal to other disk? Thank you.

fmeppo:
That's why I wrote the patch.  If you specify an OSD_DISK and an
OSD_JOURNAL, the container will assume you know what's best and start up an
OSD with separate data FS and journal.


On Wednesday, March 30, 2016, Kang Yan <notifications@github.com> wrote:

> Could I place data in a whole disk and place journal to other disk? Thank
> you.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/ceph/ceph-docker/pull/233#issuecomment-203413650>
>


-- 

--
Mike Shuey

PR #234, opened 2016-03-22
Add block device journals to osd_directory type.

potto007:
When OSD directories already created, user can provide OSD_JOURNAL{OSDid1}, OSD_JOURNAL{OSDid2} and so on.

### DIFF
--- a/daemon/README.md
+++ b/daemon/README.md
@@ -129,7 +129,7 @@ Options for OSDs (TODO: consolidate these options between the types):
 * `JOURNAL_DIR` - if provided, new OSDs will be bootstrapped to use the specified directory as a common journal area.  This is usually used to store the journals for more than one OSD on a common, separate disk.  This currently only applies to the `directory` OSD type.
 * `JOURNAL` - if provided, the new OSD will be bootstrapped to use the specified journal file (if you do not wish to use the default).  This is currently only supported by the `directory` OSD type
 * `OSD_DEVICE` - mandatory for `activate` and `disk` OSD types; this specifies which block device to use as the OSD
-* `OSD_JOURNAL` - optional override of the OSD journal file. this only applies to the `activate` and `disk` OSD types
+* `OSD_JOURNAL` - optional override of the OSD journal file.
 
 ### Without OSD_TYPE ###
 
@@ -236,6 +236,7 @@ following environment variables will control its creation:
 * `WEIGHT` is the of the OSD when it is added to the CRUSH map (default is `1.0`)
 * `JOURNAL` is the location of the journal (default is the `journal` file inside the OSD data directory)
 * `HOSTNAME` is the name of the host; it is used as a flag when adding the OSD to the CRUSH map
+* `OSD_JOURNAL<osd-id>` optional - providing for each OSD directory to be used will prepare and mount the supplied block device
 
 The old option `OSD_ID` is now unused.  Instead, the script will scan for each directory in `/var/lib/ceph/osd` of the form `<cluster>-<osd-id>`.
 
@@ -258,6 +259,9 @@ There are two workarounds, at present:
 
 To run multiple OSDs within the same container, simply bind-mount each OSD datastore directory:
 * `docker run -v /osds/1:/var/lib/ceph/osd/ceph-1 -v /osds/2:/var/lib/ceph/osd/ceph-2`
+
+To run multiple OSDs (for example: osd.5 & osd.11) in the same container, with multiple journals on block devices:
+* `docker run -v /var/lib/ceph:/var/lib/ceph -v /dev:/dev -e OSD_TYPE=directory -e OSD_JOURNAL5=/dev/sdb1 -e OSD_JOURNAL11=/dev/sdb2 ...`
 
 
 #### BTRFS and journal ####
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -276,7 +276,12 @@ function osd_directory {
   fi
 
   for OSD_ID in $(ls /var/lib/ceph/osd |  awk 'BEGIN { FS = "-" } ; { print $2 }'); do
-    if [ -n "${JOURNAL_DIR}" ]; then
+    OSD_JOURNAL_TMP="OSD_JOURNAL$OSD_ID"
+    OSD_JOURNAL=${!OSD_JOURNAL_TMP}
+    if [ -n "${OSD_JOURNAL}" ]; then
+      ceph-disk -v prepare ${OSD_JOURNAL}
+      OSD_J=${OSD_JOURNAL}
+    elif [ -n "${JOURNAL_DIR}" ]; then
        OSD_J="${JOURNAL_DIR}/journal.${OSD_ID}"
        chown -R ceph. ${JOURNAL_DIR}
     else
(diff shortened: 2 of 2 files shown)


daemon/entrypoint.sh
```diff
@@ -292,7 +297,7 @@ function osd_directory {
     if [ ! -e /var/lib/ceph/osd/${CLUSTER}-${OSD_ID}/keyring ]; then
       chown ceph. /var/lib/ceph/osd/${CLUSTER}-${OSD_ID}
       # Create OSD key and file structure
-      ceph-osd ${CEPH_OPTS} -i $OSD_ID --mkfs --mkkey --mkjournal --osd-journal ${OSD_J} --setuser ceph --setgroup ceph
+      ceph-osd ${CEPH_OPTS} -i $OSD_ID --mkfs --mkkey --mkjournal --osd-journal ${OSD_J} --setuser ceph --setgroup disk
```
leseb: why do you change the group?

potto007: I made this change in order to for block device journal creation to work. After much head pounding, and scouring, I found Issue https://github.com/ceph/ceph-docker/issues/171 and PR https://github.com/ceph/ceph-docker/pull/233. Changing group to "disk" allowed the ceph-disk operations to complete successfully.

leseb: Right forgot about those. Thanks.

daemon/README.md
```diff
@@ -236,6 +236,7 @@ following environment variables will control its creation:
 * `WEIGHT` is the of the OSD when it is added to the CRUSH map (default is `1.0`)
 * `JOURNAL` is the location of the journal (default is the `journal` file inside the OSD data directory)
 * `HOSTNAME` is the name of the host; it is used as a flag when adding the OSD to the CRUSH map
+* `OSD_JOURNAL<osd-id>` optional - providing for each OSD directory to be used will prepare and mount the supplied block device
```
leseb: perhaps we can highlight that this only works for the osd directory scenario? or at least make it a bit clearer?

daemon/entrypoint.sh
```diff
@@ -276,7 +276,12 @@ function osd_directory {
   fi
 
   for OSD_ID in $(ls /var/lib/ceph/osd |  awk 'BEGIN { FS = "-" } ; { print $2 }'); do
-    if [ -n "${JOURNAL_DIR}" ]; then
+    OSD_JOURNAL_TMP="OSD_JOURNAL$OSD_ID"
+    OSD_JOURNAL=${!OSD_JOURNAL_TMP}
+    if [ -n "${OSD_JOURNAL}" ]; then
+      ceph-disk -v prepare ${OSD_JOURNAL}
```
leseb: are you trying to prepare the journal device? or to get a new osd from that journal, it seems like you're doing the latter but then i'm missing the point.

daemon/README.md
potto007: I have this under the Ceph OSD directory section - refer to line 226. I put an example on line 260 under the "Multiple OSDs" subsection -- maybe it would help if I anchor-linked to it?

leseb: Oh right, didn't notice that with the github diff :).
Let's keep it that way.

daemon/entrypoint.sh
potto007: My intention is to prepare a journal for the corresponding OSD-id. So for example, when I have the dir tree as: /var/lib/ceph/osd/ceph-13, /var/lib/ceph/osd/ceph-25; and assuming I have created two partitions to use as journals: /dev/sdb1, /dev/sdb2; then I could pass `-e OSD_JOURNAL13=/dev/sdb1 -e OSD_JOURNAL25=/dev/sdb2` in the `docker run` command and expect that sdb1 will be prepared and made as journal for OSD.13 and sdb2 for OSD.25.

potto007: Incidentally, I do have another branch I am using on my systems that simply adds the ability to pass a flag to "preserve journals", which really just skips the `ceph-disk prepare` step. I haven't submitted it because I've only added it for osd_directory right now, as that is the only osd mode I use in ceph-docker.

diff --git c/daemon/README.md w/daemon/README.md
index f31e94c..73d6542 100644
--- c/daemon/README.md
+++ w/daemon/README.md
@@ -237,6 +237,7 @@ following environment variables will control its creation:
 * `JOURNAL` is the location of the journal (default is the `journal` file inside the OSD data directory)
 * `HOSTNAME` is the name of the host; it is used as a flag when adding the OSD to the CRUSH map
 * `OSD_JOURNAL<osd-id>` optional - providing for each OSD directory to be used will prepare and mount the supplied block device
+* `JOURNAL_PRESERVE` optional - if set, skips the prepare step for OSD_JOURNAL

 The old option `OSD_ID` is now unused.  Instead, the script will scan for each directory in `/var/lib/ceph/osd` of the form `<cluster>-<osd-id>`.

diff --git c/daemon/entrypoint.sh w/daemon/entrypoint.sh
index 8a0b5eb..5a4bacd 100755
--- c/daemon/entrypoint.sh
+++ w/daemon/entrypoint.sh
@@ -279,7 +279,9 @@ function osd_directory {
     OSD_JOURNAL_TMP="OSD_JOURNAL$OSD_ID"
     OSD_JOURNAL=${!OSD_JOURNAL_TMP}
     if [ -n "${OSD_JOURNAL}" ]; then
-      ceph-disk -v prepare ${OSD_JOURNAL}
+      if [ -z "${JOURNAL_PRESERVE}" ]; then
+        ceph-disk -v prepare ${OSD_JOURNAL}
+      fi
       OSD_J=${OSD_JOURNAL}
     elif [ -n "${JOURNAL_DIR}" ]; then
        OSD_J="${JOURNAL_DIR}/journal.${OSD_ID}"

daemon/README.md
potto007: That's what I suspected... github diffs obsfucate context quite nicely. ;)

daemon/entrypoint.sh
Ulexus: I would greatly prefer to not corrupt `osd_directory` mode with that `ceph-disk` script.  You can pass a device to the `ceph-osd ... --mkjournal --osd-journal ...` command, so there should be no need for `ceph-disk`.

daemon/README.md
```diff
@@ -259,6 +260,9 @@ There are two workarounds, at present:
 To run multiple OSDs within the same container, simply bind-mount each OSD datastore directory:
 * `docker run -v /osds/1:/var/lib/ceph/osd/ceph-1 -v /osds/2:/var/lib/ceph/osd/ceph-2`
 
+To run multiple OSDs (for example: osd.5 & osd.11) in the same container, with multiple journals on block devices:
+* `docker run -v /var/lib/ceph:/var/lib/ceph -v /dev:/dev -e OSD_TYPE=directory -e OSD_JOURNAL5=/dev/sdb1 -e OSD_JOURNAL11=/dev/sdb2 ...`
```
Ulexus: For clarity, consider underscore-separating the OSD ID from the `OSD_JOURNAL` prefix (e.g. `OSD_JOURNAL_5`).

daemon/entrypoint.sh
leseb: Agree with @Ulexus but to go back for a second to your `ceph-disk` usage. Perhaps it's me but if you run `ceph-disk prepare /dev/sdb1` you will just get a new OSD on this partition. In the end, this won't prepare a journal but create a new OSD. Am I missing something here?

leseb:
@potto007 Not sure about the status of this one, are you still looking for this patch to be merged?

PR #235, opened 2016-03-23
Correct the document wrong about default rgw port number

minshenglin:
The default port number of rgw has been changed after PR #201.


leseb:
Thanks!

PR #237, opened 2016-04-06
don't activate osd in osd_disk_prepare

rootfs:
osd is activated in `osd_activate`


leseb:
I was looking for this one!
Thanks @rootfs!

PR #238, opened 2016-04-07
daemon: multiple osd in one container

leseb:
Feature proposed by @oll-78.
It brings the ability to run multiple OSDs within a single container.
I thought we were already able to do so but didn't find any mention in
the code. The only thing I recall is a patch from @dmitryint in #24 but
since I never used it I'll let either @dmitryint or @Ulexus comment on
this. It seems that this patch uses ceph-disk instead.

Signed-off-by: Oleg Lazo <oll@niif.spb.su>

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -33,6 +33,7 @@ set -e
 : ${KV_PORT:=4001} # PORT 8500 for Consul
 
 CEPH_OPTS="--cluster ${CLUSTER}"
+MOUNT_OPTS="-t xfs -o noatime,inode64"
 
 # ceph config file exists or die
 function check_config {
@@ -226,6 +227,9 @@ function start_osd {
       activate)
          osd_activate
          ;;
+      devices)
+         osd_disks
+         ;;
       *)
          if [[ ! -d /var/lib/ceph/osd || -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then
             echo "No bootstrapped OSDs found; trying ceph-disk"
@@ -400,7 +404,7 @@ function osd_activate {
   chown ceph. /var/lib/ceph/osd
   if [[ ! -z "${OSD_JOURNAL}" ]]; then
     chown ceph. ${OSD_JOURNAL}
-    ceph-disk -v --setuser ceph --setgroup disk activate ${OSD_DEVICE}
+    ceph-disk -v --setuser ceph --setgroup disk activate $(dev_part ${OSD_DEVICE} 1)
   else
     chown ceph. $(dev_part ${OSD_DEVICE} 2)
     ceph-disk -v --setuser ceph --setgroup disk activate $(dev_part ${OSD_DEVICE} 1)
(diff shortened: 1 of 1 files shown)


Ulexus:
@leseb Yes, multiple OSDs are currently supported (and I use it), but only for osd_directory mode.  The operation is similar to your implementation.  I think the two implementations should be mergeable.  You have a lot of `ceph-disk` statements in yours, which wouldn't translate... but perhaps those are able to be abstracted?

daemon/entrypoint.sh
```diff
@@ -420,6 +424,102 @@ function osd_activate {
...
+      OSD_DEV="/dev/$(echo ${i}|sed 's/\(.*\):\(.*\)/\2/')"
+      if [[ "$(parted --script ${OSD_DEV} print | egrep '^ 1.*ceph data')" && ${OSD_FORCE_ZAP} -ne "1" ]]; then
+        echo "ERROR- It looks like this device is an OSD, set OSD_FORCE_ZAP=1 to use this device anyway and zap its content"
+        exit 1
+      elif [[ "$(parted --script ${OSD_DEV} print | egrep '^ 1.*ceph data')" && ${OSD_FORCE_ZAP} -eq "1" ]]; then
+        ceph-disk -v zap ${OSD_DEV}
+      fi
+
+      if [[ ! -z "${OSD_JOURNAL}" ]]; then
+        ceph-disk -v prepare ${OSD_DEV} ${OSD_JOURNAL}
```
Ulexus: Are the `prepare` and `activate` operations non-destructive?

daemon/entrypoint.sh
```diff
@@ -420,6 +424,102 @@ function osd_activate {
...
+#        chown ceph. ${OSD_JOURNAL}
+        ceph-disk -v --setuser ceph --setgroup disk activate $(dev_part ${OSD_DEV} 1)
+      else
+        ceph-disk -v prepare ${OSD_DEV}
+#        chown ceph. $(dev_part ${OSD_DEV} 2)
+        ceph-disk -v --setuser ceph --setgroup disk activate $(dev_part ${OSD_DEV} 1)
+      fi
+
+      OSD_ID=$(cat /var/lib/ceph/osd/$(ls -ltr /var/lib/ceph/osd/ | tail -n1 | awk -v pattern="$CLUSTER" '$0 ~ pattern {print $9}')/whoami)
```
Ulexus: This is definitely a better way of obtaining the `OSD_ID` than the `ceph_directory` implementation uses (directory name).

daemon/entrypoint.sh
Ulexus: Never mind... this is in the creation section...

daemon/entrypoint.sh
```diff
@@ -420,6 +424,102 @@ function osd_activate {
...
+   do
+    osd_id=$(echo ${i}|sed 's/\(.*\):\(.*\)/\1/')
+    osd_dev="/dev/$(echo ${i}|sed 's/\(.*\):\(.*\)/\2/')"
+    if [ ${osd_id} = ${1} ]; then
+      echo -n "${osd_dev}"
+    fi
+  done
+}
+
+function set_osd_run {
```
Ulexus: Good functional abstraction.  We should use this in `osd_directory`, too

daemon/entrypoint.sh
```diff
@@ -226,6 +227,9 @@ function start_osd {
       activate)
          osd_activate
          ;;
+      devices)
```
Ulexus: Does this really need another top-level OSD type?  Can this not be merged in with the existing `ceph-disk` stype OSD?  If it mandates compatibility breakage, then that is sufficient reason, of course.

dmitryint:
@leseb I use multiple osd in one container feature in production more than a year. Everything works fine.
As mentioned by @Ulexus it now for only osd_directory mode.
It would be great if this functionality be extended to other modes.

Lupus:
@dmitryint a bit offtopic, what do you do when individual drives (and thus mounted directories) fail and you replace a drive? Do you restart the whole container with all the OSD's?

dmitryint:
@Lupus Yes, that's right. After `ceph osd set noout` command.

daemon/entrypoint.sh
oll-78: This is for new feature (I've called it DISK_TYPE=devices), when needed zap&prepared&activate). In environment need add variable like OSD_DISKS="0:sda 1:sdb 2:sdc"

daemon/entrypoint.sh
oll-78: Why not? But the code will be more complicated, imho :)

Ulexus: No real problem with it.  We just already have several, and their names are already somewhat nebulous and difficult for newcomers to fathom.

Ulexus:
lgtm

leseb:
Thanks to everyone for the review :)

PR #239, opened 2016-04-07
WIP: add osd_activate_journal to match ceph-disk activate-journal command

rootfs:
want to float the idea first. The new command will be used when udev triggers `ceph-disk activate-journal`
@leseb

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -230,6 +230,9 @@ function start_osd {
       devices)
          osd_disks
          ;;
+      activate_journal)
+         osd_activate_journal
+         ;;
       *)
          if [[ ! -d /var/lib/ceph/osd || -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then
             echo "No bootstrapped OSDs found; trying ceph-disk"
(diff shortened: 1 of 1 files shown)


leseb:
@rootfs I've never really used `activate-journal` before but this seems like a valid patch to me :).
@Ulexus anything to add?

daemon/entrypoint.sh
```diff
@@ -592,6 +629,10 @@ case "$CEPH_DAEMON" in
       OSD_TYPE="activate"
       start_osd
       ;;
+    osd_ceph_activate_journal)
```
leseb: s/osd_ceph_activate_journal/osd_activate_journal/?

daemon/entrypoint.sh
```diff
@@ -602,8 +643,8 @@ case "$CEPH_DAEMON" in
       if [ ! -n "$CEPH_DAEMON" ]; then
           echo "ERROR- One of CEPH_DAEMON or a daemon parameter must be defined as the name "
           echo "of the daemon you want to deploy."
-          echo "Valid values for CEPH_DAEMON are MON, OSD, OSD_DIRECTORY, OSD_CEPH_DISK, OSD_CEPH_DISK_PREPARE, OSD_CEPH_DISK_ACTIVATE, MDS, RGW, RESTAPI"
-          echo "Valid values for the daemon parameter are mon, osd, osd_directory, osd_ceph_disk, osd_ceph_disk_prepare, osd_ceph_disk_activate, mds, rgw, restapi"
+          echo "Valid values for CEPH_DAEMON are MON, OSD, OSD_DIRECTORY, OSD_CEPH_DISK, OSD_CEPH_DISK_PREPARE, OSD_CEPH_DISK_ACTIVATE, OSD_CEPH_JOURNAL_ACTIVATE, MDS, RGW, RESTAPI"
```
leseb: s/OSD_CEPH_JOURNAL_ACTIVATE/OSD_CEPH_ACTIVATE_JOURNAL/?

leseb:
@rootfs do you mind rebasing you patch? Thanks!

daemon/entrypoint.sh
rootfs: that is to be consistent with the `ceph-disk` command is activate-journal :)

leseb: yes but if you use `OSD_CEPH_JOURNAL_ACTIVATE` this won't work because then the code will look for `osd_ceph_journal_activate` in the code (see the `tr` function) which does not exist. Am I missing something? See line 632, the name is different.

rootfs:
rebased, thanks!

daemon/entrypoint.sh
rootfs: right, fixed. thanks!

leseb:
@rootfs thanks!

PR #240, opened 2016-04-09
Update KV store when monmap is successfully pulled

fmeppo:
The monmap is stored in the KV store, but there's no obvious means to update it when it changes.  This patch updates the monmap whenever it's successfully pulled from peers, meaning the KV store's version will update when a mon restarts.  A bit awkward, but it ensures that if several new mons are added the KV store's monmap isn't too old.


Ulexus:
lgtm

PR #241, opened 2016-04-10
Update ceph-config image

xcompass:
It seems the image name changed from ceph/ceph-config to ceph/config.


xcompass:
I'll use a different branch to send PR. Close this for now and will open a new one

PR #242, opened 2016-04-10
Update ceph-config image

xcompass:
It seems the image name changed from ceph/ceph-config to ceph/config.


leseb:
This image is a bit deprecated as the `ceph/daemon` image has at builtin capability for that.

xcompass:
I figured. I'm using etcd as backend with daemon now. Maybe there should be a note in the readme.

leseb:
@xcompass will do!

PR #243, opened 2016-04-10
Fix failed ceph disk prepare step

xcompass:
`ceph osd crush create-or-move` is called in activate step. In
prepare stage, there is no whoami generated. So it will failed when
getting OSD_ID casuing the container failed to start.


leseb:
Good catch @xcompass, moreover this step is a bit redundant since it also happens during the activation phase.
So we should remove it.
Thanks!

PR #245, opened 2016-04-13
Add example unit files for deploying with Etcd and instructions

xcompass:


examples/coreos/units-etcd/ceph-mon@.service
```diff
@@ -0,0 +1,36 @@
...
+ExecStopPost=-/usr/bin/docker rm %p
+Restart=always
+RestartSec=10s
+TimeoutStartSec=120
+TimeoutStopSec=15
+
+[X-Fleet]
+Conflicts=ceph-mon@*.service
+
+;; Start on all machines with the mon=true metadata
```
Ulexus: I'm pretty sure instance-oriented units can't be used as global units... or at least it will lead to unexpected results.  I think it would be sufficient to document this, rather than issuing separate, non-instance-oriented units (without the `@` suffix).

examples/coreos/units-etcd/ceph-rgw@.service
```diff
@@ -0,0 +1,24 @@
...
+Description=Rados Gateway Service
+After=docker.service
+
+[Service]
+EnvironmentFile=/etc/environment
+Environment=RGW_NAME=%H
+ExecStartPre=-/usr/bin/docker kill %p
+ExecStartPre=-/usr/bin/docker rm %p
+ExecStart=/usr/bin/docker run --rm --name %p -p 80:80 \
+   --net=host \
```
Ulexus: I'd prefer to not have `--net=host` as a default in an example, where it isn't needed.  rgw works just fine without it.

examples/coreos/units-etcd/ceph-mds@.service
```diff
@@ -0,0 +1,27 @@
...
+Description=Ceph Metadata Service
+After=docker.service
+
+[Service]
+EnvironmentFile=/etc/environment
+; Note: the MDS_NAME must not be (solely?) numeric.  I use "a", "b", etc (ceph-mds@a, ceph-mds@b, etc)
+Environment=MDS_NAME=%i
+ExecStartPre=-/usr/bin/docker kill %p
+ExecStartPre=-/usr/bin/docker rm %p
+ExecStart=/usr/bin/docker run --rm --name %p --net=host \
```
Ulexus: I'm pretty sure mds doesn't require `--net=host`

Ulexus:
A few nit picks, but thanks for this!

Ulexus:
Also of note:  this depends on fleet, too, not just etcd.

examples/coreos/units-etcd/ceph-mon@.service
xcompass: Not sure if I understand your point. The global flag is commented out. Yes, conflicts and global can't be used together. Fleet will complain.

examples/coreos/units-etcd/ceph-rgw@.service
xcompass: rgw works without it but etcd needs for now, because of this: https://github.com/ceph/ceph-docker/blob/master/daemon/config.kv.sh#L121. If localhost is used, kviator won't be able to connect as there is no etcd inside container. And it will go to dead loop on line 121. 

Without --net=host, people need to specify the real IP of the host, which is not feasible, or etcd server IP in KV_IP, instead of 127.0.0.1. We probably want to avoid hard code etcd server IP in the unit file. 

Do you think it will be appropriate to update get_config function to accommodate this use case. If KV_TYPE is enabled and KV_IP is 127.0.0.1, then get docker host IP instead of using 127.0.0.1. The docker host IP can be retrieved by `/sbin/ip route|awk '/default/ { print $3 }'`.

examples/coreos/units-etcd/ceph-mds@.service
xcompass: Same above. But I copied this from yours: https://github.com/ceph/ceph-docker/blob/master/examples/coreos/units/ceph-mds@.service#L11 :)

xcompass:
Fleet is mentioned in README.md

examples/coreos/units-etcd/ceph-mds@.service
Ulexus: :)  No doubt I did, alas.

Ulexus:
Thanks!  Best practices are hard to come by, right now.  A lot depends on the networking setup used, so I don't see any need to hold this up on that long-term esoterica.

Issue #249, opened 2016-04-14
RFC: new bootstrap strategy for config 


rootfs:
Current config container uses the [following strategy](https://hub.docker.com/r/ceph/config/):
- An /etc/ceph/ceph.conf is found, do nothing.
- If a cluster configuration is available, it will be written to /etc/ceph
- If no cluster configuration is available, it will be bootstrapped. A lock mechanism is used to allow concurrent deployment of multiple hosts.

I am proposing:
- if /etc/ceph/ceph.conf is found, skip creating /etc/ceph/ceph.conf, but continue the following steps
- If a cluster configuration is available, it will be written to /etc/ceph
- If no cluster configuration is available, it will be bootstrapped. A lock mechanism is used to allow concurrent deployment of multiple hosts. In particular, for each keyring (mon, osd, admin, etc), if it exists, skip; otherwise create this keyring.

The idea is that when we deploy containers using ansible and we create ceph.conf from a template, we can still be able to allow config to bootstrap.

@leseb

Ulexus:
I can wholeheartedly agree, that is a much better process.  I think the differences are less intentional design and more just not having gotten around to it.

PR #250, opened 2016-04-14
implement #249

rootfs:
this implements #249 
@Ulexus @leseb

### DIFF
--- a/daemon/config.static.sh
+++ b/daemon/config.static.sh
@@ -27,33 +27,49 @@ ENDHERE
       sed -i '/mon host/d' /etc/ceph/${CLUSTER}.conf
       echo "mon host = ${MON_IP}" >> /etc/ceph/${CLUSTER}.conf
     fi
+    else
+      # extract fsid from ceph.conf
+      fsid=`grep "fsid" /etc/ceph/${CLUSTER}.conf |awk '{print $NF}'`
+  fi
 
+  if [ ! -e /etc/ceph/${CLUSTER}.client.admin.keyring ]; then
     # Generate administrator key
     ceph-authtool /etc/ceph/${CLUSTER}.client.admin.keyring --create-keyring --gen-key -n client.admin --set-uid=0 --cap mon 'allow *' --cap osd 'allow *' --cap mds 'allow'
+  fi
 
+  if [ ! -e /etc/ceph/${CLUSTER}.mon.keyring ]; then
     # Generate the mon. key
     ceph-authtool /etc/ceph/${CLUSTER}.mon.keyring --create-keyring --gen-key -n mon. --cap mon 'allow *'
+  fi
 
-    # Create bootstrap key directories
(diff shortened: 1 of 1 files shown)


Ulexus:
This appears to be half of #249, yes?  This does not incorporate the KV part.

rootfs:
ah, yes, working ...

rootfs:
@Ulexus @leseb what do you think about the following scenario?

In the beginning, admin generates keyrings and conf, put them into KV store, and ceph-docker gets from KV store and write to the host.

Later admin updates the config and keyrings in the KV store, should ceph-docker overwrite (current behavior) or keep the existing config?

daemon/config.static.sh
```diff
@@ -27,33 +27,49 @@ ENDHERE
...
     # Generate the RGW bootstrap key
     ceph-authtool /var/lib/ceph/bootstrap-rgw/${CLUSTER}.keyring --create-keyring --gen-key -n client.bootstrap-rgw --cap mon 'allow profile bootstrap-rgw'
+  fi
 
     # Apply proper permissions to the keys
     chown ceph. /etc/ceph/${CLUSTER}.mon.keyring /var/lib/ceph/bootstrap-osd/${CLUSTER}.keyring /var/lib/ceph/bootstrap-mds/${CLUSTER}.keyring /var/lib/ceph/bootstrap-rgw/${CLUSTER}.keyring
 
+  if [ ! -e /etc/ceph/monmap ]; then
     # Generate initial monitor map
     monmaptool --create --add ${MON_NAME} "${MON_IP}:6789" --fsid ${fsid} /etc/ceph/monmap
```
dang: This won't work, because fsid is only defined if ceph.conf does not exist.  Something like this is necessary on the end of the if !ceph.conf block:
  else
    fsid=${CEPH_FSID}
  fi

rootfs: @dang yes, that's the trick [here](https://github.com/ceph/ceph-docker/pull/250/files/22b38cdc69bf82df8e543ac4e5b9025a0efefa38#diff-0ebd42e0be1e909c06c7e6161d0d3b42R32)

daemon/config.static.sh
```diff
@@ -27,33 +27,49 @@ ENDHERE
       sed -i '/mon host/d' /etc/ceph/${CLUSTER}.conf
       echo "mon host = ${MON_IP}" >> /etc/ceph/${CLUSTER}.conf
     fi
+    else
+      # extract fsid from ceph.conf
+      fsid=`grep "fsid" /etc/ceph/${CLUSTER}.conf |awk '{print $NF}'`
```
rootfs: @dang ^^^

dang: Ah, missed it.  Good.

rootfs:
@Ulexus @leseb this fix works for static config. can we merge this part first and discuss KV at a separate issue? Thanks.

Ulexus:
I agree.  Thanks @rootfs

PR #257, opened 2016-05-05
Adds variable to pass flags to ceph-disk's prepare

### LINKED ISSUE #256: Add support for passing extra options to ceph-disk prepare
Add an `OSD_PREPARE_FLAGS` variable to pass extra options like `--dmcrypt`, `--dmcrypt-key-dir` or `--bluestore` when preparing a disk

ivotron:
fixes #256 . I've only tested this with `--bluestore`

### DIFF
--- a/daemon/entrypoint.sh
+++ b/daemon/entrypoint.sh
@@ -11,6 +11,7 @@ set -e
 : ${MDS_NAME:=mds-${HOSTNAME}}
 : ${OSD_FORCE_ZAP:=0}
 : ${OSD_JOURNAL_SIZE:=100}
+: ${CEPH_DISK_PREPARE_FLAGS:=''}
 : ${CRUSH_LOCATION:=root=default host=${HOSTNAME}}
 : ${CEPHFS_CREATE:=0}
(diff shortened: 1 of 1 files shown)


daemon/entrypoint.sh
```diff
@@ -10,6 +10,7 @@ set -e
 : ${NETWORK_AUTO_DETECT:=0}
 : ${MDS_NAME:=mds-${HOSTNAME}}
 : ${OSD_FORCE_ZAP:=0}
+: ${OSD_PREPARE_FLAGS:=''}
```
Ulexus: I would have a mild preference to name this something more related to what it actually does:  `CEPH_DISK_PREPARE_FLAGS`, perhaps.  It is verbose, but since it isn't really related to OSDs themselves (rather, the ceph-disk _way_ of manipulating OSDs), I'm uncomfortable calling it `OSD_PREPARE_FLAGS`, succinct though that is.

ivotron:
Updated with better var name. Thanks!

Ulexus:
Thanks!

PR #259, opened 2016-05-12
Branch management future

leseb:
Introducing new branch management

We now have a hierarchy tree that contains all the ceph versions
including OS versions. Each OS version has its own `base` and `daemon`
directories. When it comes to Ubuntu, 14.04 remains the default where
16.04 has symlinks to 14.04. Only the base image version changes.

We have a script that can be executed from anywhere, even if you are
outside the repository it will work.
It will look for the last commit that applied a change on a **specific**
version. We do not support multiple changes on different ceph versions
or OS. So if someone wants to make a change on a PR, the PR must only
change a single ceph version AND a single OS.

The script will do several things:

0. determine which ceph version and OS was modified
1. create a new branch name
2. delete the old branch
3. create a new branch
4. copy daemon and base files to the root
5. commit the new changes
6. tag the commit
7. push the new branch

The Docker Hub has been configured to build a new image for every tag we
create.
Since the approprieate commit (in the proper branch) is tagged
a build will automatically be triggered.

Co-Authored-By: Erwan Velu <erwan@redhat.com>


leseb:
I think I need to revert https://github.com/ceph/ceph-docker/pull/259/commits/5ce681811029951fbb0140dcf6982014ffd1d114
So we don't disrupt existing deployments.

leseb:
I also need to create all the images.

dang:
I have a Fedora jewel to add to this, but it brings up a question: how are we tagging these on the registry?  They can't all be tagged as ceph/base:jewel.  We could tag them jewel-fedora, but that fails for the multiple Ubuntu versions.  What's the thoughts here?

Note, for committing, it only effects the README.md file.

leseb:
@dang which fedora version?
I still need to work on the tagging schema.

dang:
@leseb Fedora 23

leseb:
@dang do you have the base directory for this one?

dang:
https://github.com/dang/ceph-docker/tree/branch-management-future

leseb:
@dang added.

leseb:
I need to rebase with https://github.com/ceph/ceph-docker/pull/266

PR #271, opened 2016-05-25
Use KV store with any cluster name

JrCs:
* confd create a /etc/ceph/ceph.conf file but for cluser name different
  of 'ceph' the configuration file must be /etc/ceph/$CLUSTER.conf

### DIFF
--- a/daemon/confd/src/ceph.conf.toml.in
+++ b/daemon/confd/src/ceph.conf.toml.in
@@ -1,6 +1,6 @@
 [template]
 src = "ceph.conf.tmpl"
-dest = "/etc/ceph/ceph.conf"
+dest = "/etc/ceph/@CLUSTER@.conf"
 keys = [
  "/auth/",
  "/global/",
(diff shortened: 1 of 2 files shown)


leseb:
why do you have to change the filename?

JrCs:
By default ceph use the configuration with the cluster name in it.
Also the entrypoint.sh failed if it doesn't find the right file: [entrypoint.sh#L41](https://github.com/ceph/ceph-docker/blob/master/daemon/entrypoint.sh#L41)

leseb:
Sorry if I'm missing something but why do you go from `/etc/confd/src/ceph.conf.toml.in` to `/etc/confd/conf.d/ceph.conf.toml`

JrCs:
The `/etc/confd/src/ceph.conf.toml.in`is a template that have @CLUSTER@ token.
I use this template as a source to create the `/etc/confd/conf.d/ceph.conf.toml`replacing @CLUSTER@ token with the value of $CLUSTER.

leseb:
Oh I see, makes sense, seems to be the only way to interpret this variable.

PR #274, opened 2016-05-26
Add RGW_REGION and RGW_ZONE variables

JrCs:
Add new variables to help creating multi-sites rados gateways.


leseb:
New PR should go in the ceph-releases directory

yehudasa:
@JrCs 'region' is an obsolete term, should use 'zonegroup' instead

JrCs:
@yehudasa i know that `region`is an obsolete term, but for the radosgw-admin command the option is `--rgw-region`

yehudasa:
@JrCs can now use --rgw-zonegroup instead

JrCs:
@yehudasa yes i will remake my PR

PR #277, opened 2016-05-27
Workflow readme and dev contributions

leseb:


leseb:
@dang have a look at this.

leseb:
@ErwanAliasr1 if you want to have a look as well :)

generate-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+
+test_args $@
+
+case "$1" in
+  hammer|infernalis|jewel)
+    case "$2" in
+      centos)
+        case "$3" in
+          7)
+            generate_dev_dir $@
```
ErwanAliasr1: I'd say put a break here and move that 2 calls after the global case statement.

ErwanAliasr1: And so for the others case $3 statements

generate-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+function generate_dev_dir {
+  echo_info "Generating $BASEDIR/dev/$1/$2/$3"
+  mkdir -p $BASEDIR/dev/$1/$2/$3
+}
+
+function copy_files {
+  cp -Lrv ceph-releases/$1/$2/$3/* $BASEDIR/dev/$1/$2/$3
+}
+
+function wrong_combinaison {
```
dang: English is combination...  Probably best to use that?

generate-dev-env.sh
```diff
@@ -0,0 +1,78 @@
...
+        esac
+        ;;
+      fedora)
+        case "$3" in
+          23)
+            generate_dev_dir $@
+            copy_files $@
+            ;;
+          *)
+            wrong_combinaison
```
dang: Is there a reason to hard-enforce the existing distro's / versions, rather than just doing the copy?  This is for developers, after all.

generate-dev-env.sh
leseb: Sorry I'm french, yes I'll update that.

dang:
Looks good to me now.

leseb:
You're very welcome @dang.
Now let's wait for #279, so we can use the script properly.

PR #282, opened 2016-05-29
CI: build daemon image on travis and test it

leseb:
We rely on Travis's infrastructure to build the 'daemon' image and run
all the ceph daemons. So we run a monitor, an osd (with directory
scenario, a mds and one rgw).
We have a validation script to make they all started properly. During
the validation sequence, we wait for 20 seconds to let the cluster
stabilize and populate the pgs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+language: python
+python:
+  - 2.7
+
+branches:
+  only:
+    - master
+    - /^build-master-*/
+
+services:
+  - docker
+
+before_install:
+  - sudo ./travis-builds/build_imgs.sh
+
+# NOTE(leseb): somehow putting everything in a 'script' task does not work
+# so we have to split it up that way.
+# It seems that we have an issue when not running 'docker run' from an install step
+install:
+  - docker run -d --name ceph-mon --net=host -v /etc/ceph:/etc/ceph -v /var/lib/ceph:/var/lib/ceph -e MON_IP=127.0.0.1 -e CEPH_PUBLIC_NETWORK=127.0.0.0/8 daemon mon
+  - sudo ./travis-builds/bootstrap_osd.sh
+  - docker run -d --name ceph-osd -v /etc/ceph:/etc/ceph -v /var/lib/ceph:/var/lib/ceph daemon osd_directory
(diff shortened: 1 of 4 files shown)


leseb:
@Ulexus , @dmitryint I have the CI working.
However I'm having an issue with the directory scenario, the OSD seems to start and then dies...
`docker logs ceph-osd` doesn't return anything.

Would you mind having a look and see if I'm missing anything in the populate sequence?
Thanks!

Ulexus:
Yes, something is strange about the OSD logging.  It works fine on my systems, but I know we've had problems with it in the past.

That is very strange, though, that it starts up enough to register as UP and then dies.  Can you log to systemd or something so that Docker doesn't throw away the logs?

Ulexus:
So... you're manually bootstrapping the OSD directory, rather than running through the directory's bootstrap process.  Is that intentional?

Ulexus:
I've also never tried running an OSD on a subdirectory of a mount.  I _presume_ that should be fine, but I don't have any experience doing so.  While it is mostly just files, I do know that, at least in the case of BTRFS, the OSD does some things which are more than merely file manipulation.

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - docker
+
+before_install:
+  - cd daemon
+
+install:
+  - docker build -t daemon .
+  - docker run -d --name ceph-mon --net=host -v /etc/ceph:/etc/ceph -v /var/lib/ceph:/var/lib/ceph -e MON_IP=127.0.0.1 -e CEPH_PUBLIC_NETWORK=127.0.0.0/8 daemon mon
+  - sudo ../travis-builds/build-osd.sh
+  - docker run -d --name ceph-osd -v /etc/ceph:/etc/ceph -v /var/lib/ceph:/var/lib/ceph daemon directory
```
dmitryint: Try to use a OSD_DIRECTORY instead  directory.
`docker run -d --name ceph-osd -v /etc/ceph:/etc/ceph -v /var/lib/ceph:/var/lib/ceph daemon OSD_DIRECTORY `

leseb:
Thanks @Ulexus for your comments and review. Yes it is intentional to bootstrap the OSD earlier, this also serves another purpose, I'm trying to have this part being done by ceph-ansible and then have ceph-docker only start the process. This is not used here but a good test to validate that.
Nice catch @dmitryint, I misread the code :/

Looks like it's starting properly now :D

PR #284, opened 2016-06-01
daemon: remove all the package installations

leseb:
Packages installation should only happen in the base image.


leseb:
Somehow Travis just started to complain about `ARG` in the `Dockerfile`. I don't think my change is rising the issue... Investigating.

PR #286, opened 2016-06-02
base: remove 'arg' so travis is happy

leseb:
Travis is using Docker 1.8.x and thus does not support 'ARG' in the
Dockerfile. So instead re-using the old fashion way.

[ci skip]


leseb:
@Ulexus I need your advise on this one.
I don't like the idea of changing the repo for Travis but I'm not a big fan of running a too bleeding edge version of Docker. Last time I updated I got 1.11, the package came from Ubuntu Trusty so I believe it's safe.

I can always upgrade the Docker package on Travis if we don't want to have this change.

Thoughts?

Issue #19, opened 2015-02-20
Ceph OSD docker bug


leseb:
Just put this here as a reminder and potential blocker for Docker OSD deployments.
http://tracker.ceph.com/issues/10763

Ulexus:
I can affirm that this occurs even on physical machines, as long as the OSDs are running in Docker.  Host networking does not resolve the issue.  Splitting the OSD data/sync network from the access network does not resolve the issue, either.

This likely shares the same root cause which causes me to be unable to run more than one RBD-backed KVM instance on a single host without all the ports on the machine becoming exhausted from Ceph traffic.

Interestingly, in that case, if I run the KVM instances from systemd-nspawn instead of Docker, I do _not_ have the problem.  I suspect the same would be true if I ran the OSD from systemd-nspawn, as well.

I'm also quite interested to see if the problem exists with [rocket](https://github.com/coreos/rocket).  When I get some time, I'll throw together an ACI for the OSDs and try that.

I should also note that I have tried a number of options to try and mitigate this.
* `tcp nodelay = false` : No difference
* various `rbd cache` settings: No difference
* splitting the network: No difference
* reducing the TTL for connections: No difference (not expected, since the rate at which the ports exhaust is incredible)

Ulexus:
More discussion on this, from earlier:
[Ulexus/docker-ceph #8](https://github.com/Ulexus/docker-ceph/issues/8)

Ulexus:
@bobrik Anything to add to this?

bobrik:
@Ulexus not really, you know even more than I do.

peterrosell:
I'm also struggling with this problem. My experience is that as soon as I have two osd on the same host and a third osd is started the problem occurs.  It doesn't matter if the third osd is running on the same host or on another host. I can see in the monitor's log that 

I have been think that the sharing of host between the osds (-net=host) makes this problem occur. I haven't been able to test it yet. I have an idea of running docker 1.5 and set up an ipv6 network that the ceph containers can communicate via. By doing each container gets its own ip address and I don't need --net=host.
I read somewhere that you didn't get ceph to work without --net=host and --privileged, but I don't understand the reason.
Do you think it's any idea for me to try running with ipv6?

Ulexus:
I think it's a reasonable hypothesis that the `--net=host` is related.  The OSD - Mon communication really doesn't like NAT, which is the purpose for the `--net=host`.  

[...]

PR #306, opened 2016-06-16
daemon: new directory mode

leseb:
the current directory mode aims to run multiple osd processes in a
single container. It uses runit, which is hard to find and not packaged
for some distros. This commit introduces a simple directory mode that
will only run one osd process per container. Prior to do that it will
check if a lock is already held by another osd process. If so it won't
start the osd and will just try to find another one to run. If nothing
is found or all the osds available are already running with exit the
container explaining that all the osds are already running.

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -225,6 +225,9 @@ function start_osd {
       directory)
          osd_directory
          ;;
+      directory_single)
+         osd_directory_single
+         ;;
       disk)
          osd_disk
          ;;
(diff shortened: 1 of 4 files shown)


ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -258,6 +261,51 @@ function start_osd {
 }
 
 
+########################
+# OSD_DIRECTORY_SIMPLE #
+########################
+
+function osd_directory_simple {
+  if [[ ! -d /var/lib/ceph/osd ]]; then
```
Ulexus: Perhaps:
`$(find /var/lib/ceph/osd -type d |wc -l) -gt 1`

Basically, check that the directory `/var/lib/ceph/osd` contains at least one subdirectory.

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -720,6 +768,10 @@ case "$CEPH_DAEMON" in
       OSD_TYPE="directory"
       start_osd
       ;;
+   osd_directory_simple)
```
Ulexus: Perhaps name this `osd_directory_single` instead of simple?  It is simpler, yes, but from the user's perspective, it is "single."

leseb: correct thanks will do!

Ulexus:
I definitely prefer decoupling, but we do need to keep in mind the PID namespace problem, #19 .   This was the whole reason we introduced the multiple-OSD-per-container fix.

leseb:
@Ulexus right thanks for the reminder, I completely forgot about that.

Ulexus:
Nice feature, that.  It's not terribly useful when the reference container and order of invocation is indeterminate, but at least it _can_ be done now without resorting to `--pid=host`.  Yay

Ulexus:
I would only add somewhere in the README or docs a notice that if multiple `osd_directory_single` containers are to be used on a single host, those containers **must** share their PID namespaces, either with `--pid=host` or `--pid=container:<id>`.

leseb:
@Ulexus are we good to go? :)

Ulexus:
lgtm.  Thanks!

PR #312, opened 2016-06-21
Fix bug of NETWORK_AUTO_DETECT

mkkie:
Hi,

1.
If we got some IP address such as 10.1.86.10, end with zero.
It match the rule "grep -v "0$" " and MON_IP won't set anything.
Using grep -w can help us.

2.
CEPH_PUBLIC_NETWORK alwayls grep the network 0.0.0.0/1.
Maybe using "grep -vE "^127|^0" " is a better choice.


leseb:
Good catch, thanks!

Issue #313, opened 2016-06-22
trouble with OSD_JOURNAL


immesys:
I'm having some difficulty getting an OSD up. I'm using jewel-ubuntu-16.04. I'm pretty much following the documentation and would really appreciate some help.

I start by preparing the disk, I have the journal on a separate ssd partition. The mon is already up:

```
docker run -it --net=host --privileged=true --pid=host \
   -v /dev/:/dev/ -v /srv/ceph:/var/lib/ceph \
  -e OSD_DEVICE=/dev/sdg -e OSD_JOURNAL=/dev/sda1 \
  -e OSD_FORCE_ZAP=1 -e OSD_TYPE=prepare \
  -e KV_TYPE=etcd -e KV_IP=127.0.0.1 \
  ceph/daemon:tag-build-master-jewel-ubuntu-16.04 osd
```

Which appears to work (full output https://pastee.org/zxz7c)

Then I want to activate it, normally with ceph I'd use the partition (/dev/sdg1), but the documentation suggests the whole disk, so I try that:

```
docker run -it --net=host --privileged=true --pid=host \
  -v /dev/:/dev/  -v /srv/ceph:/var/lib/ceph \
  -e OSD_DEVICE=/dev/sdg -e OSD_JOURNAL=/dev/sda1 \
  -e OSD_TYPE=activate \
  -e KV_TYPE=etcd -e KV_IP=127.0.0.1 \
ceph/daemon:tag-build-master-jewel-ubuntu-16.04 osd
```

But it fails:

```
2016-06-22T02:15:07Z silo confd[4119]: INFO Backend set to etcd
2016-06-22T02:15:07Z silo confd[4119]: INFO Starting confd
2016-06-22T02:15:07Z silo confd[4119]: INFO Backend nodes set to 127.0.0.1:4001
2016-06-22T02:15:07Z silo confd[4119]: INFO Target config /etc/ceph/ceph.conf out of sync
2016-06-22T02:15:07Z silo confd[4119]: INFO Target config /etc/ceph/ceph.conf has been updated
Adding bootstrap keyrings
main_activate: path = /dev/sdg
get_dm_uuid: get_dm_uuid /dev/sdg uuid path is /sys/dev/block/8:96/dm/uuid
command: Running command: /sbin/blkid -p -s TYPE -o value -- /dev/sdg
Traceback (most recent call last):
  File "/usr/sbin/ceph-disk", line 9, in <module>
[...]

PR #314, opened 2016-06-22
entrypoint: fix activate with OSD_JOURNAL

immesys:
Previously, the partition would be added onto OSD_DEVICE which is already assumed to contain the partition, causing activate to fail if OSD_JOURNAL is specified (see #313)


leseb:
Thanks @immesys

PR #317, opened 2016-06-23
entrypoint: accept /dev/disk/by-x names for OSD_DEVICE

immesys:
All the ceph tools support symlinks for the devices (such as /dev/disk/by-path) but `ceph-disk` list will give the canonical names, so we need to resolve the symlink when we grep the output from ceph-disk

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -445,14 +445,16 @@ function osd_activate {
   fi
   mkdir -p /var/lib/ceph/osd
   chown ceph. /var/lib/ceph/osd
+  # resolve /dev/disk/by-* names
+  ACTUAL_OSD_DEVICE=$(readlink -f ${OSD_DEVICE})
   if [[ ! -z "${OSD_JOURNAL}" ]]; then
     chown ceph. ${OSD_JOURNAL}
     ceph-disk -v --setuser ceph --setgroup disk activate ${OSD_DEVICE}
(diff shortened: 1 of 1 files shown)


immesys:
Note that there is a small problem here. If you do not specify OSD_JOURNAL, then entrypoint will helpfully try add the partition number onto the end of the name, but that won't work for by-path or by-id names. This PR is still strictly more functional than what is there at present (I do not believe it introduces regressions) but it only works for symlinked drives if you also happen to specify OSD_JOURNAL.

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -445,14 +445,16 @@ function osd_activate {
   fi
   mkdir -p /var/lib/ceph/osd
   chown ceph. /var/lib/ceph/osd
+  # resolve /dev/disk/by-* names
+  ACTUAL_OSD_DEVICE=$(readlink -f ${OSD_DEVICE})
```
leseb: this assumes OSD_DEVICE is a link, can you instead do something like `if [[ "$(find <path to device> -type l) -gt 0 ]]; then`, this will determine if the file is a link or not.

immesys: Yeah I had something like that originally, but readlink does nothing if it's not a link and just prints /dev/sdf2 for example.

leseb: Oh I see, that's great then

PR #327, opened 2016-07-07
fixing KV using etcd /w ssl 

damoon:
i found two small issues that prevent the daemon image to startup


damoon:
maybe it as also a good idea to remove the hole "pipe output to dev/null". that made it really hard to debug.

leseb:
which one? I mean which pipe output to dev/null?

MichaelSp:
:+1:  thanks

Issue #324, opened 2016-07-06
ceph_disk.main.FilesystemTypeError: Cannot discover filesystem type


Raboo:
I'm running ceph/daemon:tag-build-master-jewel-ubuntu-14.04 osd with following environment variables:
```
-e OSD_DEVICE=/dev/sda \
-e OSD_JOURNAL=/dev/nvme0n1p1 \
-e KV_TYPE=consul \
-e KV_IP=192.168.0.1 \
-e KV_PORT=8500
```
and I get following error `ceph_disk.main.FilesystemTypeError: Cannot discover filesystem type: device /dev/sda: Line is truncated:`

#### docker logs -f ceph-osd-sda
```
2016-07-06T09:07:25Z vm04 confd[7674]: INFO Backend set to consul
2016-07-06T09:07:25Z vm04 confd[7674]: INFO Starting confd
2016-07-06T09:07:25Z vm04 confd[7674]: INFO Backend nodes set to vm01.se-ix.delta.prod:8500
2016-07-06T09:07:25Z vm04 confd[7674]: INFO Target config /etc/ceph/ceph.conf out of sync
2016-07-06T09:07:25Z vm04 confd[7674]: INFO Target config /etc/ceph/ceph.conf has been updated
Adding bootstrap keyrings
No bootstrapped OSDs found; trying ceph-disk
HEALTH_OK
command: Running command: /usr/bin/ceph-osd --cluster=ceph --show-config-value=fsid
command: Running command: /usr/bin/ceph-osd --check-allows-journal -i 0 --cluster ceph
command: Running command: /usr/bin/ceph-osd --check-wants-journal -i 0 --cluster ceph
command: Running command: /usr/bin/ceph-osd --check-needs-journal -i 0 --cluster ceph
get_dm_uuid: get_dm_uuid /dev/sda uuid path is /sys/dev/block/8:0/dm/uuid
command: Running command: /usr/bin/ceph-osd --cluster=ceph --show-config-value=osd_journal_size
get_dm_uuid: get_dm_uuid /dev/sda uuid path is /sys/dev/block/8:0/dm/uuid
get_dm_uuid: get_dm_uuid /dev/sda uuid path is /sys/dev/block/8:0/dm/uuid
get_dm_uuid: get_dm_uuid /dev/sda uuid path is /sys/dev/block/8:0/dm/uuid
command: Running command: /usr/bin/ceph-conf --cluster=ceph --name=osd. --lookup osd_mkfs_type
[...]

PR #328, opened 2016-07-07
daemon: better osd scenario prediction

leseb:
When 'osd' is only set in the docker run command as $1 we need to detect
which scenario is desired. The commit splits the logic into a new
function and makes the overall detection clearer in term of code
readability.

Half fix for #324


Ulexus:
Looks good to me

PR #334, opened 2016-07-11
daemon: fix osd_dir path per distro

leseb:
In a previous comment I was thinking that using non-full path would work
(to call the binaries). However it seems that we need fullpath.
Updating the path with the correct location for redhat and debian based
systems.

### DIFF
--- a/ceph-releases/jewel/centos/7/daemon/my_init
+++ b/ceph-releases/jewel/centos/7/daemon/my_init
@@ -1,1 +1,0 @@
-../../../ubuntu/14.04/daemon/my_init
--- a/ceph-releases/jewel/centos/7/daemon/my_init/LICENSE.txt
+++ b/ceph-releases/jewel/centos/7/daemon/my_init/LICENSE.txt
@@ -1,0 +1,19 @@
+Copyright (c) 2013-2014 Phusion
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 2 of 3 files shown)


damoon:
i just tried https://hub.docker.com/r/ceph/daemon/builds/bjbauvcs7jrhdyrgjukycjj/ but it still shows the error
i assume because it is based on https://hub.docker.com/r/ceph/base/builds/bppwmcjmdf4npbb4ykxryjr/ and that base image is not build yet

leseb:
@damoon yup tag-build-master-jewel-centos-7 is still building.

damoon:
i just downloaded and tried all current daemon images
latest tag-build-master-jewel-ubuntu-16.04 tag-build-master-jewel-ubuntu-14.04 tag-build-master-jewel-centos-7 show
```
Adding bootstrap keyrings
*** Running /etc/rc.local...
*** Booting runit daemon...
*** Runit started as PID 54
*** Runit exited with status 127
warning: /etc/service/ceph-2: unable to open supervise/ok: file does not exist
*** Killing all processes...
```
tag-build-master-jewel-fedora-24 shows
```
Adding bootstrap keyrings
*** Booting runit daemon...
*** Runit started as PID 55
*** Runit exited with status 127
sh: /usr/bin/sv: No such file or directory
sh: /usr/bin/sv: No such file or directory
*** Killing all processes...
```

russell-islam:
@leseb Can we just avoid smlink for each of the distribution. In that case we can update distribution specific bugs.

leseb:
@russell-islam we will once there is something that diverges :), just like I did here

leseb:
just pushed new branches with #336 so this should work now...

PR #335, opened 2016-07-12
osd_trying_to_determine_scenario better OSD detect

Raboo:
Sometimes the old code used to go into osd_disk function when already an OSD.
So it should go to osd_activate function.

My suggested change will most likely better detect already prepared OSD's.

Note that I haven't tested this code yet in a live node, but it's such a minor change, I don't see why it shouldn't work.
I can test it if you deem it necessary, otherwise it will get tested it as soon as it's merged pushed to docker hub :-)

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -79,25 +79,19 @@ function dev_part {
 }
 
 function osd_trying_to_determine_scenario {
-  if [[ ! -d /var/lib/ceph/osd || -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then
-    echo "No bootstrapped OSDs found; trying ceph-disk"
-    osd_disk
-    return
-  fi
-
   if [ -z "${OSD_DEVICE}" ]; then
     echo "Bootstrapped OSD(s) found; using OSD directory"
     osd_directory
     return
-  fi
-
-  if $(parted --script ${OSD_DEVICE} print | egrep -sq '^ 1.*ceph data'); then
+  elif $(parted --script ${OSD_DEVICE} print | egrep -sq '^ 1.*ceph data'); then
     echo "Bootstrapped OSD found; activating ${OSD_DEVICE}"
     osd_activate
+    return
   else
     echo "Device detected, assuming ceph-disk scenario is desired"
     echo "Preparing and activating ${OSD_DEVICE}"
     osd_disk
+    return
   fi
 }
 
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -82,16 +82,13 @@ function osd_trying_to_determine_scenario {
   if [ -z "${OSD_DEVICE}" ]; then
     echo "Bootstrapped OSD(s) found; using OSD directory"
     osd_directory
-    return
   elif $(parted --script ${OSD_DEVICE} print | egrep -sq '^ 1.*ceph data'); then
     echo "Bootstrapped OSD found; activating ${OSD_DEVICE}"
     osd_activate
-    return
   else
     echo "Device detected, assuming ceph-disk scenario is desired"
     echo "Preparing and activating ${OSD_DEVICE}"
     osd_disk
-    return
   fi
 }
 


ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -79,25 +79,19 @@ function dev_part {
 }
 
 function osd_trying_to_determine_scenario {
-  if [[ ! -d /var/lib/ceph/osd || -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then
```
leseb: If you remove this, the next if is wrong.
If we pass this `if` this means `/var/lib/ceph/osd` has some content, thus the next condition can safely assume if no  `OSD_DEVICE` is declared that we can try `osd_directory`.

leseb:
Can you describe the problem you had?
There is no need to do a long `if`, the purpose of independent `if` is for readability.

Raboo:
Well the problem occured when stepping a ceph/daemon:jewel to ceph/daemon:tag-build-master-jewel-ubuntu-14.04.

Starting the container without OSD_TYPE environment variable I would assume it would go into osd_activate/activate.
But it tried use osd_disk function instead.

```
2016-07-12T07:04:38Z vm03 confd[85864]: INFO Backend set to consul
2016-07-12T07:04:38Z vm03 confd[85864]: INFO Starting confd
2016-07-12T07:04:38Z vm03 confd[85864]: INFO Backend nodes set to vm01.se-ix.delta.prod:8500
2016-07-12T07:04:38Z vm03 confd[85864]: INFO Target config /etc/ceph/ceph.conf out of sync
2016-07-12T07:04:38Z vm03 confd[85864]: INFO Target config /etc/ceph/ceph.conf has been updated
Adding bootstrap keyrings
No bootstrapped OSDs found; trying ceph-disk
HEALTH_WARN 1 pgs degraded; 1 pgs peering; 1 pgs stale; 1 pgs undersized; recovery 1206/944742 objects degraded (0.128%); 1/35 in osds are down; noout,sortbitwise flag(s) set
ERROR- It looks like /dev/sda is an OSD, set OSD_FORCE_ZAP=1 to use this device anyway and zap its content
You can also use the zap_device scenario on the appropriate device to zap it
```

leseb:
It's weird that this condition failed:  `if [[ ! -d /var/lib/ceph/osd || -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then`. What's the state of `/var/lib/ceph/osd`?

Raboo:
Currently inside one of those OSD's that ended up in a reboot loop(currently started with "activate").
```
root@vm03:~# docker exec -ti ceph-osd-sdc /bin/bash
root@vm03:/# find /var/lib/ceph/osd -prune -empty
root@vm03:/# if [[ ! -d /var/lib/ceph/osd || -n "$(find /var/lib/ceph/osd -prune -empty)" ]]; then echo true; fi
```
So I don't know why it failed.

leseb:
@Raboo when you do bash, you're not entering a bash shell within the container.
Just keep on using exec for the find and if commands.

leseb:
@Raboo can you still reproduce your error?

Raboo:
Yes

Raboo:
Or I don't know, I just started it manually (not using chef) using a slightly modified entrypoint.sh (added set -x) and it worked.
Investigating further..

Raboo:
I was able to add set -x to entrypoint.sh using dirty hack..
`docker stop ceph-osd-sda ; docker start ceph-osd-sda ; docker exec -ti ceph-osd-sda sed -i '/^set -e$/aset -x' /entrypoint.sh`
So now I got bash -x output..

```
+ : ceph
+ :
+ : osd
+ : 0
+ : vm03
+ : vm03
+ : 0
+ : mds-vm03
+ : 0
+ : 100
+ : root=default host=vm03
+ : 0
+ : cephfs
+ : cephfs_data
+ : 8
+ : cephfs_metadata
+ : 8
+ : vm03
+ :
+ :
+ : 8080
+ : 0
+ : 9000
+ : 0.0.0.0
+ : 0.0.0.0
+ : 5000
+ : /api/v0.1
+ : warning
+ : /var/log/ceph/ceph-restapi.log
+ : consul
+ : vm01.se-ix.delta.prod
+ : 8500
+ '[' '!' -z '' ']'
+ CEPH_OPTS='--cluster ceph'
+ MOUNT_OPTS='-t xfs -o noatime,inode64'
+ case "$KV_TYPE" in
+ source /config.kv.sh
++ set -e
++ sed -r s/@CLUSTER@/ceph/g /etc/confd/conf.d/ceph.conf.toml.in
++ [[ consul == \e\t\c\d ]]
++ tr '[:upper:]' '[:lower:]'
++ echo osd
+ CEPH_DAEMON=osd
+ case "$CEPH_DAEMON" in
+ start_osd
+ get_config
+ CLUSTER_PATH=ceph-config/ceph
+ kviator --kvstore=consul --client=vm01.se-ix.delta.prod:8500 get ceph-config/ceph/monSetupComplete
+ confd -onetime -backend consul -node vm01.se-ix.delta.prod:8500 -prefix=/ceph-config/ceph/
2016-07-12T14:51:03Z vm03 confd[72378]: INFO Backend set to consul
2016-07-12T14:51:03Z vm03 confd[72378]: INFO Starting confd
2016-07-12T14:51:03Z vm03 confd[72378]: INFO Backend nodes set to vm01.se-ix.delta.prod:8500
+ mkdir -p /var/lib/ceph/bootstrap-osd /var/lib/ceph/bootstrap-mds /var/lib/ceph/bootstrap-rgw
+ echo 'Adding bootstrap keyrings'
+ kviator --kvstore=consul --client=vm01.se-ix.delta.prod:8500 get ceph-config/ceph/bootstrapOsdKeyring
Adding bootstrap keyrings
+ kviator --kvstore=consul --client=vm01.se-ix.delta.prod:8500 get ceph-config/ceph/bootstrapMdsKeyring
+ kviator --kvstore=consul --client=vm01.se-ix.delta.prod:8500 get ceph-config/ceph/bootstrapRgwKeyring
+ check_config
+ [[ ! -e /etc/ceph/ceph.conf ]]
+ create_socket_dir
+ mkdir -p /var/run/ceph
+ chown ceph. /var/run/ceph
+ '[' 0 -eq 1 ']'
+ case "$OSD_TYPE" in
+ osd_trying_to_determine_scenario
+ [[ ! -d /var/lib/ceph/osd ]]
++ find /var/lib/ceph/osd -prune -empty
+ [[ -n /var/lib/ceph/osd ]]
+ echo 'No bootstrapped OSDs found; trying ceph-disk'
+ osd_disk
+ osd_disk_prepare
No bootstrapped OSDs found; trying ceph-disk
+ [[ -z /dev/sda ]]
+ '[' '!' -e /var/lib/ceph/bootstrap-osd/ceph.keyring ']'
+ timeout 10 ceph --cluster ceph --name client.bootstrap-osd --keyring /var/lib/ceph/bootstrap-osd/ceph.keyring health
HEALTH_WARN 5 pgs degraded; 5 pgs stuck unclean; 5 pgs undersized; recovery 4914/944763 objects degraded (0.520%); 1/35 in osds are down; noout,sortbitwise flag(s) set
+ mkdir -p /var/lib/ceph/osd
+ chown ceph. /var/lib/ceph/osd
++ egrep '^ 1.*ceph data'
++ parted --script /dev/sda print
+ [[ -n  1      1049kB  299GB  299GB  xfs          ceph data ]]
+ [[ 0 -ne 1 ]]
+ echo 'ERROR- It looks like /dev/sda is an OSD, set OSD_FORCE_ZAP=1 to use this device anyway and zap its content'
+ echo 'You can also use the zap_device scenario on the appropriate device to zap it'
+ exit 1
ERROR- It looks like /dev/sda is an OSD, set OSD_FORCE_ZAP=1 to use this device anyway and zap its content
You can also use the zap_device scenario on the appropriate device to zap it
```

leseb:
@Raboo well it looks like `/var/lib/ceph/osd` is simply empty.
Can you confirm?

Raboo:
Looks empty 

```
docker stop ceph-osd-sda ; docker start ceph-osd-sda ; docker exec -ti ceph-osd-sda find /var/lib/ceph/osd ; docker exec -ti ceph-osd-sda find /var/lib/ceph/osd -prune -empty
ceph-osd-sda
ceph-osd-sda
/var/lib/ceph/osd
/var/lib/ceph/osd
```

leseb:
Then if it's empty this is expected.
If we don't have any mounted devices this is a problem.
So please remount all the drives if possible.

Stoping the container won't umount the osd drive, however a reboot will and that's an issue.

Raboo:
This is weird, I just created/started the docker container manually and it worked.

```
docker run -d --net=host --name ceph-osd-sda --privileged=true --pid=host -m 2g --restart no --cpuset-cpus="0,1" -v /etc/ceph:/etc/ceph -v /var/lib/ceph/:/var/lib/ceph/ -v /dev/:/dev/ -e OSD_DEVICE=/dev/sda -e OSD_JOURNAL=/dev/nvme0n1p1 -e KV_TYPE=consul -e KV_IP=vm01.se-ix.delta.prod -e KV_PORT=8500 ceph/daemon:tag-build-master-jewel-ubuntu-14.04 osd
```

So the problem occurs when I use chef to create the docker container..

leseb:
Hum weird, do you a module in Chef or is it raw command that you're calling?

Raboo:
I just noticed the difference. It's the volumes..
Chef log:

```
    - stopping ceph-osd-sda  (will kill after 60s)
    - deleting ceph-osd-sda
    - update ceph-osd-sda
    -   set memory                      to 2048000000 (was 2147483648)
    -   set restart_maximum_retry_count to 3 (was 0)
    -   set restart_policy              to "always" (was "no")
    -   set volumes_binds               to ["/dev/:/dev/"] (was ["/etc/ceph:/etc/ceph", "/var/lib/ceph/:/var/lib/ceph/", "/dev/:/dev/"])
    - starting ceph-osd-sda
```

Raboo:
I guessing I shouldn't mount /etc/ceph and /var/lib/ceph when I use a KV.

Raboo:
When mounted it has a few directories..

```
09:50:22 root@vm03:~# find /var/lib/ceph/osd
/var/lib/ceph/osd
/var/lib/ceph/osd/ceph-38
/var/lib/ceph/osd/ceph-43
/var/lib/ceph/osd/ceph-36
/var/lib/ceph/osd/ceph-41
/var/lib/ceph/osd/ceph-35
/var/lib/ceph/osd/ceph-37
/var/lib/ceph/osd/ceph-40
/var/lib/ceph/osd/ceph-42
/var/lib/ceph/osd/ceph-39
```

leseb:
The bindmount is not needed but recommended in this case. If you don't do it, `/var/lib/ceph/osd` will always be empty that's for sure.

Raboo:
Ok, I feel that it's not needed.

To quote you from IRC "if /var/lib/ceph/osd is empty then we think there is no OSD so we prepare and active the OSD_DEVICE, if not then we assume there are existing OSD and then **only** activate OSD_DEVICE"

I would say, bindmount is only needed when you want to achieve "OSD_TYPE=directory", and thus you would need to have persistence outside the docker container.
And directory type already checks if /var/lib/ceph/osd exists, so if [ -z "${OSD_DEVICE}" ] would be sufficient.

When using OSD_DEVICE=/dev/sd? the folders in /var/lib/ceph/osd/ will always only contain empty ceph-[ID] folders, and before actually running the whole whoami, you don't know the ID. I don't understand why you need to bindmount /var/lib/ceph/osd when it's only used to mount the OSD device to that location and everything gets persisted on the OSD block device.
So in effect you are bind mounting a folder that is always empty.
And if you want a idempotent way of setting up docker OSD's for the first time and the times after when restarting or upgrading the after that, you shouldn't need to create /var/lib/ceph/osd manually for the first time for the sole reason that it should contain empty folders.

I think my changes makes this more idempotent and abides by the rules of "Without OSD_TYPE"

> 
> If the operator does not specify an OSD_TYPE autodetection happens:
> -     disk is used if no bootstrapped OSD is found. OSD_FORCE_ZAP=1 must be set at this point.
> -     activate is used if a bootstrapped OSD is found and OSD_DEVICE is also provided.
> -     directory is used if a bootstrapped OSD is found and no OSD_DEVICE is provided.

leseb:
Your point is valid, I'll go into your direction.
I looked at the code and it's fine to try `osd_directory` when `OSD_DEVICE` is empty.
If `/var/lib/ceph/osd` is not bind mount then the entrypoint will fail because of https://github.com/ceph/ceph-docker/blob/master/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh#L323

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -79,25 +79,19 @@ function dev_part {
...
-    echo "No bootstrapped OSDs found; trying ceph-disk"
-    osd_disk
-    return
-  fi
-
   if [ -z "${OSD_DEVICE}" ]; then
     echo "Bootstrapped OSD(s) found; using OSD directory"
     osd_directory
     return
```
leseb: if we are part of the same `if` you don't need to `return`, this is implied by the `if`

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -79,25 +79,19 @@ function dev_part {
...
-  fi
-
-  if $(parted --script ${OSD_DEVICE} print | egrep -sq '^ 1.*ceph data'); then
+  elif $(parted --script ${OSD_DEVICE} print | egrep -sq '^ 1.*ceph data'); then
     echo "Bootstrapped OSD found; activating ${OSD_DEVICE}"
     osd_activate
+    return
```
leseb: ditto

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -79,25 +79,19 @@ function dev_part {
...
   else
     echo "Device detected, assuming ceph-disk scenario is desired"
     echo "Preparing and activating ${OSD_DEVICE}"
     osd_disk
+    return
```
leseb: ditto

Raboo:
Ok cool, I removed the returns.

PR #340, opened 2016-07-20
Provision a user and bucket with environment variables

deviantintegral:
This PR:

* Optionally sets up a user, keys, and a bucket.
* Includes s3cmd to create the bucket. I'm pretty new to Ceph, and I couldn't find any built-in tools to do this, but if they exist I'd be glad to remove s3cmd. I usually use aws-cli, but it required 20+ MB of packages while s3cmd is less than 1MB.
* Since s3cmd requires a configuration file and doesn't let auth be specified via flags, we include an example configuration.


deviantintegral:
Also, this branch is built over at https://hub.docker.com/r/andrewberry/ceph-docker/, so it can be tested with `docker pull andrewberry/ceph-docker`.

Ulexus:
I'm struggling to the see the intent, here.  

While RGW does supply S3 compatibility, neither the S3 protocol nor the s3cmd tool is required to use it.  What is the intent, here?  Is it to demonstrate use of RGW?

deviantintegral:
We're using ceph as a replacement for S3 for local development, and this container is so far the most reliable solution we've found. With these variables, we can use docker-compose to spin up appropriate credentials and a bucket, without our developers having to learn how to use the ceph CLI tools. With this PR, if your goal is to see if a specific S3 client works against Ceph, it's a very simple process.

deviantintegral:
I've merged https://github.com/ceph/ceph-docker/pull/343 into this PR to trigger a Docker hub build, but I'm happy to revert it and set up a separate build.

Ulexus:
@deviantintegral Sorry for my density, but I'm still not undestanding what you are after.  

There appear to be two parts of your PR:
  - inclusion of s3cmd
  - addition of an rgw user in the demo

To the first, why not simply run s3cmd in another container within the same Pod (or whatever Docker Compose uses to the same effect)?  Or bind-mount it into the container you want?  Or put it in a separate volume and attach it into the container?

For the demo, I think it seems reasonable to include a default demo RGW user.  I haven't looked at the existing one with respect to RGW, though, so I'll pitch that one in @leseb  direction.

Am I missing something?

ceph-releases/jewel/ubuntu/14.04/demo/Dockerfile
```diff
@@ -7,8 +7,17 @@
 FROM ceph/base:tag-build-master-jewel-ubuntu-14.04
 MAINTAINER Sébastien Han "seb@redhat.com"
 
+# Install s3cmd to support creation of buckets
+RUN ["apt-get", "update", "-y"]
```
leseb: we never used brackets in the repository, just for consistency purpose could you please re-use the following schema? https://github.com/ceph/ceph-docker/blob/master/ceph-releases/jewel/ubuntu/14.04/base/Dockerfile

is it a best practice btw?

deviantintegral:
> To the first, why not simply run s3cmd in another container within the same Pod

That's an interesting thought. However, it seems like extra complexity for a demo case, where I would think the first order of business is "how quickly can I get a client talking to ceph?". But if there's strong objections to not provisioning a bucket, and to including a lightweight client app, then I'm OK with pulling it out of this PR.

ceph-releases/jewel/ubuntu/14.04/demo/Dockerfile
deviantintegral: It's preferred for CMD at least, so I've always used it for RUN and EXEC just for consistency.

https://docs.docker.com/engine/reference/builder/#/cmd

ceph-releases/jewel/ubuntu/14.04/demo/entrypoint.sh
```diff
@@ -155,12 +155,7 @@ function bootstrap_osd {
     chown -R ceph. /var/lib/ceph/osd/${CLUSTER}-0
     ceph-osd ${CEPH_OPTS} -i 0 --mkfs --setuser ceph --setgroup ceph
     ceph ${CEPH_OPTS} auth get-or-create osd.0 osd 'allow *' mon 'allow profile osd' -o /var/lib/ceph/osd/${CLUSTER}-0/keyring
-    if [ $(hostname -s) == 'default' ]; then
-      echo "A hostname of 'default' as used by docker-machine conflicts with Ceph."
-      echo "Rename your machine to fix this issue."
-      echo "https://github.com/ceph/ceph-docker/issues/168"
-    fi
-    ceph ${CEPH_OPTS} osd crush add 0 1 root=default host=$(hostname -s)
+    ceph ${CEPH_OPTS} osd crush add 0 1 root=default host=localhost
```
leseb: isn't this #343?

leseb:
I'm good with this one for the demo container.
Thanks!

PR #339, opened 2016-07-20
Throw a warning about hostname -s == 'default'

deviantintegral:
This PR doesn't solve the issue, but at least gives users a hint as to what's going on and how to fix it.

```
$ docker run --net=host -e MON_IP=192.168.99.101 -e CEPH_PUBLIC_NETWORK=192.168.99.0/24 b5d4a5cf6fa2
creating /etc/ceph/ceph.client.admin.keyring
creating /etc/ceph/ceph.mon.keyring
monmaptool: monmap file /etc/ceph/ceph.monmap
monmaptool: set fsid to c28732cf-18a8-45cd-98ff-a218bd18bb2b
monmaptool: writing epoch 0 to /etc/ceph/ceph.monmap (1 monitors)
creating /tmp/ceph.mon.keyring
importing contents of /etc/ceph/ceph.client.admin.keyring into /tmp/ceph.mon.keyring
importing contents of /etc/ceph/ceph.mon.keyring into /tmp/ceph.mon.keyring
ceph-mon: set fsid to c28732cf-18a8-45cd-98ff-a218bd18bb2b
ceph-mon: created monfs at /var/lib/ceph/mon/ceph-default for mon.default
'/var/lib/ceph/mon/ceph-default' already exists and is not empty: monitor may already exist
set pool 0 size to 1
0
2016-07-20 16:55:57.969410 7f20ffbf4800 -1 journal FileJournal::_open: disabling aio for non-block journal.  Use journal_force_aio to force use of aio anyway
2016-07-20 16:55:57.991209 7f20ffbf4800 -1 journal FileJournal::_open: disabling aio for non-block journal.  Use journal_force_aio to force use of aio anyway
2016-07-20 16:55:57.992330 7f20ffbf4800 -1 filestore(/var/lib/ceph/osd/ceph-0) could not find #-1:7b3f43c4:::osd_superblock:0# in index: (2) No such file or directory
2016-07-20 16:55:58.004823 7f20ffbf4800 -1 created object store /var/lib/ceph/osd/ceph-0 for osd.0 fsid c28732cf-18a8-45cd-98ff-a218bd18bb2b
A hostname of 'default' as used by docker-machine conflicts with ceph.
Rename your machine to fix this issue.
https://github.com/ceph/ceph-docker/issues/168
Error EINVAL: (22) Invalid argument
```

PR #343, opened 2016-07-21
Use localhost since everything runs in one container

deviantintegral:
A followup from https://github.com/ceph/ceph-docker/pull/339

This is working for me with the default hostname for docker-machine. Is there any reason we can't just lock this down to localhost?

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/demo/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/demo/entrypoint.sh
@@ -155,12 +155,7 @@ function bootstrap_osd {
     chown -R ceph. /var/lib/ceph/osd/${CLUSTER}-0
     ceph-osd ${CEPH_OPTS} -i 0 --mkfs --setuser ceph --setgroup ceph
     ceph ${CEPH_OPTS} auth get-or-create osd.0 osd 'allow *' mon 'allow profile osd' -o /var/lib/ceph/osd/${CLUSTER}-0/keyring
-    if [ $(hostname -s) == 'default' ]; then
-      echo "A hostname of 'default' as used by docker-machine conflicts with Ceph."
-      echo "Rename your machine to fix this issue."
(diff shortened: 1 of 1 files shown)


ceph-releases/jewel/ubuntu/14.04/demo/entrypoint.sh
```diff
@@ -155,12 +155,7 @@ function bootstrap_osd {
     chown -R ceph. /var/lib/ceph/osd/${CLUSTER}-0
     ceph-osd ${CEPH_OPTS} -i 0 --mkfs --setuser ceph --setgroup ceph
     ceph ${CEPH_OPTS} auth get-or-create osd.0 osd 'allow *' mon 'allow profile osd' -o /var/lib/ceph/osd/${CLUSTER}-0/keyring
-    if [ $(hostname -s) == 'default' ]; then
-      echo "A hostname of 'default' as used by docker-machine conflicts with Ceph."
-      echo "Rename your machine to fix this issue."
-      echo "https://github.com/ceph/ceph-docker/issues/168"
-    fi
-    ceph ${CEPH_OPTS} osd crush add 0 1 root=default host=$(hostname -s)
+    ceph ${CEPH_OPTS} osd crush add 0 1 root=default host=localhost
```
leseb: Using localhost will bring a lot of confusion from a Ceph perspective.
If all of your OSDs are running on the host localhost this might raise some issues, I'd like to keep your previous patch.

deviantintegral: Does the container support running multiple OSDs within the same container instance? It didn't seem to me that was really a use case supported by the demo.

leseb: Oops I didn't notice this was the demo container :).
It's still not clear to me what you're trying to solve here...

PR #347, opened 2016-07-26
entrypoint: accept /dev/disk/by-x names for OSD_DEVICE

immesys:
I was having some difficulty on some servers that would somewhat arbitrarily reorganize the /dev/sdx devices. I solved it on one server by fully specifying /dev/disk/by-x names for the OSD device and journal, but on some of the servers, I wanted to just specify OSD_DEVICE and let the journal selection be automatic. The process is essentially 

- find the actual device the symlink points to
- form the requested partition of that device as a /dev/sdx name
- find the symlink that is most similar to the original symlink that points to the partition we want

This allows you to use wwn or label or pci path or whatever symlink to specify your disk, and the automatic journal choice will be an equivalent symlink.

This is pretty easy to test, just source the functions into your shell and invoke dev_part a couple times. I also tested it in production.

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -63,6 +63,17 @@ function check_admin_key {
   fi
 }
 
+# Given two strings, return the length of the shared prefix
+function prefix_length {
+  local maxlen=${#1}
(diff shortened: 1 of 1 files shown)


immesys:
Do you have some feedback on this PR or something I can do to ease merging?

leseb:
Sorry I was on PTO, this looks good to me, thanks!

PR #349, opened 2016-07-28
Changes to allow for clusters not named ceph

danielin917:
Signed-off-by: daniel lin <danielin@umich.edu>

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -246,8 +246,8 @@ function start_mon {
       exit 1
     fi
 
-    if [ ! -e /etc/ceph/monmap ]; then
-      echo "ERROR- /etc/ceph/monmap must exist.  You can extract it from your current monitor by running 'ceph mon getmap -o /etc/ceph/monmap' or use a KV Store"
+    if [ ! -e /etc/ceph/monmap-${CLUSTER} ]; then
+      echo "ERROR- /etc/ceph/monmap-${CLUSTER} must exist.  You can extract it from your current monitor by running 'ceph mon getmap -o /etc/ceph/monmap-<cluster>' or use a KV Store"
       exit 1
     fi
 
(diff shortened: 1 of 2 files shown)


immesys:
I think this will conflict with my changes in #346

immesys:
Will this make existing monitors fail to start if they pull an updated container?

danielin917:
get_mon_config is called right before so if the monmap is incorrect or doesn't exist the correct one should be created before it is declared an error by the entrypoint. As of now I have not tested on an old monitor but in theory it shouldn't matter and the old monmap would be left behind as junk. The changes in #346 are equivalent but since we have CEPH_OPTS already defined I figured it was a cleaner solution.

dang:
If there is only one mon, then @danielin917 is correct, and the newly created monmap will be identical to the old one, and the cluster should be unaffected.  If there are multiple monitors, then that monmap is largely irrelevant, since the updated mon will connect to the existing mons and get a new, valid monmap.

It would certainly be possible to check if $CLUSTER == ceph and move monmap to monmap-ceph.

leseb:
@danielin917 can you rebase your change?

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -246,8 +246,8 @@ function start_mon {
       exit 1
     fi
 
-    if [ ! -e /etc/ceph/monmap ]; then
-      echo "ERROR- /etc/ceph/monmap must exist.  You can extract it from your current monitor by running 'ceph mon getmap -o /etc/ceph/monmap' or use a KV Store"
+    if [ ! -e /etc/ceph/monmap-${CLUSTER} ]; then
```
leseb: For existing clusters, this file will never exit, can we keep the previous check and then `mv` with the desired cluster name? Then we can safely move to this check and just do nothing.

dang: Good point.  Seems like a good idea.

dang: Okay, the fix is now it get_mon_config(), so this should succeed.

immesys: You need to change all instances of get_mon_config not just in config.static. This has broken my cluster.

danielin917:
Pull request has been submitted

Issue #342, opened 2016-07-21
Update Kubernetes example to use a pinned versions of official ceph-docker releases


elsonrodriguez:
The current kubernetes example was merged in using images built by @cornelius-keller, at some point the example needs to be updated to an image built by ceph-docker, and pinned to a unique tag (not latest)

This will ensure that the example continues to work for people who wish to try it.

leseb:
@elsonrodriguez can we close this thanks to #353?

elsonrodriguez:
It's an improvement, but still not ideal.

The situation that needs to be avoided is someone trying the example and having it not work due to a change in the container image. 

A solution could be either of these things:

- A special tag for ceph-daemon (for example `tag-build-k8s-example`) which is updated when someone validates the k8s example against it.

- A more granular versioning system for all containers in the repo, so that a tagged version persists for as long as hub.docker.com allows it, and have the k8s example pinned to a version. The k8s example could then be updated as needed.

I would lean more towards granular, permanent tags and then have people update the k8s example when they can. However something like the special tag may be easier to implement, and would ensure stability in the near term until more granular versions are available.

leseb:
We can work with tag for k8s, I'm just not sure if they are necessary. Everytime we push a new tag, everything should be functional.
We still don't have a first stable release to push since we don't have the necessary bits to test all the scenarios.

PR #353, opened 2016-08-10
Use the official ceph-docker images for Kubernetes examples

hunter:
A minor update to the Kubernetes resources to use the official repo. It is using the `latest` tag but should partly solve the issue raised in #342


leseb:
Thanks @hunter

PR #356, opened 2016-08-12
Add support for NFS Gateway

dang:
Add the Ganesha NFS gateway to the daemon.  This allows accessing the
RadosGW via NFS.

Signed-off-by: Daniel Gryniewicz <dang@redhat.com>

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -25,6 +25,7 @@ set -e
 : ${RGW_REMOTE_CGI:=0}
 : ${RGW_REMOTE_CGI_PORT:=9000}
 : ${RGW_REMOTE_CGI_HOST:=0.0.0.0}
+: ${RGW_USER:="cephnfs"}
 : ${RESTAPI_IP:=0.0.0.0}
 : ${RESTAPI_PORT:=5000}
 : ${RESTAPI_BASE_URL:=/api/v0.1}
@@ -33,6 +34,8 @@ set -e
 : ${KV_TYPE:=none} # valid options: consul, etcd or none
 : ${KV_IP:=127.0.0.1}
 : ${KV_PORT:=4001} # PORT 8500 for Consul
+: ${GANESHA_OPTIONS:=""}
+: ${GANESHA_EPOCH:=""} # For restarting
 
 if [ ! -z "${KV_CA_CERT}" ]; then
 	KV_TLS="--ca-cert=${KV_CA_CERT} --client-cert=${KV_CLIENT_CERT} --client-key=${KV_CLIENT_KEY}"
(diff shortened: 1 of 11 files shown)


leseb:
A couple of things:

* can you also update the README?
* this: https://github.com/ceph/ceph-docker/blob/master/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh#L917 needs to be update to
* can we have a 'demo' container version too?

Thanks!

leseb:
Thanks for updating.
Code changed in the meantime because I just merged #352, I was referring to this line: https://github.com/ceph/ceph-docker/blob/master/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh#L947

dang:
Ah, right.  Fixed.

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -913,8 +965,8 @@ case "$CEPH_DAEMON" in
   if [ ! -n "$CEPH_DAEMON" ]; then
     echo "ERROR- One of CEPH_DAEMON or a daemon parameter must be defined as the name "
     echo "of the daemon you want to deploy."
-    echo "Valid values for CEPH_DAEMON are MON, OSD, OSD_DIRECTORY, OSD_CEPH_DISK, OSD_CEPH_DISK_PREPARE, OSD_CEPH_DISK_ACTIVATE, OSD_CEPH_ACTIVATE_JOURNAL, MDS, RGW, RESTAPI, ZAP_DEVICE, RBD_MIRROR"
-    echo "Valid values for the daemon parameter are mon, osd, osd_directory, osd_ceph_disk, osd_ceph_disk_prepare, osd_ceph_disk_activate, osd_ceph_activate_journal, mds, rgw, restapi, zap_device, rbd_mirror"
+    echo "Valid values for CEPH_DAEMON are MON, OSD, OSD_DIRECTORY, OSD_CEPH_DISK, OSD_CEPH_DISK_PREPARE, OSD_CEPH_DISK_ACTIVATE, OSD_CEPH_ACTIVATE_JOURNAL, MDS, RGW, RGW_USER, RESTAPI, ZAP_DEVICE, RBD_MIRROR"
+    echo "Valid values for the daemon parameter are mon, osd, osd_directory, osd_ceph_disk, osd_ceph_disk_prepare, osd_ceph_disk_activate, osd_ceph_activate_journal, mds, rgw, rgw_user, restapi, zap_device, rbd_mirror"
```
leseb: and `nfs`?

dang:
Sorry.  I saw RGW and read it as NFS...

ktdreyer:
I don't see nfs-ganesha-rgw in EPEL 7 http://dl.fedoraproject.org/pub/epel/7/x86_64/n/ . Where does this package come from?

Why include nfs-ganesha-vfs ? Does the RGW FSAL depend on that?

dang:
EPEL only includes the base FSALs for Ganesha, currently.  This is because it doesn't currently include Ceph or Gluster.

I included VFS in the package for testing purposes:  That way I (or a user) could test Ganesha independent of the Ceph cluster, and therefore verify each step separately.  VFS is not necessary for RGW or CEPH to work.

ktdreyer:
If EPEL doesn't have the -rgw FSAL, does that mean this change will cause the yum operation to fail?

Issue #247, opened 2016-04-13
Branch management, distro support and CI


leseb:
Dear contributors,

@theanalyst ,@walkerlee ,@mbessler ,@minshenglin ,@bnoon ,@a-nldisr ,@rochaporto ,@BugRoger ,@tobilg ,@tobegit3hub ,@xcompass ,@nazar-pc ,@lorieri ,@CoRfr ,@fmeppo ,@TelanJ ,@hookenz ,@thecanadianbaker ,@rootfs ,@h0tbird ,@streamnsight ,@hunter ,@dmitryint, @Ulexus, @jimcurtis 

Many users reported this issue already and finally got the time to start addressing it. We don't have any proper release management. We are missing a couple of things:

* No branch management
* No CI

# Branch management

If we take a little step back, and for those of you who are not familiar with the repo entirely. We are only putting efforts and `ceph/base` and `ceph/daemon`.
Why do we need branches? We need to support several distributions **and** Ceph versions as well. However one of the main issue we are facing is the Docker Hub. Basically each branch in our git repo reflects one image and we don't have the ability to do some templating with the `Dockerfile`. We explicitly need to set values and can not do any sort of conditions (if distro then etc etc). So for each OS and each Ceph version we need to modify those:

* https://github.com/ceph/ceph-docker/blob/master/base/Dockerfile#L5
* https://github.com/ceph/ceph-docker/blob/master/base/Dockerfile#L10

All the installation steps: https://github.com/ceph/ceph-docker/blob/master/base/Dockerfile#L24-L28
And eventually modify: https://github.com/ceph/ceph-docker/blob/master/daemon/Dockerfile#L7 accordingly. This is really painful and not practical to manage branches with all of that, spending time to resolve conflicts every time we are trying to merge master into a branch.
So because of all of that we need to arrange our own workflow.
After discussing with one of my colleague (@ErwanAliasr1), we ended up with what I believe could be a pretty good way to manage our branches.

The basic idea is to manage **everything** from the master branch, so would have several files in the base directory with this format `Dockerfile-<ceph version>-<distro>`, ie `Dockerfile-infernalis-ubuntu`.
Then every time someone wants to PR something it will end up in master.
After this, repo's owner (and later a CI) will execute a script that will:

* create branches like this: `<ceph-version>-<distro>`, ie: `infernalis-ubuntu`.
* push each  `Dockerfile-<ceph version>-<distro>` into their respective branch by simply copying it to `Dockerfile``

[...]

PR #361, opened 2016-08-18
Fix daemon entrypoint script for non-Ubuntu distros.

mgilham:
It looks like the current entrypoint script will only work on Ubuntu, and that my_init is designed to provide distribution-independence.

I tested this change on CentOS 7 (and Kubernetes).

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -436,7 +436,7 @@ EOF
     chmod +x /etc/service/${CLUSTER}-${OSD_ID}/run
   done
 
-exec /usr/bin/runsvdir -P /etc/service
+exec /sbin/my_init
 }
 
 


hookenz:
Out of curiosity, why are the ceph docker team spending effort maintaining multiple distributions inside the container?  I think that misses the point of docker.  Who cares what environment is used inside the container.  You're not actually running an OS inside of it, you're only running the binary. It's just an Ubuntu or CentOS layout with various libraries.  But I think the Ubuntu environment is more up to date and the docker base is much smaller than CentOS base.  My 2c.

Ulexus:
@hookenz It's not really volitious, it's just that I wrote this originally for the Ubuntu image, but now it's mostly maintained by @leseb and other Redhat guys.  No one has wanted to cause upset by forcing everyone over to CentOS, but the desire from the newer folks has been to use a Redhat-like system.

leseb:
@hookenz please read https://github.com/ceph/ceph-docker/issues/247
The general idea is to have the ability to support multiple vendors. Companies that need professional services support, can only do this with a vendor and its own supported OS. Thus Red Hat, Suse or others will never support containers running on Ubuntu. Simple as that :).

The way everything is setup at the moment is that all the entrypoints point to Ubuntu as a reference.

leseb:
@mgilham not sure to understand the purpose of the commit. The commit message says "only works on Ubuntu" but the commit change points to Ubuntu. And then you said you tested this on CentOS (but still applying a change to Ubuntu). You lost me on this one.

khogeland:
>The commit message says "only works on Ubuntu" but the commit change points to Ubuntu. And then you said you tested this on CentOS (but still applying a change to Ubuntu).

The change applies to all images; CentOS and Fedora symlink to the entrypoint script from the Ubuntu image:
https://github.com/ceph/ceph-docker/blob/master/ceph-releases/jewel/fedora/24/daemon/entrypoint.sh
https://github.com/ceph/ceph-docker/blob/master/ceph-releases/jewel/centos/7/daemon/entrypoint.sh

leseb:
My bad, looks like I was still sleepy...

hookenz:
Sure. I can understand.  But it does create extra work for you.     The container is meant to be like a black box that will run on any host that supports docker. Oh well...

PR #362, opened 2016-08-18
entrypoint: stricter osd PID check

### LINKED ISSUE #358: entrypoint: mixup between osd processes
Towards the end of the osd activation ([here](https://github.com/ceph/ceph-docker/blob/492fe894430c1eb53bcc98bd858ed8598dc6ad3e/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh#L553)) the entrypoint script attempts to get the PID of the recently launched ceph osd process. 

The problem is that the filter here is too general. The recommended docker incantation for starting the daemon includes --pid=host which means that ps gives all the processes. This can lead to multiple PIDs matching, which crashes the test below with 

```
/entrypoint.sh: line 556: [: too many arguments
```

I happened to hit this because my docker --name matched my unit (which was ceph-osd-0.service) and I was also be running journalctl -f on the unit, so there were three processes that matched.

I imagine though that people could also hit this with osd.1 clashing with osd.10 etc.

I'm a little overworked atm so I can't PR this any time soon.

immesys:
This fixes #358 

It is similar to the regex suggested by @leseb with the following changes:
- it tests that the user is ceph, which translates to some arbitrary uid that is shared by instances of this container. This means that we are only contending with other instances of the container, not other commands.
- it checks the $CLUSTER in case there are two osds with the same IDs on the same host but in different clusters

I tested it and it appears to work both when there is a PID (I deliberately launched the container twice) and when there isn't. I never encountered a case where the PID was present without having had a duplicate container. The comments seem to suggest that ceph-disk activate may launch it? It's not a race, I added a delay to make sure.

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -550,7 +550,7 @@ function osd_activate {
 
   # ceph-disk activiate has exec'ed /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID}
   # wait till docker stop or ceph-osd is killed
-  OSD_PID=$(ps -ef |grep ceph-osd |grep osd.${OSD_ID} |awk '{print $2}')
+  OSD_PID=$(pgrep -U ceph -f "^/usr/bin/ceph-osd \-\-cluster $CLUSTER.*\-i ${OSD_ID} \-\-setuser")
   if [ -n "${OSD_PID}" ]; then
       echo "OSD (PID ${OSD_PID}) is running, waiting till it exits"
       while [ -e /proc/${OSD_PID} ]; do sleep 1;done


ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -550,7 +550,7 @@ function osd_activate {
 
   # ceph-disk activiate has exec'ed /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID}
   # wait till docker stop or ceph-osd is killed
-  OSD_PID=$(ps -ef |grep ceph-osd |grep osd.${OSD_ID} |awk '{print $2}')
+  OSD_PID=$(pgrep -U ceph -f "^/usr/bin/ceph-osd \-\-cluster $CLUSTER.*\-i ${OSD_ID} \-\-setuser")
```
leseb: Just for consistency, can we keep `{CLUSTER}`?

immesys:
Oops, this slipped off my stack completely. I'll make that change now

leseb:
Thanks @immesys I think we are good with this.

PR #371, opened 2016-09-07
entrypoint: ignore pgrep exit code

### LINKED ISSUE #370: OSD entrypoint script exits when pgrep fails
I'm trying to get some OSDs up and running on CoreOS. They fail when I try to run `activate` and I believe the `entrypoint.sh` script has a small bug.

```
Sep 06 08:33:41 test-storage-001 docker[15543]: create-or-move updated item name 'osd.0' weight 0.01 at location {host=test-storage-001,root=default} to crush map
Sep 06 08:33:41 test-storage-001 systemd[1]: ceph-osd.service: Main process exited, code=exited, status=1/FAILURE
```

https://github.com/ceph/ceph-docker/blob/master/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh#L556

When the entrypoint script reaches that line it is aborting, I assume because `pgrep` exits with status 1 when there are no result.

```
root@test-storage-001:/# pgrep -U ceph -f "^/usr/bin/ceph-osd \-\-cluster ceph.*\-i 0 \-\-setuser"
root@test-storage-001:/# echo $?
1
```

When I patch in a hacked script that skips the check and just runs the `else` case, the OSD starts as expected.

immesys:
pgrep can exit 1 if no results are found

This should fix #370

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -553,7 +553,7 @@ function osd_activate {
 
   # ceph-disk activiate has exec'ed /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID}
   # wait till docker stop or ceph-osd is killed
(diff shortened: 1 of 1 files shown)


immesys:
I am busy testing this now. Will report when I am done

immesys:
I reproduced #370 and with this applied, it now works. Sorry everyone.

linuxwind:
+1

leseb:
Thanks @immesys

PR #374, opened 2016-09-09
fixed #373: config.kv.sh: conf file path error

### LINKED ISSUE #373: config.kv.sh: hammer ceph conf file path error
ceph.conf.toml.in file in /etc/confd/conf.d   not  /etc/confd/src

linuxwind:
closes: #373


leseb:
the build doesn't pass, I don't think it's related to your patch

PR #376, opened 2016-09-09
adds sample files for running kubernetes on coreos

coffeepac:
this is an accumlation of files in the coreos example directory
and additional pieces required to make it work.  the goal was to
have a simple and reliable way to both run ceph on kubernetes on
coreos and be able to access ceph resources (CephFS and RBD) from
any kubernetes node.


leseb:
Thanks @coffeepac

elsonrodriguez:
@cornelius-keller This might be of interest to you

PR #382, opened 2016-09-21
Fix osd startup failure in fedora

### LINKED ISSUE #381: Runit exits with status 127
With the latest master, using fedora 24 images, not able to start osd and it exits with 
```
*** Booting runit daemon...
*** Runit started as PID 20
*** Runit exited with status 127
sh: /usr/bin/sv: No such file or directory
sh: /usr/bin/sv: No such file or directory
*** Killing all processes...
```

I see fixes #334 and #336, but seems like fedora 24 base image does not seem to have runit included in it.

ganeshmaharaj:
Runit is not part of the default package in fedora. This patch builds it
from source in the fedora base image.
This fixes #381 

Signed-off-by: Ganesh Maharaj Mahalingam <ganesh.mahalingam@intel.com>

### DIFF
--- a/ceph-releases/jewel/fedora/24/base/Dockerfile
+++ b/ceph-releases/jewel/fedora/24/base/Dockerfile
@@ -35,6 +35,12 @@ RUN dnf install -y ceph ceph-radosgw rbd-mirror nfs-ganesha-rgw nfs-ganesha-vfs 
 ADD https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl /usr/local/bin/kubectl
 RUN chmod +x /usr/local/bin/kubectl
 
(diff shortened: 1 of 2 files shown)


ganeshmaharaj:
I understand the patch does the bad think of symlinking files to /usr/bin. Please let me know if you rather have 'my_init' script do a shutil.which('sv') instead of this and use that for making the calls.

leseb:
I like the second option better, please keep the binary where it should be (`/usr/sbin` IIRC) and do a which in `my_init`. Thanks!

ganeshmaharaj:
Thanks @leseb. Fixed the patch to do the second option. When compiled from source, it sticks it into /usr/local/bin.

leseb:
Thanks @ganeshmaharaj

PR #383, opened 2016-09-22
add rhel image

leseb:
@dang @rootfs @font @jimcurtis 

[ci skip]


font:
I see references to version 1.3 so it looks like you are merging from the rhcs `master` branch. The latest code we've been working with is on the `v2.0` branch so I would use that.

leseb:
Thanks @font fixed this.

PR #388, opened 2016-09-30
upgrade examples/kubernetes/generator/ceph-key.py to python 3

slang800:
Noticed this when the script broke on my arch-linux box. Python3 is pretty much the de-facto version of Python nowadays, right?

Also I found some dead imports and a markdown file in the generator folder that I missed in #385.


leseb:
Thanks @slang800

PR #391, opened 2016-10-05
Remove nfs-ganesha-rgw for Fedora

hunter:
It looks like I was a little too eager. `nfs-ganesha-rgw` is currently not being built for Fedora.

This should fix the image build error


ktdreyer:
Thanks for catching this. Looks like the Fedora maintainer enabled the RGW FSAL at one point, and then turned it off before the 2.4.0 GA build. I've requested they turn it back on at https://bugzilla.redhat.com/1388690

PR #392, opened 2016-10-06
Removing runit

hunter:
It seems we've had a few issues with runit recently and thought I'd start a discussion (with code) for switching it out for something else.

We've used forego (a go based foreman replacement) in other projects so decided to test it out an alternative for launching multiple processes in ceph-docker. It's currently untested but thought I'd put it out there first before spending more time getting it ready.

The main change is that forego will run the processes specified in a 'Procfile' which is generated slightly differently to the runit. Overall there are slightly fewer dependencies with a single binary that will work across all base OSes. Logs are still sent to stdout/err with each line prefixed with the process name (cluster name and osd ID).

Thoughts? Not a visible change but may simplify things slightly

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
@@ -388,6 +388,10 @@ function osd_directory {
     echo "created folder /var/lib/ceph/osd/${CLUSTER}-${OSD_ID}"
   fi
 
+	# Create the directory and an empty Procfile
+	mkdir -p /etc/forego/${CLUSTER}
+	echo "" > /etc/forego/${CLUSTER}/Procfile
+
   for OSD_ID in $(ls /var/lib/ceph/osd |  awk 'BEGIN { FS = "-" } ; { print $2 }'); do
     if [ -n "${JOURNAL_DIR}" ]; then
        OSD_J="${JOURNAL_DIR}/journal.${OSD_ID}"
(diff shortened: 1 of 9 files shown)


ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -617,9 +612,12 @@ function osd_disks {
         echo "Device ${OSD_DEV} is corrupt for /var/lib/ceph/osd/${CLUSTER}-${OSD_ID}"
         exit 1
       fi
-      set_osd_run ${CLUSTER} ${OSD_ID}
+
+			echo "${CLUSTER}-${OSD_ID}: /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID} --setuser ceph --setgroup disk" | tee -a /etc/forego/${CLUSTER}/Procfile
```
leseb: what's that indent?

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -617,9 +612,12 @@ function osd_disks {
...
+
     done
-    exec /sbin/my_init
+
+		exec /usr/local/bin/forego start -f /etc/forego/${CLUSTER}/Procfile
```
leseb: ditto

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -652,9 +650,14 @@ function osd_disks {
           echo "OSD (PID ${OSD_PID}) is running, waiting till it exits"
           while [ -e /proc/${OSD_PID} ]; do sleep 1;done
       fi
-      set_osd_run ${CLUSTER} ${OSD_ID}
+
+			echo "${CLUSTER}-${OSD_ID}: /usr/bin/ceph-osd ${CEPH_OPTS} -f -d -i ${OSD_ID} --setuser ceph --setgroup disk" | tee -a /etc/forego/${CLUSTER}/Procfile
```
leseb: ditto

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
```diff
@@ -652,9 +650,14 @@ function osd_disks {
...
+
+
     done
-    exec /sbin/my_init
+
+
+		exec /usr/local/bin/forego start -f /etc/forego/${CLUSTER}/Procfile
```
leseb: ditto

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
hunter: Not quite sure actually. Something got messed up along the way

leseb:
This looks like an interesting change, I'm not strongly against it, it's true that the binary makes things much easier so we don't have to find dependancies for all distros. The CI seems to be failing building the image though, see https://travis-ci.org/ceph/ceph-docker/builds/165453217#L1322

ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh
leseb: Thanks!

hunter:
Made a few changes to fix the build errors and rogue tabs. If this is something we want to use then I can update any missing entrypoints (can't remember if I got everything)

leseb:
@hunter Looks good to me, sticking with Jewel is fine for most of the new changes.

PR #394, opened 2016-10-06
Update to etcdctl 2.3.7

hunter:
etcdctl is a few releases behind the latest. Updating to avoid any issues with newer versions


leseb:
Unrelated CI error

PR #405, opened 2016-10-25
tests: update requirements file

font:
Signed-off-by: Ivan Font <ifont@redhat.com>


font:
@alfredodeza Please take a look. I've updated in order to install all required dependencies to reproduce some of the issues filed.

ktdreyer:
This hard-codes a lot of version numbers, which has caused trouble for other projects in the past (see s3tests or teuthology). What were the specific libs you needed?

font:
Fair enough. I've updated the PR to include only the libs without versions and without its dependencies.

alfredodeza:
@font this is really odd. You mentioned some issues were filed but I couldn't find anything related to this PR. Mind listing them here?

Nothing other than py.test and docker-py is needed. We don't use anything else. `virtualenv` could be just used from the distro directly and not meant to install from a requirements file.

font:
I am no longer running into any module dependency issues. I am closing this PR. Thanks!

PR #409, opened 2016-10-27
Add common logging function

font:
I'm only applying this to Jewel for now.

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/config.k8s.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/config.k8s.sh
@@ -3,7 +3,7 @@ set -e
 
 function get_admin_key {
    # No-op for static
-   echo "k8s: does not generate admin key. Use secrets instead."
+   log "k8s: does not generate admin key. Use secrets instead."
 }
 
 function get_mon_config {
@@ -20,6 +20,6 @@ function get_mon_config {
 
 function get_config {
    # No-op for static
-   echo "k8s: config is stored as k8s secrets."
+   log "k8s: config is stored as k8s secrets."
 }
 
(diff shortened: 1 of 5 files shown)


ceph-releases/jewel/ubuntu/14.04/daemon/config.k8s.sh
```diff
@@ -1,9 +1,20 @@
 #!/bin/bash
 set -e
 
+# log arguments with timestamp
+function log {
```
alfredodeza: since you are adding a log function, can you make it so that we have different levels? no definition in levels is going to make it hard to understand what is informational vs error reporting

font: Once we have a common logging function, we can start to add more features to it. So I don't want to add different levels to this PR. Most of the error cases typically add `ERROR` to the string. But since that isn't always consistent, we'd likely either have to update all the strings to prefix some sort of debug level, or pass an extra argument to the log function.

leseb: +1 it'd be nice to have more granularity later. Ideally things like function name being executed etc. We can start slowly with this PR.

leseb: Oh and the entrypoint.sh sources all the other entrypoints so we don't need to replicate this function, just keep it in `entrypoint.sh`

alfredodeza:
This looks OK to me

ceph-releases/jewel/ubuntu/14.04/daemon/config.k8s.sh
font: Oops, yes. I've removed it now.

font:
Also, here is a sample of a successful run:

```
2016-10-31 16:56:22  /entrypoint.sh: SUCCESS
```

And a sample of an error case:
```
2016-10-31 16:56:40  /entrypoint.sh: static: does not generate config
2016-10-31 16:56:40  /entrypoint.sh: ERROR- /etc/ceph/ceph.conf must exist; get it from your existing mon
```

leseb:
tes this please

leseb:
test this please

font:
test this please

Issue #414, opened 2016-10-31
`/var/lib/ceph` is owned by Ceph in Hammer for 16.04 image


alfredodeza:
_________ TestHammer.test_socket_dir_is_owned_by_root[ceph/daemon:tag-build-master-hammer-ubuntu-16.04] ________
    
    self = <test_monitors.TestHammer object at 0x7f133fa8e150>, hammer_containers = {'Id': '92da58573c98107944218b3d4d834b11d49064332630b30df4d9a5c32e6c9055', 'Warnings': None}
    client = <docker.client.Client object at 0x7f133fafca10>
    
        def test_socket_dir_is_owned_by_root(self, hammer_containers, client):
            command = client.exec_create(hammer_containers, cmd='ls -ld /var/run/ceph')
            result = client.exec_start(command)
            if client.exec_inspect(command)['ExitCode'] != 0:
                raise AssertionError(result)
    >       assert result.split()[2] == 'root'
    E       assert 'ceph' == 'root'
    E         - ceph
    E         + root
    
    base/test_monitors.py:57: AssertionError

PR #417, opened 2016-10-31
remove ubuntu 16.04 for hammer release

leseb:
Ubuntu 16,04 comes with the Jewel release of Ceph so there is no point
of forcing hammer's installation.

Closes: #414


alitvak69:
There is a problem with that.  You still can install hammer image for 16.04.  However it install jewel.   Didn't realize until it upgraded osd on my test cluster.

leseb:
@alitvak69 what do you mean? Can you be more specific?

alitvak69:
There is a hammer image for 16.0.4.  However when you run it,  it installs jewel.

PR #423, opened 2016-10-31
Prune unsupported Fedora 23

alfredodeza:
And corrects some mismatched tags for Jewel/Hammer tests


alfredodeza:
@font or @leseb can I get a review for this?

Issue #420, opened 2016-10-31
No monmap exists for daemon:tag-build-master-jewel-ubuntu-16.04


alfredodeza:
__ TestJewel.test_monmap_exists[ceph/daemon:tag-build-master-jewel-ubuntu-16.04] __
    
    self = <test_monitors.TestJewel object at 0x7f7cd2b1b110>, jewel_containers = {'Id': 'ae08c848ae9432402b110b9b10d4f05d178dc1e200f4867006fed71406b6f05b', 'Warnings': None}
    client = <docker.client.Client object at 0x7f7cd2ba8690>
    
        def test_monmap_exists(self, jewel_containers, client):
            command = client.exec_create(jewel_containers, cmd='stat /etc/ceph/ceph.monmap')
            result = client.exec_start(command)
            if client.exec_inspect(command)['ExitCode'] != 0:
    >           raise AssertionError(result)
    E           AssertionError: stat: cannot stat '/etc/ceph/ceph.monmap': No such file or directory
    
    base/test_monitors.py:47: AssertionError

leseb:
It's because filename is `monmap-ceph`, looks like a CI issue to me.

Issue #419, opened 2016-10-31
Jewel Fedora 24 missing /etc/ceph/ceph.monmap


alfredodeza:
Similar to issue #418 there are no monmap files for Jewel Fedora 24 on the daemon image:

    __ TestJewel.test_monmap_exists[ceph/daemon:tag-build-master-jewel-fedora-24] __
    
    self = <test_monitors.TestJewel object at 0x7fc974116590>, jewel_containers = {'Id': '822ccb18c94a705745d1760603809c50e93d3c534166cf33cc8946be65b08e53', 'Warnings': None}
    client = <docker.client.Client object at 0x7fc97424d090>
    
        def test_monmap_exists(self, jewel_containers, client):
            command = client.exec_create(jewel_containers, cmd='stat /etc/ceph/ceph.monmap')
            result = client.exec_start(command)
            if client.exec_inspect(command)['ExitCode'] != 0:
    >           raise AssertionError(result)
    E           AssertionError: stat: cannot stat '/etc/ceph/ceph.monmap': No such file or directory
    
    base/test_monitors.py:47: AssertionError

font:
@alfredodeza This appears to be the same issue you reported in #422.

leseb:
@alfredodeza why are you testing `/etc/ceph/ceph.monmap` instead of `/etc/ceph/monmap-ceph`?

PR #428, opened 2016-11-02
test: fix monmap name check

leseb:
Closes: #421, #419  and #420

### DIFF
--- a/tests/base/test_monitors.py
+++ b/tests/base/test_monitors.py
@@ -30,6 +30,12 @@ class TestAll(object):
         if client.exec_inspect(command)['ExitCode'] != 0:
             raise AssertionError(result)
 
+    def test_monmap_exists(self, mon_containers, client):
+        command = client.exec_create(mon_containers, cmd='stat /etc/ceph/monmap-ceph')
+        result = client.exec_start(command)
+        if client.exec_inspect(command)['ExitCode'] != 0:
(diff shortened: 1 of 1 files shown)


leseb:
@alfredodeza

leseb:
test this please

tests/base/test_monitors.py
```diff
@@ -30,6 +30,12 @@ def test_keyring_exists(self, mon_containers, client):
         if client.exec_inspect(command)['ExitCode'] != 0:
             raise AssertionError(result)
 
+    def test_monmap_exists(self, mon_containers, client):
+        command = client.exec_create(mon_containers, cmd='stat /etc/ceph/monmap-ceph')
```
alfredodeza: the reason I had this different for hammer/jewel is that some entrypoint scripts use `monmap-ceph` while others use `ceph.monmap`. Does this change mean that you expect to normalize all monmap files to follow this convention of `monmap-ceph` ?

leseb: Correct. Now I see where the `ceph.monmap` comes from. It's only on the demo images. I'm going to normalize that.

alfredodeza:
We don't have CI hooked up to call these tests. Merging as it looks OK to me

leseb:
Thanks!

PR #434, opened 2016-11-03
Ceph RGW Example for Kubernetes

elsonrodriguez:
This provides a common interface (s3) for object storage with Ceph.

Will open a separate PR for documentation.


hunter:
Nice work @elsonrodriguez!

We've got a bunch of these [examples](https://github.com/AcalephStorage/ceph-docker/tree/more-kube-examples/examples/kubernetes/host) lying around but haven't had the time to clean up and raise a PR. I'm starting to think a Helm chart may be a better approach.

leseb:
Thanks again @elsonrodriguez!

elsonrodriguez:
@hunter In the same boat, I was in the middle of rewriting some of the docs but ran out of time, so I just threw in what I could.

A helm chart would definitely be a good thing, I did one for Selenium a while back and it was a slick workflow, it's probably even better now.

Issue #389, opened 2016-10-01
kubernetes mon fails to start


slang800:
I'm trying to deploy ceph to kubernetes, on CoreOS. The config files I'm using are up here: https://github.com/slang800/coreos-k8s-config/tree/master/ceph (they're copied from the kubernetes example and just rewritten to use a Makefile for secret generation).

When I deploy the monitor it crash-loops and each container has the following logs:

``` bash
$ docker logs 98644da38a92
/sbin/ip
monmaptool: invalid ip:port '<no'
```

How should I go about debugging this? The container dies and exits before I can attach to it.

hunter:
@slang800 any reason why you're trying to launch both the deployment and the daemonset for the MONs?

It seems that there is an issue with the IP or Port being passed to the monmap (part of the `entrypoint.sh`). We haven't used the examples in this repo for a while so can't remember if we ran into that specific issue. Worth having a look at passing in ENV vars to make the network easier to detect though, we have some examples here which are known to be working:

https://github.com/AcalephStorage/ceph-docker/blob/more-kube-examples/examples/kubernetes/host/ceph-mon-v1-dp.yaml#L69-L84

slang800:
Oh, no - I didn't actually know what a DaemonSet was supposed to do (still learning how to use Kubernetes). Thanks for pointing that out!

I'll try using that config example and report back if I find out what the problem is.

PR #438, opened 2016-11-14
Ensure non-empty IP address for containers passed to monmaptool

alanmeadows:
This ensures that monmaptool is only passed containers with valid IPv4/IPv6 addresses.

This partially resolves https://github.com/ceph/ceph-docker/issues/389


leseb:
@hunter @elsonrodriguez can you have a quick look at this?

hunter:
The code seems ok.

I am curious as to how to trigger the Mon without a IP address though. It's not really something that K8s should do so would like to make sure its not a symptom of another bug. @alanmeadows can you give some info about how you're launching?

PR #453, opened 2017-01-06
Updates to Kubernetes example

font:
- Add label and namespace to ceph tests
- Cleanup storage node-type label from all nodes
- Script to just create kube ceph secrets
- Default to use IPv4 even if IPv6 address exists

### DIFF
--- a/examples/kubernetes/ceph-mon-v1-ds.yaml
+++ b/examples/kubernetes/ceph-mon-v1-ds.yaml
@@ -50,7 +50,7 @@ spec:
             - name: KV_TYPE
               value: k8s
             - name: NETWORK_AUTO_DETECT
-              value: "1"
+              value: "4"
             - name: CLUSTER
               value: ceph
           volumeMounts:
--- a/examples/kubernetes/create_secrets.sh
+++ b/examples/kubernetes/create_secrets.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -ex
+
+cd generator
+./generate_secrets.sh all `./generate_secrets.sh fsid`
+
+kubectl create namespace ceph
+
+kubectl create secret generic ceph-conf-combined --from-file=ceph.conf --from-file=ceph.client.admin.keyring --from-file=ceph.mon.keyring --namespace=ceph
+kubectl create secret generic ceph-bootstrap-rgw-keyring --from-file=ceph.keyring=ceph.rgw.keyring --namespace=ceph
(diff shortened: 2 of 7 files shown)


examples/kubernetes/create_secrets.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
leseb: Don't we need to include this script execution into `create_ceph_cluster.sh`?

examples/kubernetes/ceph-mon-v1-ds.yaml
```diff
@@ -50,7 +50,7 @@ spec:
             - name: KV_TYPE
               value: k8s
             - name: NETWORK_AUTO_DETECT
-              value: "1"
+              value: "4"
```
leseb: Just for my info, what does `value` determine?

examples/kubernetes/delete_ceph_cluster.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 
 kubectl delete namespace ceph
+kubectl label nodes --all node-type-
```
leseb: anything wrong with that command or is it just me? :)

examples/kubernetes/create_secrets.sh
font: Sure, we could do that. The intent was just to have a way of creating secrets without creating everything else along with it.

examples/kubernetes/ceph-mon-v1-ds.yaml
font: This determines what address we use for the monitor and for the ceph public network. See https://github.com/ceph/ceph-docker/blob/c8c0bdf92a8bb866aef25ea691fd66502aead1a3/ceph-releases/jewel/ubuntu/14.04/daemon/entrypoint.sh#L221-L246.

examples/kubernetes/delete_ceph_cluster.sh
font: Haha yes it looks funny! This tells kubernetes to remove the `node-type` label from all nodes. The remove is indicated by the trailing `-`.

examples/kubernetes/ceph-mon-v1-ds.yaml
leseb: Oh yes right, it makes sense now.

examples/kubernetes/delete_ceph_cluster.sh
leseb: Alright thanks for the explanation :)

PR #456, opened 2017-01-11
toxification of functional tests

alfredodeza:
This is an initial take on harnessing the functional testing from ceph-ansible in ceph-docker.

This will greatly simplify the setup, validation, and teardown of the environment since ceph-ansible already has a way to do all those, including the needed support for docker itself.

For this to work it will need ceph-ansible changes to allow consuming a dev image that `build_imgs.sh` will create, and changes to ceph-build to call tox

### DIFF
--- a/tests/tox.sh
+++ b/tests/tox.sh
@@ -46,7 +46,7 @@ bash $TOXINIDIR/ceph-ansible/tests/scripts/generate_ssh_config.sh $CEPH_ANSIBLE_
 
 export ANSIBLE_SSH_ARGS="-F $CEPH_ANSIBLE_SCENARIO_PATH/vagrant_ssh_config"
 
-ansible-playbook -vv -i $CEPH_ANSIBLE_SCENARIO_PATH/hosts $TOXINIDIR/ceph-ansible/site-docker.yml.sample --extra-vars="fetch_directory=$CEPH_ANSIBLE_SCENARIO_PATH/fetch"
+ansible-playbook -vv -i $CEPH_ANSIBLE_SCENARIO_PATH/hosts $TOXINIDIR/ceph-ansible/site-docker.yml.sample --extra-vars="ceph_docker_dev_image=true fetch_directory=$CEPH_ANSIBLE_SCENARIO_PATH/fetch"
 
 ansible-playbook -vv -i $CEPH_ANSIBLE_SCENARIO_PATH/hosts $TOXINIDIR/ceph-ansible/tests/functional/setup.yml
 
--- a/tests/tox.sh
+++ b/tests/tox.sh
@@ -1,0 +1,38 @@
+#!/bin/bash -ex
+
+# Proxy script from tox. This is an intermediate script so that we can setup
+# the environment properly then call ceph-ansible for testing, and finally tear
+# down, while keeping tox features of simplicity and combinatorial confgiruation.
+#
+# NOTE: Do not run this script directly as it depends on a few environment
+# variables that tox will set, like ceph-ansible's scenario path
+
+# setup
+#################################################################################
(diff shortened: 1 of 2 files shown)


tests/tox.sh
```diff
@@ -0,0 +1,58 @@
...
+#################################################################################
+cd "$CEPH_ANSIBLE_SCENARIO_PATH"
+vagrant up --no-provision --provider=$VAGRANT_PROVIDER
+bash $TOXINIDIR/tests/scripts/generate_ssh_config.sh $CEPH_ANSIBLE_SCENARIO_PATH
+
+ansible-playbook -vv -i $CEPH_ANSIBLE_SCENARIO_PATH/hosts $TOXINIDIR/site-docker.yml.sample --extra-vars="fetch_directory=$CEPH_ANSIBLE_SCENARIO_PATH/fetch"
+
+ansible-playbook -vv -i $CEPH_ANSIBLE_SCENARIO_PATH/hosts $TOXINIDIR/tests/functional/setup.yml
+
+testinfra -n 4 --sudo -v --connection=ansible --ansible-inventory=$CEPH_ANSIBLE_SCENARIO_PATH/hosts $TOXINIDIR/tests/functional/tests
```
andrewschoen: Same issue here, incorrect path to the tests directory.

tox.ini
```diff
@@ -0,0 +1,25 @@
...
+    git
+
+passenv=*
+setenv=
+  TOXINIDIR = {toxinidir}
+  ANSIBLE_ACTION_PLUGINS = {toxinidir}/ceph-ansible/plugins/actions
+  # only available for ansible >= 2.2
+  ANSIBLE_STDOUT_CALLBACK = debug
+  centos7_cluster: CEPH_ANSIBLE_SCENARIO_PATH = {toxinidir}/ceph-ansible/tests/functional/centos/7/docker-cluster
+  xenial_cluster: CEPH_ANSIBLE_SCENARIO_PATH = {toxinidir}/ceph-ansible/tests/functional/ubuntu/xenial/docker-cluster
```
andrewschoen: There is not an ``ubuntu/xenial/docker-cluster`` testing scenario in ceph-ansible. I'm guessing we'll need to use the same scenario from ``centos/7/docker-cluster`` but tell it to use a different image.

tests/tox.sh
```diff
@@ -0,0 +1,58 @@
...
+sudo "$WORKSPACE"/travis-builds/build_imgs.sh
+
+# test
+#################################################################################
+
+# TODO: get the output image from build_imgs.sh to pass onto ceph-ansible
+
+# vars
+#################################################################################
+ANSIBLE_SSH_ARGS = -F $CEPH_ANSIBLE_SCENARIO_PATH/vagrant_ssh_config
```
andrewschoen: Does ansible actually respect ``ANSIBLE_SSH_ARGS`` when it's defined here?

tox.ini
```diff
@@ -0,0 +1,25 @@
...
+  ceph_ansible2.1: CEPH_ANSIBLE_BRANCH = stable-2.1
+
+deps=
+  ansible==2.2
```
andrewschoen: Don't you need to install the test ``requirements.txt`` from ceph-ansible here? If not we won't get testinfra installed or pytest.  Or I suppose you could just define them here directly.

tox.ini
alfredodeza: yes, correct. I didn't think it would be an issue to just have it defined already

tox.ini
alfredodeza: I can't because ceph-ansible is not checked out yet at this point, so that would mean it would need to assume the right dependencies of ceph-ansible, with the possibility of them getting out of sync. That is why the dependencies are installed in tox.sh

andrewschoen:
There is still a bit of work here to have ceph-ansible deploy the images we build as part of the process, but I can take on that work and submit a new PR for it.

PR #463, opened 2017-01-24
Update egrep of zap_device function

galindro:
This PR changes the egrep's regexp of zap_device function in entrypoint.sh to make possible to find devices that begins with **_xv_**, like **_/dev/xvdi_**.


leseb:
Thanks @galindro

galindro:
@leseb when this change will be available in ceph/daemon docker hub's image?

leseb:
Yes soon, I need to push a new version. Doing this now.

PR #478, opened 2017-02-07
osd: do not start daemon during activation

leseb:
ceph-disk got a new activation option to not run the OSD process. So we
force this and let our entrypoint run the OSD.

Fixes: #479


font:
Looks like we were tackling the same problem today but probably for different reasons. :) See https://github.com/ceph/ceph-docker/pull/480.

font:
Also, this doesn't include the changes for Jewel in Ubuntu 14.04 and all the OSes that symbolically link to it.

leseb:

Issue #487, opened 2017-02-14
Missing "rgw dns name" in ceph/demo


mgibula:
Hi,
default config created by ceph/demo does not include "rgw dns name", which causes radosgw to return signature error when using S3 interface. AFAICT there is no easy way to inject this variable into generated config (other than mounting external volume, which is not my use case).

Any chance of having this set from env variable? Would you accept pull request for this?

leseb:
Absolutely :)

PR #488, opened 2017-02-15
Set "rgw dns name" based on RGW_NAME variable

mgibula:
Hi, 
this is proposed resolution of #487 . Lighly tested on my test environment.

### DIFF
--- a/ceph-releases/kraken/ubuntu/16.04/demo/entrypoint.sh
+++ b/ceph-releases/kraken/ubuntu/16.04/demo/entrypoint.sh
@@ -215,6 +215,12 @@ function bootstrap_rgw {
     mkdir -p /var/lib/ceph/radosgw/${RGW_NAME}
     ceph ${CEPH_OPTS} auth get-or-create client.radosgw.gateway osd 'allow rwx' mon 'allow rw' -o /var/lib/ceph/radosgw/${RGW_NAME}/keyring
     chown -R ceph. /var/lib/ceph/radosgw/${RGW_NAME}
+
+    #configure rgw dns name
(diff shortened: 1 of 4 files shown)


ceph-releases/kraken/ubuntu/16.04/demo/entrypoint.sh
```diff
@@ -215,6 +215,12 @@ function bootstrap_rgw {
     mkdir -p /var/lib/ceph/radosgw/${RGW_NAME}
     ceph ${CEPH_OPTS} auth get-or-create client.radosgw.gateway osd 'allow rwx' mon 'allow rw' -o /var/lib/ceph/radosgw/${RGW_NAME}/keyring
     chown -R ceph. /var/lib/ceph/radosgw/${RGW_NAME}
+
+    #configure rgw dns name
+    cat <<ENDHERE >>/etc/ceph/${CLUSTER}.conf
```
leseb: Just in case we restart the container, please check if section exists by doing:

`if [[ ! "$(egrep "\[client.radosgw.gateway\]" /etc/ceph/${CLUSTER}.conf)" ]]; then`

mgibula: In that case, do you want to move it out of 
```if [ ! -e /var/lib/ceph/radosgw/${RGW_NAME}/keyring ]```   condition? 

I though that was enough in case of container restart, but can change that.

leseb: You're right this is enough, forget about what I said :)

PR #490, opened 2017-02-15
Copy to bin directory, otherwise if bin isn't a directory it creates …

qrpike:
Copy to bin directory, otherwise if bin isn't a directory it creates bin as a file with the last copied contents. On my vanilla CoreOS systems this was not a directory by default.

### DIFF
--- a/examples/kubernetes-coreos/startup.sh
+++ b/examples/kubernetes-coreos/startup.sh
@@ -7,11 +7,13 @@ checksum()
 	md5sum $1 | awk '{print $1}'
 }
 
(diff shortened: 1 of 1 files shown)


bzub:
Thanks I was thinking of submitting a PR to fix this.

`/etc/udev/rules.d` doesn't exist by default on CoreOS, similarly.

The only other change I wanted to make was replace the container images used to ceph/daemon which has the ceph client commands and gets updates so it would match the ceph version uses in the rest of ceph-docker.

PR #491, opened 2017-02-15
daemon: fix osd_ceph_disk scenario

dmonakhov:
Commit 0cf40ef6bb59 reorganized script scructure, but break osd_ceph_disk


leseb:
@dmonakhov Thanks

Issue #485, opened 2017-02-14
Question: Mon persistence 


qrpike:
I have ceph up and running in kubernetes ( THANK YOU ).

I see the warning says: "This example does not survive Kubernetes cluster restart! The Monitors need persistent storage. This is not covered here."

I have set the osd's using durable storage on each's host machine ( daemonset ). Do I also need to set the mon's (also daemonset) to use host machine mount point? If so what directory(s).

Are those the only 2 services that require data to persist restarts?

Thanks again for your hard work on this project. Cheers!

leseb:
Yes this is the only 2 services that need persistent storage to survive restart. Have you looked at our k8s examples?

galexrt:
@leseb But there is an issue with the mon persistence when all mon pods are deleted at the same time.
The data is gone as the data is saved in a directory named `$CLUSTER_NAME-$MON_NAME` and the `MON_NAME` (/hostname) changes on every pod "recreation" when the hostname changes and you **can't** use the same `MON_NAME` for all the mons in the cluster.
I personally think that this is related to this questions as it is an issue/question for/about the persistence.
Is there a workaround for the Ceph mon data "loss" on recreate issue?

qrpike:
@leseb Yes, I used the kubernetes example to get my cluster up and running. However,

The Mon dp/ds does not have a volume mount:
- https://github.com/ceph/ceph-docker/blob/master/examples/kubernetes/ceph-mon-v1-ds.yaml

The Osd does have a volume mount, which I changed to be a host mount:
- https://github.com/ceph/ceph-docker/blob/master/examples/kubernetes/ceph-osd-v1-ds.yaml#L39

So if the Mon dp/ds needs to write to a directory, it's not clear where to mount the host directory.

leseb:
@galexrt If you lose everything at once then you have a big problem. I don't think we have any workaround if you loose all the monitors. What we could do is to connect k8s to an external storage entity, maybe another Ceph cluster that can provide PVC to the containers.

galexrt:
@leseb The fix for that maybe to use a fixed data directory and not a data directory named after `$CLUSTER_NAME-$HOSTNAME`, as seen in my submitted PR #492

qrpike:
@leseb I put the monitors on specified hosts, so I just need to know the root directory to mount and it should be there on any reboots.

qrpike:
@galexrt Your PR makes a lot of sense and would fix this issue.

leseb:
[...]

PR #492, opened 2017-02-15
Add MON_DATA_DIR var to allow data to be saved in a set directory

galexrt:
This fixes issues when all Ceph Mon Pods are deleted and recreated. As they don't come up with the old data and are missing their original data.

Closes https://github.com/ceph/ceph-docker/issues/485


galexrt:
@leseb It seems that you are right about the CI. ;)

Issue #451, opened 2016-12-21
ceph w/ Kubernetes: Health check noises


bluecmd:
Hi,

I've successfully managed to turn up my Ceph cluster using examples/kubernetes as a base. Thanks a lot for making these examples available!

I've come to the part where I want to understand what is going on in my logs. As far as I can tell, everything is fine - except this one line that keeps coming up in both ceph-mon and ceph-mds. As it occurs every 10 seconds I'm sure that it's related to the liveliness probe (I haven't actually tried removing it).

ceph-mon: 
```
2016-12-21 02:28:30.679910 7f4e89191700  0 -- 10.188.21.39:6789/0 >> :/0 pipe(0x555e2e331400 sd=24 :6789 s=0 pgs=0 cs=0 l=0 c=0x555e2d91dc00).accept failed to getpeername (107) Transport endpoint is not connected
2016-12-21 02:28:30.679939 7f4e89292700  0 -- 10.188.21.39:6789/0 >> :/0 pipe(0x555e2e332800 sd=30 :6789 s=0 pgs=0 cs=0 l=0 c=0x555e2d91c700).accept failed to getpeername (107) Transport endpoint is not connected
2016-12-21 02:28:31.478650 7f4e8a4b5700  0 log_channel(cluster) log [INF] : pgmap v9974: 512 pgs: 512 active+clean; 70906 MB data, 139 GB used, 23581 GB / 23752 GB avail; 43041 B/s rd, 4035 B/s wr, 1 op/s
2016-12-21 02:28:33.508439 7f4e8a4b5700  0 log_channel(cluster) log [INF] : pgmap v9975: 512 pgs: 512 active+clean; 70906 MB data, 139 GB used, 23581 GB / 23752 GB avail
2016-12-21 02:28:34.930393 7f4e88827700  0 mon.ceph-mon-0@0(leader).data_health(422) update_stats avail 43% total 232 GB, used 131 GB, avail 101 GB
2016-12-21 02:28:35.538042 7f4e8a4b5700  0 log_channel(cluster) log [INF] : pgmap v9976: 512 pgs: 512 active+clean; 70906 MB data, 139 GB used, 23581 GB / 23752 GB avail; 32089 B/s rd, 453 kB/s wr, 1 op/s
[...]

PR #517, opened 2017-02-23
helm: add initial support for k8s helm

leseb:

### DIFF
--- a/examples/helm/ceph/templates/mon/statefulset.yaml
+++ b/examples/helm/ceph/templates/mon/statefulset.yaml
@@ -1,0 +1,123 @@
+---
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  labels:
+    version: v0.1.0
+    app: ceph
+    daemon: mon
+    release: {{ .Release.Name }}
+  name: ceph-mon
+spec:
+  serviceName: {{ .Values.service.mon.name | quote }}
+  replicas: {{ .Values.resources.mon.replicas }}
+  template:
+    metadata:
+      name: ceph-mon
+      labels:
+        version: v0.1.0
+        app: ceph
+        daemon: mon
+        release: {{ .Release.Name }}
+      annotations:
+        scheduler.alpha.kubernetes.io/affinity: >
+          {
+            "podAntiAffinity": {
+              "RequiredDuringSchedulingIgnoredDuringExecution": [{
+                "labelSelector": {
+                  "matchExpressions": [{
+                    "key": "daemon",
+                    "operator": "In",
+                    "values":["mon"]
+                  }]
+                },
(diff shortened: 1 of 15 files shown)


examples/helm/ceph/templates/mon/statefulset.yaml
```diff
@@ -0,0 +1,119 @@
...
+                    - /remove-mon.sh
+          ports:
+            - containerPort: {{ .Values.service.mon.port }}
+          env:
+            - name: CEPH_DAEMON
+              value: MON
+            - name: KV_TYPE
+              value: k8s
+            - name: NETWORK_AUTO_DETECT
+              value: "1"
```
hunter: Would it be easier to let Kubernetes manage the Network detection? MON_IP can be set from the podIP - https://github.com/AcalephStorage/ceph-docker/blob/more-kube-examples/examples/kubernetes/host/ceph-mon-v1-dp.yaml#L76-L79

examples/helm/ceph/templates/mon/statefulset.yaml
```diff
@@ -0,0 +1,119 @@
...
+              readOnly: true
+            - name: ceph-bootstrap-mds-keyring
+              mountPath: /var/lib/ceph/bootstrap-mds/ceph.keyring
+              subPath: ceph.keyring
+              readOnly: true
+            - name: ceph-bootstrap-rgw-keyring
+              mountPath: /var/lib/ceph/bootstrap-rgw/ceph.keyring
+              subPath: ceph.keyring
+              readOnly: true
+          livenessProbe:
```
hunter: We've noticed a high volume of logs from the Probes (mentioned in #451). Although they work, it might be good to find an alternative (or somehow silence them?)

examples/helm/ceph/templates/mon/statefulset.yaml
intlabs: @hunter This is an excellent suggestion. @leseb using the method Hunter has would be a nice improvement I think.
``` yaml
            - name: NETWORK_AUTO_DETECT
              value: "0"
            - name: MON_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.podIP
```

examples/helm/ceph/templates/mon/statefulset.yaml
intlabs: Using an exec check, as suggested in https://github.com/ceph/ceph-docker/issues/451 would potentially be a nice solution. @hunter have you looked into this at all? I'm sure we could come up with a pretty simple and quite methodology quite simply.

leseb: I suggest we merge this, we can always improve this later.

PR #522, opened 2017-03-01
zap_device: 'Payload offset' unit is sector size

ErwanAliasr1:
'Payload Offset' field from crypsetup return a number of physical sectors.

The actual code is considering the sector size to be 512 when devices can report a different value like 4096.

This patch is about reading the physical block device size by using blockdev, and use it on the 'dd' command line to specify the proper 'bs' field.
If we fail at detecting that value, then we default to 512 which is the legacy value for sectors.


leseb:

leseb:

leseb:
jenkins test this please

PR #529, opened 2017-03-03
start_mon: Reworking network detection

ErwanAliasr1:
The actual code is using a lot of regexp and a complex logic to define if we should use ipv4 or ipv6.

This patch try to simplify the read of the start_mon() function by:
- adding basic functions to extract ip or network information
- embedded the logic if 'ip' is missing directly in those functions
- improve the ipv6 support
- simplify the logic by making common calls between ipv4|6

First, the get_ip & get_network functions are created to externalize
the detection code from the main function. This way the logic is easier
to read.

This logic is simplified by unificating cases where NETWORK_AUTO_DETECT > 1 (aka 4 or 6).
We now have the same code base for both.

The newly created functions embedded the logic to detect the network configuration regarding if 'ip' is present or not.

For IPV6, some code is added to read ipv6_route or if_inet6 from /proc to extract the relevant information.
A flat_to_ipv6 function is added to transform ipv6 address in a short expression.

This patch embedded a fix for the NETWORK_AUTO_DETECT=6 case where the network was detecting by using an IPv4 regexp.

### DIFF
--- a/ceph-releases/kraken/ubuntu/16.04/daemon/config.static.sh
+++ b/ceph-releases/kraken/ubuntu/16.04/daemon/config.static.sh
@@ -7,6 +7,9 @@ function get_admin_key {
 }
 
 function get_mon_config {
+  # IPv4 is the default unless we specify it
+  IP_LEVEL=${1:-4}
+
   if [ ! -e /etc/ceph/${CLUSTER}.conf ]; then
     local fsid=$(uuidgen)
     cat <<ENDHERE >/etc/ceph/${CLUSTER}.conf
@@ -21,13 +24,8 @@ public network = ${CEPH_PUBLIC_NETWORK}
 cluster network = ${CEPH_CLUSTER_NETWORK}
 osd journal size = ${OSD_JOURNAL_SIZE}
 ENDHERE
-
-    if command -v ip; then
-      if [[ -n "$(ip -6 -o a | grep scope.global | awk '/eth/ { sub ("/..", "", $4); print $4 }' | head -n1)" ]]; then
-        echo "ms_bind_ipv6 = true" >> /etc/ceph/${CLUSTER}.conf
-        sed -i '/mon host/d' /etc/ceph/${CLUSTER}.conf
-        echo "mon host = ${MON_IP}" >> /etc/ceph/${CLUSTER}.conf
-      fi
+    if [ $IP_LEVEL -eq 6 ]; then
+      echo "ms bind ipv6 = true" >> /etc/ceph/${CLUSTER}.conf
     fi
   else
     # extract fsid from ceph.conf
--- a/ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
+++ b/ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
@@ -1,6 +1,74 @@
 #!/bin/bash
 set -e
 
+IPV4_REGEXP='[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}'
+IPV4_NETWORK_REGEXP="$IPV4_REGEXP/[0-9]\{1,2\}"
+
+function flat_to_ipv6 {
+  # Get a flat input like fe800000000000000042acfffe110003 and output fe80::0042:acff:fe11:0003
+  # This input usually comes from the ipv6_route or if_inet6 files from /proc
+
+  # First, split the string in set of 4 bytes with ":" as separator
+  value=$(echo "$@" | sed -e 's/.\{4\}/&:/g' -e '$s/\:$//')
+
+  # Let's remove the useless 0000 and "::"
+  value=${value//0000/:};
+  while $(echo $value | grep -q ":::"); do
+    value=${value//::/:};
+  done
+  echo $value
+}
+
+function get_ip {
+  NIC=$1
+  # IPv4 is the default unless we specify it
+  IP_VERSION=${2:-4}
+  # We should avoid reporting any IPv6 "scope local" interface that would make the ceph bind() call to fail
+  if is_available ip; then
+    ip -$IP_VERSION -o a s $NIC | grep "scope global" | awk '{ sub ("/..", "", $4); print $4 }' || true
+  else
+    case "$IP_VERSION" in
+      6)
+        # We don't want local scope, so let's remove field 4 if not 00
+        ip=$(flat_to_ipv6 $(grep $NIC /proc/net/if_inet6 | awk '$4==00 {print $1}'))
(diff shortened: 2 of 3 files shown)


ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -1,6 +1,74 @@
...
+    value=${value//::/:};
+  done
+  echo $value
+}
+
+function get_ip {
+  NIC=$1
+  # IPv4 is the default unless we specify it
+  IP_VERSION=${2:-4}
+  # We should avoid reporting any IPV6 "scope local" interface that would make the ceph bind() call to fail
```
leseb: s/IPV6/IPv6/

let's stay consistent with what we wrote for IPv4 :)

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -1,6 +1,74 @@
...
+}
+
+function get_network {
+  NIC=$1
+  # IPv4 is the default unless we specify it
+  IP_VERSION=${2:-4}
+
+  case "$IP_VERSION" in
+    6)
+      if command -v ip &>/dev/null; then
```
leseb: since we test the existence of that command, shouldn't we have a function for this?

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -1,6 +1,74 @@
...
+        ip -$IP_VERSION route show dev $NIC | grep proto | awk '{ print $1 }' | grep -v default | grep -vi ^fe80 || true
```
leseb: why `|| true`?

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -1,6 +1,74 @@
...
+        line=$(grep $NIC /proc/1/task/1/net/ipv6_route | awk '$2==40' | grep -v ^fe80 || true)
+        base=$(echo $line | awk '{ print $1 }')
+        base=$(flat_to_ipv6 $base)
+        mask=$(echo $line | awk '{ print $2 }')
+        echo "$base/$((16#$mask))"
+      fi
+      ;;
+    *)
+      if command -v ip &>/dev/null; then
+        ip -$IP_VERSION route show dev $NIC | grep proto | awk '{ print $1 }' | grep -v default | grep "/" || true
```
leseb: ditto

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -1,6 +1,74 @@
...
+  # IPv4 is the default unless we specify it
+  IP_VERSION=${2:-4}
+
+  case "$IP_VERSION" in
+    6)
+      if command -v ip &>/dev/null; then
+        ip -$IP_VERSION route show dev $NIC | grep proto | awk '{ print $1 }' | grep -v default | grep -vi ^fe80 || true
+      else
+        # We don't want the link local routes
+        line=$(grep $NIC /proc/1/task/1/net/ipv6_route | awk '$2==40' | grep -v ^fe80 || true)
```
leseb: ditto

ceph-releases/kraken/ubuntu/16.04/daemon/config.static.sh
```diff
@@ -21,14 +24,9 @@ public network = ${CEPH_PUBLIC_NETWORK}
...
 ENDHERE
-
-    if command -v ip; then
-      if [[ -n "$(ip -6 -o a | grep scope.global | awk '/eth/ { sub ("/..", "", $4); print $4 }' | head -n1)" ]]; then
-        echo "ms_bind_ipv6 = true" >> /etc/ceph/${CLUSTER}.conf
-        sed -i '/mon host/d' /etc/ceph/${CLUSTER}.conf
-        echo "mon host = ${MON_IP}" >> /etc/ceph/${CLUSTER}.conf
-      fi
-    fi
+	if [ $IP_LEVEL -eq 6 ]; then
```
ErwanAliasr1: fixed

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
ErwanAliasr1: fixed

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -1,6 +1,74 @@
...
+  echo $value
+}
+
+function get_ip {
+  NIC=$1
+  # IPv4 is the default unless we specify it
+  IP_VERSION=${2:-4}
+  # We should avoid reporting any IPV6 "scope local" interface that would make the ceph bind() call to fail
+  if command -v ip &>/dev/null; then
+    ip -$IP_VERSION -o a s $NIC | grep "scope global" | awk '{ sub ("/..", "", $4); print $4 }' || true
```
ErwanAliasr1: we are running under bash -e meaning a faulty commands makes the whole aborting.
As we are trying to detect something, I don't want to fail here because of a mis-dectection.
That's partciulary true with IPv6 since we try to detect something that is very unlikely to be present.

We have some code checking lately if MON_IP is empty and fail at that point with a clear message.

So I'm putting here a || true to let that code triggering the real exception.

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
ErwanAliasr1: same as previous comment.

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
ErwanAliasr1: ditto ;)

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
ErwanAliasr1: ditto ;)

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
leseb: Ack  thanks.

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
ErwanAliasr1: Good idea. done with is_available()

PR #537, opened 2017-03-09
Creating a loop & move mkdir

ErwanAliasr1:
Less code, less bugs


ktdreyer:
The commit logs don't mention dropping the chown to `ceph`. Why drop it?

ErwanAliasr1:
@ktdreyer because the last line of create_mandatory_directories() does it. That's implicit. Should I make it explicit in the commit message ?

ktdreyer:
Whoops, I should've noticed that. Thanks for pointing it out to me!

PR #539, opened 2017-03-09
ensure packages are installed after "yum install"

ktdreyer:
When "yum install <list-of-packages>" fails to install package(s) in the list, it will simply print a warning, eg, "No package ceph-radosgw available". It still exits with a "0" exit code in this case, leading Docker to continue on with the build process as if nothing is wrong.

This can happen if one of the packages is accidentally unavailable in our Yum repositories.

If we fail to install a package in this list, we need the build process to fail immediately so we know the container will not be usable.

Rather than only relying on "yum install"'s exit code, use rpm to double-check that the packages were indeed installed locally.

### DIFF
--- a/ceph-releases/hammer/centos/7/base/Dockerfile
+++ b/ceph-releases/hammer/centos/7/base/Dockerfile
@@ -20,7 +20,8 @@ RUN yum install -y runit-2.1.1-7.el7.centos.x86_64
 RUN rpm --import 'https://download.ceph.com/keys/release.asc'
 RUN rpm -Uvh https://download.ceph.com/rpm-${CEPH_VERSION}/el7/noarch/ceph-release-1-1.el7.noarch.rpm
 RUN yum install -y epel-release && yum clean all
-RUN yum install -y ceph ceph-radosgw sharutils && yum clean all
+ARG PACKAGES="ceph ceph-radosgw sharutils"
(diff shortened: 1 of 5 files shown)


ErwanAliasr1:
Can't we have variables to avoid duplicating the package list ? That's really make my eyes bleeding.

+1 for the double check but it requires that the list is the real packages names and not a 'provide' name because the package name could be different.

ktdreyer:
Thanks for the review. I've pushed a new version with your suggestion of using a variable for the package list.

ErwanAliasr1:
jenkins test this please

leseb:

leseb:

leseb:
@ktdreyer your patch is producing an error on the CI, looks like the build is failing.

ktdreyer:
`package nfs-ganesha-rgw is not installed` ... looks like this uncovered a real problem on CentOS. Currently we don't point at any Yum repo that has this package.

ErwanAliasr1:
@ktdreyer Noticed that too and added a build hint on that kind of issue.

ktdreyer:
I've pushed a second commit that adds the nfs-ganesha CI repository until we can come up with something more stable.

leseb:
jenkins test this please

ErwanAliasr1:
jenkins test this please

PR #548, opened 2017-03-16
DNM: osd_disk_prepare: printing parted failure

ErwanAliasr1:
When the parted print fails it means something was really wrong but as
parted's return code isn't that useful, let's print the std{out|err} to
ease the debug.

PR #552, opened 2017-03-16
daemon: explicitly install all the ceph packages

leseb:
Prior to this commit we were relying on the "ceph" package pulling all
the ceph-mon, ceph-osd, ceph-mgr packages. So now we call each packages
individually.


ktdreyer:
:+1: once this passes Jenkins

leseb:
Jenkins failure is "expected" the CI is broken at the moment and we are investigating this as part of https://github.com/ceph/ceph-docker/pull/548

ktdreyer:
Thanks, merged in that case

PR #555, opened 2017-03-17
Added MON variable for Kubernetes hostNetwork mode

galexrt:
This will use the nodeNames instead of the pod names for the monmap

***

This allows users to use `hostNetwork: true` mode for Ceph mons in Kubernetes.

### DIFF
--- a/ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
+++ b/ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
@@ -88,7 +88,7 @@ function start_mon {
       CEPH_PUBLIC_NETWORK=$(get_network ${NIC_MORE_TRAFFIC} ${NETWORK_AUTO_DETECT})
       IP_VERSION=${NETWORK_AUTO_DETECT}
     else # Means -eq 1
-      MON_IP=$(get_ip ${NIC_MORE_TRAFFIC} 6)
+      MON_IP="[$(get_ip ${NIC_MORE_TRAFFIC} 6)]"
       CEPH_PUBLIC_NETWORK=$(get_network ${NIC_MORE_TRAFFIC} 6)
       IP_VERSION=6
       if [ -z "$MON_IP" ]; then
@@ -129,7 +129,7 @@ function start_mon {
     ceph-mon --setuser ceph --setgroup ceph -i ${MON_NAME} --inject-monmap $MONMAP --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
     # Ignore when we timeout in most cases that means the cluster has no qorum or
     # no mons are up and running
-    timeout 7 ceph mon add ${MON_NAME} "${MON_IP}:6789" || true
+    timeout 7 ceph mon add "${MON_NAME}" "${MON_IP}:6789" || true
   fi
 
   log "SUCCESS"
--- a/ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
+++ b/ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
@@ -129,7 +129,7 @@ function start_mon {
     ceph-mon --setuser ceph --setgroup ceph -i ${MON_NAME} --inject-monmap $MONMAP --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
     # Ignore when we timeout in most cases that means the cluster has no qorum or
     # no mons are up and running
-    timeout 7 ceph mon add "${MON_NAME}" "${MON_IP}:6789" || true
+    timeout 7 ceph ${CLI_OPTS} mon add "${MON_NAME}" "${MON_IP}:6789" || true
   fi
 
   log "SUCCESS"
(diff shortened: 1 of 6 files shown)


hunter:
We were just about to file a similar PR. We'd started with a variable similar to `K8S_HOST_NETWORK` but [opted](https://github.com/ceph/ceph-docker/compare/master...AcalephStorage:monmapadd-name) instead to avoid an extra option and use `MON_NAME`. That way a fieldRef override to `HOSTNAME` could be used.

e.g.

```
- name: HOSTNAME
     valueFrom:
       fieldRef:
         fieldPath: spec.nodeName
```

galexrt:
@hunter The `HOSTNAME` will be already set to the node's hostname when running in `hostNetwork: true`. But the problem this PR solves is the monmap generation that still uses the pod name and not the nodeName when in `hostNetwork` mode.

hunter:
Yep, understood. I used a confusing example (updated above) but it operates in a similar way. We override the `HOSTNAME` (and cascading to `MON_NAME`) in the entrypoint to make that things are consistent

galexrt:
@hunter The problem is not entirely with the `MON_NAME`.
If you look at this code https://github.com/galexrt/ceph-docker/blob/f099b4c332e5982b74b5355249843c99f8cad279/ceph-releases/kraken/ubuntu/16.04/daemon/config.k8s.sh#L19 it will use the pod name for the mon name in the **monmap** and not the nodeName like I added https://github.com/galexrt/ceph-docker/blob/f099b4c332e5982b74b5355249843c99f8cad279/ceph-releases/kraken/ubuntu/16.04/daemon/config.k8s.sh#L21 the monmap needs to use all nodeNames when running in `hostNetwork` mode.

hunter:
Yes, I understand what you mean. Not sure if you saw the link to the code that was part of my first comment - https://github.com/ceph/ceph-docker/compare/master...AcalephStorage:monmapadd-name#diff-816f945fdcfa475983d3dbe3e91341bdR19

We're trying to solve the same problem. We just opted to move the logic to the K8s spec. In `hostNetwork` mode it is important to add the hostname to the monmap. We'd also like the flexibility to use a consistent StatefulSet Pod name in place of that hostname.

galexrt:
@hunter The problem I see from you change is, that the `MON_NAME` will now be used for all monitors in the cluster as the kubernetes "template" used here is a loop over all found pods.

hunter:
You make a good point. We'd missed the issue since we'd used a StatefulSet. We'd been trying to avoid an extra variable but no other way unless we try the new DNS discovery with Kraken :)

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -129,7 +129,7 @@ function start_mon {
     ceph-mon --setuser ceph --setgroup ceph -i ${MON_NAME} --inject-monmap $MONMAP --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
     # Ignore when we timeout in most cases that means the cluster has no qorum or
     # no mons are up and running
-    timeout 7 ceph mon add ${MON_NAME} "${MON_IP}:6789" || true
+    timeout 7 ceph mon add "${MON_NAME}" "${MON_IP}:6789" || true
```
leseb: Just spotted this, the command needs `$CLI_OPTS` after `ceph` so: `ceph $CLI_OPTS`.
Do you mind fixing this in this PR?

galexrt:
@leseb Added the `CLI_OPTS`.

ceph-releases/kraken/ubuntu/16.04/daemon/start_mon.sh
```diff
@@ -129,7 +129,7 @@ function start_mon {
     ceph-mon --setuser ceph --setgroup ceph -i ${MON_NAME} --inject-monmap $MONMAP --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
     # Ignore when we timeout in most cases that means the cluster has no qorum or
     # no mons are up and running
-    timeout 7 ceph mon add ${MON_NAME} "${MON_IP}:6789" || true
+    timeout 7 ${CLI_OPTS} ceph mon add "${MON_NAME}" "${MON_IP}:6789" || true
```
leseb: The other way around: `ceph ${CLI_OPTS}`

galexrt: My bad, fixed it.

leseb:
@galexrt @hunter did you get to an agreement on the k8s part?

hunter:
Yes, it was my mistake. @galexrt's implementation works :)

galexrt:
One thing to note is that when using `hostNetwork` you:
* **a)** specify the mon hosts in the `mon_host` setting directly (as the cluster internal dns isn't used in `hostNetwork` mode, see **b)** for more info)
* **b)** wait for Kubernetes 1.6.x to implement the `dnsPolicy: ClusterFirstWithHostNet` (see https://github.com/kubernetes/kubernetes/pull/29378)
* **c)** create a dns name with the mons addresses in it (round robin style ~= `clusterIP: none` style) and wait until **b)** has been implemented and releasesd (this is currently the way I do it, until  Kubernetes 1.6.x is released

hunter:
Brilliant! I didn't know that `dnsPolicy: ClusterFirstWithHostNet` is being implemented. Will definitely be useful.

We're still using `mon_host` with StatefulSets and *c* with a minor change to use the Pod name/HostIP instead of Hostname.

PR #561, opened 2017-03-21
Helm: Prometheus exporter

hunter:
Adds basic support for the digitalocean ceph-exporter to allow metrics
collection

### DIFF
--- a/examples/helm/ceph/templates/exporter/deployment.yaml
+++ b/examples/helm/ceph/templates/exporter/deployment.yaml
@@ -1,0 +1,55 @@
+{{- if .Values.ceph.enabled.exporter }}
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: ceph-exporter
+  labels:
+    version: {{ .Chart.Version }}
+    app: ceph
+    daemon: exporter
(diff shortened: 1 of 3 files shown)


examples/helm/ceph/templates/exporter/deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+      containers:
+      - name: ceph-exporter
+        image: {{ .Values.images.exporter }}
+        imagePullPolicy: {{ .Values.image_policy.pull }}
+        ports:
+          - name: ceph-exporter
+            containerPort: {{ .Values.service.exporter.port }}
+        volumeMounts:
+          - name: ceph-conf
+            mountPath: /etc/ceph/ceph.conf
```
leseb: Do we have a way to expose a different cluster name?

hunter: Possible but may need some rethinking across all of the resources in the helm chart. Something that can be done in another PR?

leseb: Absolutely.

leseb:
I think something went wrong during your rebase, can you rebase again against master?

PR #562, opened 2017-03-21
Helm: Configurable OSD directory or devices

hunter:
I thought I'd raise this PR to get some discussion going on the best way to handle multiple OSDs in a cluster. There has been some discussion in [various](https://github.com/att-comdev/openstack-helm/pull/158) places about real disk support.

We've been using an approach with daemonsets per device across a cluster. It gives more control to the OSD lifecycle with drive types, crush location and rolling restarts. Its served us reasonable well for clusters with identical servers/disks but may be difficult to manage in a varied cluster.

It may be that the best solution is a ThirdPartyResources/Operator but for now this does allow the use of real devices with Helm. Please jump in with thoughts, alternatives and ideas

### DIFF
--- a/examples/helm/ceph/values.yaml
+++ b/examples/helm/ceph/values.yaml
@@ -125,3 +125,31 @@ storageclass:
   admin_secret_name: pvc-ceph-conf-combined-storageclass
   user_id: admin
   user_secret_name: pvc-ceph-client-key
+
+## osd_devices defines the osd layout for each node. Each
+## item will create a DaemonSet for the given configuration
+## but will not activate the OSDs yet. To activate the OSD,
+## the node should be labeled with `device_{device}={type}`,
+## eg. `device_sdb=hdd`. This will activate `/dev/sdb` to
+## become an OSD daemon.
+##
+## Devices need to have a GPT partition or activation may fail
+##
+## the format for an OSD item is:
+##   device: {the block storage. eg. `/dev/sdb` should just be `sdb`}
+##   type: {osd classification. typically `hdd` and `ssd` but can be anything}
+##   crush_location: {optional. updates the crush location of the osd}
+##   journal: {optional. if specified, will not prepare the disk and will not collocate the journal}
+##
+osd_directory:
+  enabled: true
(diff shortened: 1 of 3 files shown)


leseb:
Do you mind rebasing this on top of master please?
I'm looking into it.

examples/helm/ceph/values.yaml
```diff
@@ -125,3 +125,31 @@ storageclass:
...
+## the format for an OSD item is:
+##   device: {the block storage. eg. `/dev/sdb` should just be `sdb`}
+##   type: {osd classification. typically `hdd` and `ssd` but can be anything}
+##   crush_location: {optional. updates the crush location of the osd}
+##   journal: {optional. if specified, will not prepare the disk and will not collocate the journal}
+##
+osd_directory:
+  enabled: true
+osd_devices:
+  - device: sdb
```
leseb: This is an issue for people using persistent path such as `/dev/disk/by-`.
Can we ask for the full path instead?

hunter: Yes, that is the downside with this approach and one of the reasons why I'd opened the PR for discussion.

We use the `device` for building the name and selectors in the daemonset (e.g. `name: ceph-osd-{{ $value.device }}-{{ $value.type }}`) so won't work for the longer `by-*` paths. I haven't really found with a decent alternative that will work for all cases. Any thoughts?

leseb: Alright, so I've been thinking and put my thoughts into this PR, would you mind having a look at the README and tell me what you think? I left some questions too. Thanks!
https://github.com/ceph/ceph-docker/pull/610

The PR is obviously incomplete at the moment but the general idea is to keep one OSD per container and not to have to declare an OSD device for it. In an ideal world someone would ask k8s, deploy me a Ceph cluster from a specific set of labeled machines (storage node), take all the disks and use them to build my Ceph cluster.

leseb:
I'm closing this since this should probably go in https://github.com/ceph/ceph-helm/.
Thanks

Issue #558, opened 2017-03-21
Using PPA ?


ErwanAliasr1:
On build https://jenkins.ceph.com/job/ceph-docker-prs-ceph_ansible-kraken-xenial_cluster/179/console, we fail because of

The following packages have unmet dependencies:
 nfs-ganesha-fsal : Depends: glusterfs-common (>= 3.8.8) but it is not going to be installed: Unable to correct problems, you have held broken packages

The packages are coming from a PPA. Is it safe to build our containers against a PPA which is pretty unsupported ? Can't we find a more reliable stream of package ?

ktdreyer:
At this point I think we should stop using the PPA and use Shaman (I've done this for CentOS 7 in https://github.com/ceph/ceph-docker/pull/539). The RGW FSAL is under heavy development today, and there are some bugfixes that introduce strict dependencies on the very latest `LIBRGW_FILE_VER` from Ceph. Maybe in six months this will be more stable and flexible.

andrewschoen:
@ktdreyer I think using the nfs-ganesha repo from shaman is fine, but if we plan to use it like this we'll need to add some config to chacra to ensure that we always have a nfs-ganesha repo available in shaman. I think it's built often enough we wont have to worry, but better safe than sorry.

We do something similar for a few other projects: https://github.com/ceph/chacra/blob/master/deploy/playbooks/roles/common/templates/prod.py.j2#L126

PR #564, opened 2017-03-22
daemon: resolve broken package dependancy

leseb:
Error:  nfs-ganesha-fsal : Depends: glusterfs-common (>= 3.8.8) but it
is not going to be installed

Solution: temporary remove the package and re-introduce it once the
issue is resolved.

This is urgent as this breaks all the builds on Xenial.

### DIFF
--- a/ceph-releases/kraken/ubuntu/16.04/daemon/entrypoint.sh
+++ b/ceph-releases/kraken/ubuntu/16.04/daemon/entrypoint.sh
@@ -112,6 +112,9 @@ case "$CEPH_DAEMON" in
     start_rbd_mirror
     ;;
   nfs)
+    echo "Temporarily disabled due to broken package dependencies with nfs-ganesha"
+    echo "For more info see: https://github.com/ceph/ceph-docker/pull/564"
+    exit 1
     source start_nfs.sh
     start_nfs
     ;;
--- a/ceph-releases/kraken/ubuntu/16.04/demo/Dockerfile
+++ b/ceph-releases/kraken/ubuntu/16.04/demo/Dockerfile
@@ -9,11 +9,10 @@ ENV CEPH_VERSION kraken
 # Install Ceph and prerequisites
 RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y wget unzip uuid-runtime python-setuptools udev sharutils python3 s3cmd && \
     apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FE869A9 && \
-    echo "deb http://ppa.launchpad.net/gluster/nfs-ganesha/ubuntu xenial main" | tee /etc/apt/sources.list.d/nfs-ganesha.list && \
     echo "deb http://ppa.launchpad.net/gluster/libntirpc/ubuntu xenial main" | tee /etc/apt/sources.list.d/libntirpc.list && \
     wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add - && \
     echo "deb http://download.ceph.com/debian-$CEPH_VERSION/ xenial main" | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list && \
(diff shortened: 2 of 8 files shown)


leseb:

leseb:
jenkins test this please

ktdreyer:
Without the FSAL, NFS-Ganesha does not support CephFS or RGW.

Some options here:

A) Drop nfs-ganesha-fsal from Ubuntu containers without dropping nfs-ganesha: this just bloats our containers with a non-functional daemon, I'd prefer we not do this.

B) Drop nfs-ganesha entirely from Ubuntu containers. I'd prefer this over Option A.

C) Switch from the nfs-ganesha PPA to the Shaman repo (discussed in https://github.com/ceph/ceph-docker/issues/558).

Longer-term, I think it makes sense to split the nfs-ganesha-fsal package (https://github.com/nfs-ganesha/nfs-ganesha-debian/issues/2) so we can only install the RGW FSAL in the container.

leseb:
@ktdreyer we can not have images diverging, like on distro providing more/less feature than another. So it's either Ubuntu and CentOS or nothing I'm afraid.

I'd suggest we merge this one quickly to unlock the situation so yes the nfs_rgw implementation will be broken but I can add a message for that.

What do you think?

ktdreyer:
Why keep nfs-ganesha here when you are dropping nfs-ganesha-fsal?

leseb:
@ktdreyer this PR is not completed yet, but yes I'll remove that package too.

leseb:
@ktdreyer new version, so the packages are still installed on Ubuntu but the feature is dropped for both Ubuntu and CentOS.

ceph-releases/kraken/ubuntu/16.04/daemon/entrypoint.sh
```diff
@@ -112,6 +112,9 @@ case "$CEPH_DAEMON" in
     start_rbd_mirror
     ;;
   nfs)
+    echo "Temporary disabled due to broken package dependencies with nfs-ganesha"
```
ktdreyer: s/Temporary/Temporarily/

ceph-releases/kraken/ubuntu/16.04/demo/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y wget unz
     echo "deb http://ppa.launchpad.net/gluster/libntirpc/ubuntu xenial main" | tee /etc/apt/sources.list.d/libntirpc.list && \
     wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add - && \
     echo "deb http://download.ceph.com/debian-$CEPH_VERSION/ xenial main" | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list && \
-    apt-get update && apt-get install -y --force-yes ceph-mon ceph-osd ceph-mds ceph-mgr ceph-base ceph-common radosgw rbd-mirror nfs-ganesha nfs-ganesha-fsal && \
+    apt-get update && apt-get install -y --force-yes ceph-mon ceph-osd ceph-mds ceph-mgr ceph-base ceph-common radosgw rbd-mirror nfs-ganesha && \
```
ktdreyer: This `Dockerfile` needs further adjustment as well, right? (It would be great to just drop Trusty altogether to avoid this type of confusion)

ceph-releases/kraken/ubuntu/16.04/daemon/entrypoint.sh
leseb: Fixed.

ceph-releases/kraken/ubuntu/16.04/demo/Dockerfile
leseb: Fixed.

PR #566, opened 2017-03-23
daemon: improve debug mode

leseb:
This patch enhances the current support for debugging a container during
runtime.
This patch introduces the ability to specify more than one debug option.
We currently have 3 options when activating the DEBUG option:

  * DEBUG=verbose will enable 'set -x'
  * DEBUG=fstree along with an URL, so DEBUG=fstree=http://myfstree will
download the content of an URL (expecting an fs tree) and uncompress it
in /. This is useful if you want to test a patch and/or add new versions
of a binary.
  * DEBUG=stayalive will launch a shell when the container fails

You can combine these options using comma as separators like:

-e DEBUG=verbose,fstree=http://myfstree,stayalive


ceph-releases/kraken/ubuntu/16.04/daemon/debug.sh
```diff
@@ -0,0 +1,33 @@
...
+      set -x
+      ;;
+    fstree*)
+      echo "FSTREE: uncompressing content of $(extract_param $option)"
+      # NOTE (leseb): the entrypoint should already be running from /
+      # This is just a safeguard
+      pushd / > /dev/null
+      wget -q $(extract_param $option) -O patch.tar
+      echo "SHA1 of the archive is: $(sha1sum patch.tar)"
+      for f in $(tar xfpv patch.tar); do
```
ErwanAliasr1: Adding comment "Let's print the sha1 of each file of the archive to ease understanding if we have the right file in place"

Issue #572, opened 2017-03-27
source /etc/sysconfig/ceph (for tcmalloc settings, etc.)


ktdreyer:
The non-containerized services source `/etc/sysconfig/ceph` when running under systemd or upstart: https://github.com/ceph/ceph/blob/master/etc/sysconfig/ceph (see `EnvironmentFile=-/etc/sysconfig/ceph` in `ceph/systemd/ceph-osd@.service`, for example)

The most important env variable here is `TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES`, which improves performance significantly.

How can we make the containers source this file as well, to keep parity between the non-containerized services and containerized?

PR #584, opened 2017-04-03
daemon: source ceph env files

leseb:
As described in https://github.com/ceph/ceph-docker/issues/572, we
currently don't source `/etc/sysconfig/ceph` on RedHat based systems and
`/etc/default/ceph` on Ubuntu based systems.
This is important we source this file so we can enable
TCMALLOC_MAX_TOTAL_THREAD_CACHE_BYTES.

This commits adds some functions to determine on which OS the container
image is running. Then during the OSD execution sources the right file.

Fixes: https://github.com/ceph/ceph-docker/issues/572

### DIFF
--- a/ceph-releases/kraken/ubuntu/16.04/daemon/common_functions.sh
+++ b/ceph-releases/kraken/ubuntu/16.04/daemon/common_functions.sh
@@ -174,3 +174,24 @@ function to_lowercase {
 function comma_to_space {
   echo "${@//,/ }"
 }
+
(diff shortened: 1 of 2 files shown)


ktdreyer:
No need to detect the OS here, right? just test if `/etc/sysconfig/ceph` exists, and if so, source it?

leseb:
@ktdreyer I was actually thinking that we might need that for something else later so these OS detect function might be useful.

PR #597, opened 2017-04-12
use /etc/ceph/.conf in config.k8s.sh get_mon_config()

### LINKED ISSUE #596: Kubernetes mon config use /etc/ceph/ceph.conf file instead of /etc/ceph/${CLUSTER}.conf like other providers
Hello,
in **config.k8s.sh** the function **get_mon_config** expects the file /etc/ceph/ceph.conf to exists.

So if we use a custom CLUSTER value to deploy multiple cephs instances in different kubernetes namespaces the file  /etc/ceph/ceph.conf must exists for the mon config but also the file ${CLUSTER}.conf for osd and mds. If we use the same secret like in the example we have to duplicate the file.

It would be better to use only  ${CLUSTER}.conf to make it more simple to use a custom cluster name and use the same logic as other config providers where only ${CLUSTER}.conf is used.

osixia:
Fixes #596

### DIFF
--- a/ceph-releases/infernalis/ubuntu/14.04/daemon/config.k8s.sh
+++ b/ceph-releases/infernalis/ubuntu/14.04/daemon/config.k8s.sh
@@ -8,7 +8,7 @@ function get_admin_key {
 
 function get_mon_config {
   # Get FSID from ceph.conf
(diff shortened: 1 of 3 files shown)


leseb:
Thanks.

Merging even if the CI is failing since this patch touches a different code path.

Issue #614, opened 2017-04-25
Ownership of off-OSD bluestore block.wal and block.db devices is not correctly set during prepare.


netservers:
When preparing a bluestore OSD, where the wal and/or db pre-exists on a separate device (like an SSD), the osd_disk_prepare.sh script does not set ownership of the /dev/sdXX device to ceph:ceph.

Lines 77 to 80 of osd_disk_prepare.sh iterate over partitions 1 to 4 of the OSD device setting ownership of /dev/sdX[1-4] to ceph.

Consequently if the wal and db devices have not manually been chown'd the OSD will fail to start as follows:

>2017-04-25 17:52:48.041705 7f58e8d69a40 -1 bluestore(/var/lib/ceph/tmp/mnt.PFV7CE) _setup_block_symlink_or_file failed to open block.wal file: (13) Permission denied
2017-04-25 17:52:48.041733 7f58e8d69a40 -1 OSD::mkfs: ObjectStore::mkfs failed with error -13
2017-04-25 17:52:48.041837 7f58e8d69a40 -1  ** ERROR: error creating empty object store in /var/lib/ceph/tmp/mnt.PFV7CE: (13) Permission denied

leseb:
That's a bit tricky but I should be able to come up with something decent soon.
Thanks for reporting this.

leseb:
First quick stab in: https://github.com/ceph/ceph-docker/pull/616 please test

netservers:
Sorry for leaving this for so long. I've been tied up with other things. I've still got a few issues with the fix as currently written. I've got my own modifications variation on the patch that work for me, in the sense that all permissions are correctly set and the daemon starts when it should.

However, as I hinted at before I think there's an underlying problem with Luminous, whereby the OSDs never come "up" after I reboot the host system. I get no such problem with Kraken, so I don't think this restart problem related this this issue or patch. It just means that I can't say 100% that my fix is good.

If it's OK with you I'd like to submit a fresh pull request with the latest changes I've made and see what you think.

leseb:
Sure, if you don't mind taking over this PR: https://github.com/ceph/ceph-docker/pull/616 so I can save my earlier commit. Just re-work the parts that need to be re-written.

What do you think? Thanks!
Anyway, I'll have a fresh look into this this week.

PR #616, opened 2017-04-25
daemon: fix bluestore wal and db partition ownership

leseb:
Fixes: https://github.com/ceph/ceph-docker/issues/614 and https://github.com/ceph/ceph-docker/pull/616

### DIFF
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
@@ -74,9 +74,17 @@ function osd_disk_prepare {
     wait_for_file ${OSD_JOURNAL}
     chown ceph. ${OSD_JOURNAL}
   elif [[ ${OSD_BLUESTORE} -eq 1 ]]; then
-    for part in $(seq 1 4); do
-      wait_for_file $(dev_part ${OSD_DEVICE} $part)
-      chown ceph. $(dev_part ${OSD_DEVICE} $part)
+    dev_real_path=$(resolve_symlink $OSD_BLUESTORE_BLOCK_WAL $OSD_BLUESTORE_BLOCK_DB $OSD_DEVICE)
+    for partition in $(list_dev_partitions $dev_real_path); do
+      if [[ "$(get_part_typecode $partition)" == "5ce17fce-4087-4169-b7ff-056cc58472be" ]]; then
+        chown --verbose ceph. $partition
+      fi
+      if [[ "$(get_part_typecode $partition)" == "30cd0809-c2b2-499c-8879-2d6b785292be" ]]; then
+        chown --verbose ceph. $partition
+      fi
+      if [[ "$(get_part_typecode $partition)" == "89c57f98-2fe5-4dc0-89c1-f3ad0ceff2be" || "$(get_part_typecode $partition)" == "cafecafe-9b03-4f30-b4c6-b4b80ceff106" ]]; then
+        chown --verbose ceph. $partition
+      fi
     done
   else
     wait_for_file $(dev_part ${OSD_DEVICE} 2)
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
@@ -72,7 +72,7 @@ function osd_disk_prepare {
 
   if [[ -n "${OSD_JOURNAL}" ]]; then
     wait_for_file ${OSD_JOURNAL}
-    chown ceph. ${OSD_JOURNAL}
+    chown --verbose ceph. ${OSD_JOURNAL}
   elif [[ ${OSD_BLUESTORE} -eq 1 ]]; then
     dev_real_path=$(resolve_symlink $OSD_BLUESTORE_BLOCK_WAL $OSD_BLUESTORE_BLOCK_DB $OSD_DEVICE)
     for partition in $(list_dev_partitions $dev_real_path); do
@@ -88,6 +88,6 @@ function osd_disk_prepare {
     done
   else
     wait_for_file $(dev_part ${OSD_DEVICE} 2)
-    chown ceph. $(dev_part ${OSD_DEVICE} 2)
+    chown --verbose ceph. $(dev_part ${OSD_DEVICE} 2)
   fi
 }
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
@@ -61,7 +61,7 @@ function osd_disk_prepare {
     echo "Unmounting LOCKBOX directory"
     # NOTE(leseb): adding || true so when this bug will be fixed the entrypoint will not fail
     # Ceph bug tracker: http://tracker.ceph.com/issues/18944
-    DATA_UUID=$(blkid -o value -s PARTUUID ${OSD_DEVICE}1)
+    DATA_UUID=$(get_part_uuid ${OSD_DEVICE}1)
     umount /var/lib/ceph/osd-lockbox/${DATA_UUID} || true
   else
     ceph-disk -v prepare ${CLI_OPTS} --journal-uuid ${OSD_JOURNAL_UUID} ${OSD_DEVICE} ${OSD_JOURNAL}
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
@@ -70,24 +70,5 @@ function osd_disk_prepare {
   # watch the udev event queue, and exit if all current events are handled
   udevadm settle --timeout=600
 
-  if [[ -n "${OSD_JOURNAL}" ]]; then
-    wait_for_file ${OSD_JOURNAL}
-    chown --verbose ceph. ${OSD_JOURNAL}
-  elif [[ ${OSD_BLUESTORE} -eq 1 ]]; then
-    dev_real_path=$(resolve_symlink $OSD_BLUESTORE_BLOCK_WAL $OSD_BLUESTORE_BLOCK_DB $OSD_DEVICE)
-    for partition in $(list_dev_partitions $dev_real_path); do
-      if [[ "$(get_part_typecode $partition)" == "5ce17fce-4087-4169-b7ff-056cc58472be" ]]; then
-        chown --verbose ceph. $partition
-      fi
-      if [[ "$(get_part_typecode $partition)" == "30cd0809-c2b2-499c-8879-2d6b785292be" ]]; then
-        chown --verbose ceph. $partition
-      fi
-      if [[ "$(get_part_typecode $partition)" == "89c57f98-2fe5-4dc0-89c1-f3ad0ceff2be" || "$(get_part_typecode $partition)" == "cafecafe-9b03-4f30-b4c6-b4b80ceff106" ]]; then
-        chown --verbose ceph. $partition
-      fi
-    done
-  else
-    wait_for_file $(dev_part ${OSD_DEVICE} 2)
-    chown --verbose ceph. $(dev_part ${OSD_DEVICE} 2)
-  fi
+  apply_ceph_ownership_to_disks
 }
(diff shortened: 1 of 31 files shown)


ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_prepare.sh
```diff
@@ -74,9 +74,25 @@ function osd_disk_prepare {
     wait_for_file ${OSD_JOURNAL}
     chown ceph. ${OSD_JOURNAL}
   elif [[ ${OSD_BLUESTORE} -eq 1 ]]; then
-    for part in $(seq 1 4); do
-      wait_for_file $(dev_part ${OSD_DEVICE} $part)
-      chown ceph. $(dev_part ${OSD_DEVICE} $part)
+    # todo: remove /dev/ just keep the device
+    # todo: readlink -f
+    dev_real_path=$(resolve_symlink $OSD_BLUESTORE_BLOCK_WAL)
+    for partition in $(list_dev_partitions ${dev_real_path#/dev/}); do
```
leseb: The 3 loops can probably be solved with one, still thinking about this.

netservers:
One comment I have on this fix is that I'm not sure the dependence on a partition name of "ceph block.wal" is a good idea. I say this because I am going to place wal and db devices for multiple OSDs on a single NVMe device. For ease of management I'm going to put the OSD ID in the partition name like "osd_0_block.wal". Obviously when the wal and db are on the OSD this is would not be a consideration.

netservers:
I think it might be best the follow the example set by the udev rules in a non-dockerized environment. 
Those rules check the typecode of the partitions and then chown the dev devices accordingly. By following that example you would be doing not more harm that udev normally does if someone were to pass you in the wrong device.

So in our case, I'd suggest identifying block.db by the typecode 30cd0809-c2b2-499c-8879-2d6b78529876, and the clock.wal by the typecode 5ce17fce-4087-4169-b7ff-056cc58473f9.
What do you think?

netservers:
I just found a fly in the ointment. When I reboot by Ubuntu Xenial host system, it comes back with my block.wal and block.db partitions owned by root. Consequently my dockerized OSDs won't start.
Looking in osd_disk_activate.sh, I see that a workaround that fixes the journal is already in place:
```
  if [[ -n "${OSD_JOURNAL}" ]]; then
    wait_for_file ${OSD_DEVICE}
    chown ceph. ${OSD_JOURNAL}
  else
    wait_for_file $(dev_part ${OSD_DEVICE} 1)
    chown ceph. $JOURNAL_PART
  fi
```
It looks like we need the same thing for the block.wal and block.db.

leseb:
Using typecode is fine for me too, and yes this is only valid when 

You're right `ceph-disk` already applies typecode for these partitions:

* block.wal: 5ce17fce-4087-4169-b7ff-056cc58472be
* block.db: 30cd0809-c2b2-499c-8879-2d6b785292be

Let me re-work the patch for this.

leseb:
@netservers new version, I also added other commits too.

leseb:
jenkins test  ceph_ansible-kraken-xenial_cluster

leseb:

netservers:
OK, I'm testing and it's not sorting out permissions when my OSD containers start.

I've not really looked too closely, but I suspect that $OSD_BLUESTORE_BLOCK_WAL and $OSD_BLUESTORE_BLOCK_DB are undefined. They are normally only defined during prepare. If they are undefined, I suppose they should be initialised by reading the symlink in the OSD folder.

leseb:
Oh right I haven't tested the activation, only the prepare which went well.
You're right OSD_BLUESTORE_BLOCK_DB and OSD_BLUESTORE_BLOCK_WAL will only exist if you do prepare and activate at the same time which is not recommended.
Worst in the case, the activate you will, OSD_DEVICE will exist but OSD_BLUESTORE_BLOCK_WAL and OSD_BLUESTORE_BLOCK_DB won't be declared and point to OSD_DEVICE.

This bluestore activation is going to be though...

leseb:
Actually the activation won't *rework* **only** after a reboot.

netservers:
I'm getting close here, but I think I see a problem with the typecodes you're matching against. My WAL and DB partitions have typecodes as follows:
DB: 30cd0809-c2b2-499c-8879-2d6b78529876
WAL: 5ce17fce-4087-4169-b7ff-056cc58473f9

I think that the typecodes you're matching against are the precursors to those typecodes. See here:
[https://raw.githubusercontent.com/ceph/ceph/72f0b2aa1eb4b7b2a2222c2847d26f99400a8374/src/ceph-disk/ceph_disk/main.py](url)

Codes ending 2be are marked as "tobe", and the forms I'm using are marked as "ready".

Arguably all of these typecodes are candidates for chowning, and they are all ceph paritions.

leseb:
@netservers I think I only looked at one of the two, since the beginning is same I didn't pay attention. Resending a new version right away.

leseb:
@netservers the re-activation after a server reboot should be not be a problem if we pass the `OSD_BLUESTORE_BLOCK_DB`  and `OSD_BLUESTORE_BLOCK_WAL` during the activation container. Then once the machine is up, a simple `docker start` should work.

netservers:
@leseb I did a bit more work on this last night and my modified version is now working for me. As I'm unsure how to create a pull request for a pull request, I'll just paste my patch here.

I decided against passing-in the OSD_BLUE... variables again, as it struck me that it's superfluous and potentially contradictory information. We already know which devices we need to chown, as they're symlinked from the OSD directory.

Although this modification does what it sets out to do and perfectly chmods the device, I've still got a (probably unrelated) problem whereby the OSDs don't come "up" when I reboot the host system. The OSD daemons are running without any error output, but according to the mons they are "down". Like I say this is probably unrelated, but I thought I should mention it..

```diff --git a/ceph-releases/luminous/ubuntu/16.04/daemon/common_functions.sh b/ceph-releases/luminous/ubuntu/16.04/daemon/common_functions.sh                                                                                                                  │···························
index 6cd13f6..2329986 100755                                                                                                                                                                                                                                 │···························
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/common_functions.sh                                                                                                                                                                                          │···························
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/common_functions.sh                                                                                                                                                                                          │···························
@@ -255,15 +255,15 @@ function apply_ceph_ownership_to_disks {                                                                                                                                                                                                │···························
     wait_for_file ${OSD_JOURNAL}                                                                                                                                                                                                                             │···························
     chown --verbose ceph. ${OSD_JOURNAL}                                                                                                                                                                                                                     │···························
   elif [[ ${OSD_BLUESTORE} -eq 1 ]]; then                                                                                                                                                                                                                    │···························
-    dev_real_path=$(resolve_symlink $OSD_BLUESTORE_BLOCK_WAL $OSD_BLUESTORE_BLOCK_DB $OSD_DEVICE)                                                                                                                                                            │···························
-    for partition in $(list_dev_partitions $dev_real_path); do                                                                                                                                                                                               │···························
-      if [[ "$(get_part_typecode $partition)" == "5ce17fce-4087-4169-b7ff-056cc58472be" ]]; then                                                                                                                                                             │···························
-        chown --verbose ceph. $partition                                                                                                                                                                                                                     │···························
-      fi                                                                                                                                                                                                                                                     │···························
-      if [[ "$(get_part_typecode $partition)" == "30cd0809-c2b2-499c-8879-2d6b785292be" ]]; then                                                                                                                                                             │···························
-        chown --verbose ceph. $partition                                                                                                                                                                                                                     │···························
-      fi                                                                                                                                                                                                                                                     │···························
-      if [[ "$(get_part_typecode $partition)" == "89c57f98-2fe5-4dc0-89c1-f3ad0ceff2be" || "$(get_part_typecode $partition)" == "cafecafe-9b03-4f30-b4c6-b4b80ceff106" ]]; then                                                                              │···························
+    dev_real_path=$(resolve_symlink $OSD_BLUESTORE_BLOCK_WAL $OSD_BLUESTORE_BLOCK_DB)                                                                                                                                                                        │···························
+    for partition in $(list_dev_partitions $OSD_DEVICE) $dev_real_path; do                                                                                                                                                                                   │···························
+      part_code="$(get_part_typecode $partition)"                                                                                                                                                                                                            │···························
+      if [[ "$part_code" == "5ce17fce-4087-4169-b7ff-056cc58472be"  ||                                                                                                                                                                                       │···························
+            "$part_code" == "5ce17fce-4087-4169-b7ff-056cc58473f9"  ||                                                                                                                                                                                       │···························
+            "$part_code" == "30cd0809-c2b2-499c-8879-2d6b785292be"  ||                                                                                                                                                                                       │···························
+            "$part_code" == "30cd0809-c2b2-499c-8879-2d6b78529876"  ||                                                                                                                                                                                       │···························
+            "$part_code" == "89c57f98-2fe5-4dc0-89c1-f3ad0ceff2be"  ||                                                                                                                                                                                       │···························
+            "$part_code" == "cafecafe-9b03-4f30-b4c6-b4b80ceff106" ]]; then                                                                                                                                                                                  │···························
         chown --verbose ceph. $partition                                                                                                                                                                                                                     │···························
       fi                                                                                                                                                                                                                                                     │···························
     done                                                                                                                                                                                                                                                     │···························
diff --git a/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_activate.sh b/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_activate.sh                                                                                    │···························
index 8e52971..7427e88 100644                                                                                                                                                                                                                                 │···························
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_activate.sh                                                                                                                                                                           │···························
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_activate.sh                                                                                                                                                                           │···························
@@ -15,8 +15,6 @@ function osd_activate {                                                                                                                                                                                                                     │···························
   # watch the udev event queue, and exit if all current events are handled                                                                                                                                                                                   │···························
   udevadm settle --timeout=600                                                                                                                                                                                                                               │···························
                                                                                                                                                                                                                                                              │···························
-  apply_ceph_ownership_to_disks                                                                                                                                                                                                                              │···························
-                                                                                                                                                                                                                                                             │···························
   DATA_PART=$(dev_part ${OSD_DEVICE} 1)                                                                                                                                                                                                                      │···························
   MOUNTED_PART=${DATA_PART}                                                                                                                                                                                                                                  │···························
                                                                                                                                                                                                                                                              │···························
@@ -36,6 +34,11 @@ function osd_activate {                                                                                                                                                                                                                    │···························
   OSD_PATH=$(get_osd_path $OSD_ID)                                                                                                                                                                                                                           │···························
   OSD_KEYRING="$OSD_PATH/keyring"                                                                                                                                                                                                                            │···························
   OSD_WEIGHT=$(df -P -k $OSD_PATH | tail -1 | awk '{ d= $2/1073741824 ; r = sprintf("%.2f", d); print r }')                                                                                                                                                  │···························
+  OSD_BLUESTORE_BLOCK_DB=$(resolve_symlink "${OSD_PATH}block.db")                                                                                                                                                                                            │···························
+  OSD_BLUESTORE_BLOCK_WAL=$(resolve_symlink "${OSD_PATH}block.wal")                                                                                                                                                                                          │···························
+  apply_ceph_ownership_to_disks                                                                                                                                                                                                                              │···························
+                                                                                                                                                                                                                                                             │···························
   ceph ${CLI_OPTS} --name=osd.${OSD_ID} --keyring=$OSD_KEYRING osd crush create-or-move -- ${OSD_ID} ${OSD_WEIGHT} ${CRUSH_LOCATION}                                                                                                                         │···························
                                                                                                                                                                                                                                                              │···························
   log "SUCCESS"
```

leseb:
@netservers I just pushed a neww version based on your patch, I added you as a Co-Author on the last commit. Can I get a proper sign-off from you? (couldn't find one so I just used your github profile page...).

I added a small change on the activate to make sure the `readlink` does not fail if not using bluestore.

leseb:
@netservers I have the last version working, do you mind testing again?
Thanks!

netservers:
@leseb I've not tested this yet, but I see an error that I had corrected in my patch. This line:
`for partition in $(list_dev_partitions $OSD_DEVICE $dev_real_path); do`
Should really be:
`for partition in $(list_dev_partitions $OSD_DEVICE) $dev_real_path; do `
You don't want to pass the wal and db devices into list_dev_partitions. If I remember right that function enumerates partitions within the devices that it's passed. "$dev_real_path" is already a set of partitions.

leseb:
@netservers hum I remember having issues with this line, that's why I changed it this way. Any chance to give this PR a try?

leseb:
@netservers Alright I'm resurrecting this one, as I need some commits for other PRs.

leseb:
CI issue is the race condition again... Moving ahead with this one.

PR #617, opened 2017-04-26
Fixes weight calculations for bluestore OSD whereby weights are alway…

netservers:
…s set to 0.00.

This was due to osd_disk_activate.sh and/or osd_common.sh calculating the weight as
the number of TiB for the OSD metadata partition which is normally 100MB.
Instead it will calculate the number of TiB on the ${OSD_PATH}/block device.


leseb:
I'd prefer to put Bluestore fixes in Luminous only since it *might* be considered stable once Luminous is out.

Issue #338, opened 2016-07-20
Replace Sigil in Kubernetes Example with more common tool


elsonrodriguez:
Sigil is a bit of a barrier to entry, need to either replace the generation of the templates with a more common pattern, perhaps based on bash:

https://github.com/kubernetes/kubernetes/blob/master/cluster/aws/config-default.sh
https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html

leseb:
@elsonrodriguez good idea

PR #618, opened 2017-04-27
Jinja template processor for Kubernetes example

timn:
These two commits add templates and code to use jinja2 as template processor. sigil can still be enabled setting the TEMPLATE_ENGINE variable.

sigil is not easy to come by, while jinja2 is available on Fedora (python3-jinja2-cli package) and can be installed easily using pip ([jinja2-cli](https://pypi.python.org/pypi/jinja2-cli) package). The language is used for popular projects like ansible or Django.

This addresses issue #338 even though it is not entirely without dependencies.

### DIFF
--- a/examples/kubernetes/generator/generate_secrets.sh
+++ b/examples/kubernetes/generator/generate_secrets.sh
@@ -1,5 +1,23 @@
 #!/bin/bash
 
+TEMPLATE_ENGINE=jinja
+
+proc-template() {
+  FILE=$1
+  shift
+  if [ "$TEMPLATE_ENGINE" == "sigil" ]; then
+    echo $(sigil -p -f ${FILE}.tmpl "$@")
+  else
+    TMPFILE=$(mktemp)
+    for a in $@; do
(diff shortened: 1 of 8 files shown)


examples/kubernetes/generator/generate_secrets.sh
```diff
@@ -1,55 +1,73 @@
 #!/bin/bash
 
+TEMPLATE_ENGINE=jinja
+
+proc-template() {
```
leseb: please use indent of 2.

timn: I have done so and squashed the changes in the original commit.

Didn't see this as I use a tab size of 2, sorry.

font:
Switching to use jinja with these changes actually breaks the use of environment variables to override defaults as required by this step https://github.com/ceph/ceph-docker/tree/master/examples/kubernetes#override-the-default-network-settings. The `jinja2` CLI requires the use of `-D key=value` to define template variables to override the defaults. With `sigil`, we relied on environment variables to override defaults.

PR #619, opened 2017-04-27
Kubernetes example compatibility improvements

timn:
This pull request makes the osd container unprivileged by default. That can make it easier to run on a Kubernetes cluster if privileged containers are disallowed by default. I have added comments that state what needs to be done in order to use a raw disk on an OSD. The second patch just adds a comment with an image to use the CentOS version. The Ubuntu version consistently failed for me when running more than one OSD, while the CentOS image worked. This might make it easier for others to use an alternative image.

I understand that the privileged context might be intended and this is why I split this pull request off of my previous one. It just fits the default example with an emptydir for storage where privileges are not required.

### DIFF
--- a/examples/kubernetes/ceph-osd-v1-ds.yaml
+++ b/examples/kubernetes/ceph-osd-v1-ds.yaml
@@ -43,6 +43,7 @@ spec:
       containers:
         - name: osd-pod
           image: ceph/daemon:latest
+          #image: ceph/daemon:tag-build-master-kraken-centos-7
           imagePullPolicy: Always
           volumeMounts:
             - name: devices
--- a/examples/kubernetes/ceph-osd-v1-ds.yaml
+++ b/examples/kubernetes/ceph-osd-v1-ds.yaml
@@ -17,9 +17,12 @@ spec:
       nodeSelector:
         node-type: storage
       volumes:
-        - name: devices
-          hostPath:
-            path: /dev
+# Comment back in to use direct device access.
+# Note that this will also require enabling privileged access and
+# the appropriate volumeMount below.
+#        - name: devices
+#          hostPath:
+#            path: /dev
         - name: ceph
           emptyDir: {}
 #          hostPath:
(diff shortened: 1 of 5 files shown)


examples/kubernetes/ceph-osd-v1-ds.yaml
```diff
@@ -17,9 +17,12 @@ spec:
       nodeSelector:
         node-type: storage
       volumes:
-        - name: devices
-          hostPath:
-            path: /dev
+# Comment back in to use direct device access.
+# Note that this will also require enabling privileged access and
```
leseb: 👎 
This is only true when you do journal collocation and when you use a journal file for Ceph.
If you use a raw partition for the journal you will need access to that device.

timn: I don't understand, isn't that direct device access, no matter whether using it as data or journal drive?

I'm happy to change the comment and would welcome some suggestion if the general idea to run non-privileged by default by itself is acceptable and has a chance of being merged.

leseb: If you don't expose `/dev/` how to you access the device to use it as an OSD? Moreover without privileged you won't be able to partition the dev etc.
How did you test this?

timn: The default example uses emptyDir storage. Therefore, I think the example should also run unprivileged and therefore this pull request. The comments are meant to show that if moving away from the example by using actual storage disks, the container must run privileged.

This also holds true when using a host directory for data storage. In our use case, we use all machines in a robot lab (otherwise also used by developers as desktops) to run experiments in simulation, and hence we do not have dedicated storage disks or partitions. We just want to achieve some temporary failure safety until the relevant data from the experiment has been extracted. Then we dispose of the cluster setup and can re-create it later for another experiment. In that sense we probably do not have the typical data-center storage cluster for long-running services, but use it every off and on as necessary. We use ceph mostly to get minor fail-safety (say a single machine down at a time) and to pool storage space of all nodes.

ErwanAliasr1:
Please rebase against master as PR #927 got merged.
If you need help to understand how to update your PR against the new scheme, feel free to ping me.

leseb:
Closing this due to inactivity.

PR #632, opened 2017-05-10
rm NOT_MAINTAINED_ANYMORE

ktdreyer:
It is literally not maintained any more, so delete it from the tree to avoid clutter.

The purpose of this commit is to make it easier to browse through the currently-maintained code.


GregMeno:
woo hoo!

PR #635, opened 2017-05-11
changing tox factor to include tests on latest stable ceph-ansible

GregMeno:
Signed-off-by: Gregory Meno <gmeno@redhat.com>


andrewschoen:
The CI for pull requests only test ceph-ansible@master so this change will not be tested by the PR tests here, will merge without waiting.

Issue #626, opened 2017-05-08
Kubernetes: monitors should not nuke their own monmaps


paralin:
```
e10 not in monmap and have been in a quorum before; must have been removed
```

This is due to the startup script overwriting the monmap. Even if I map in /var/lib/ceph/mon to the container, the mons will not start up correctly again if the pod is re-created. The script needs to be adjusted to not clobber an existing monmap.

Edit: investigating a bit more, this might not actually be the complete story - I'm pretty frustrated at this point that I keep losing my cluster's data

PR #637, opened 2017-05-12
daemon: try to fetch the last version of monmap

leseb:
When the container restarts we re-inject the old version of the map
which results in the monitor failing to start/join the existing quorum.
So now we fetch the latest version prior to re-run it.

Fixes: https://github.com/ceph/ceph-docker/issues/626
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1448954

(cherry picked from commit b59bcbec66c3ca1a444bac5676d9773933883d66)


leseb:
@ktdreyer can we have a downstream build with this patch?
Do you need a tag?

leseb:
@ktdreyer never mind, I need to do the push downstream, forgot about that detail

PR #644, opened 2017-05-16
fix kube-dns addr, added secret type

ReSearchITEng:
fixed kube-dns address, added notes on jinja2, added secret type:     `--type="kubernetes.io/rbd"`


leseb:
Thanks!

PR #641, opened 2017-05-15
tests: split os-type and scenario factors

andrewschoen:
This allows us to add new testing scenarios without repeating the
configuration needed for different operating systems.

Signed-off-by: Andrew Schoen <aschoen@redhat.com>

PR #645, opened 2017-05-16
build_imgs.sh: simplify non-PR case in copy_dirs()

ktdreyer:
build_imgs.sh's copy_dirs() method is a no-op if the top-level "base", "daemon", and "demo" directories already exist. We use this conditional to determine what we're testing in Travis CI: are we testing a GitHub PR, or a pushed branch?

Simplify the code here to simply exit the function early. This eliminates a level of indentation so the function is easier to read and understand.


andrewschoen:
tests are failing here because https://github.com/ceph/ceph-docker/pull/641 renames them

leseb:
jenkins test this please

PR #652, opened 2017-05-22
Fix sigil and provide jinja2 installation instructions

font:


leseb:
@font please look at the CI https://jenkins.ceph.com/job/ceph-docker-lint/70/consoleFull#-6924622055ce3975f-219a-4f41-bab6-8eeac49fe04f and fix your bash issues.

font:
@leseb Haha, I didn't actually modify that code. :) But in any case I've fixed the bash quote issues introduced by #618. Looks like the CI missed those errors previously.

Also, moving to jinja as introduced by #618 actually breaks the use of environment variables to override defaults as required by this step https://github.com/ceph/ceph-docker/tree/master/examples/kubernetes#override-the-default-network-settings. Jinja requires the use of `-D key=value` to define template variables to override the defaults.

leseb:
@font thanks for fixing the other portions of the code :p
About this `-D` issue, do we have a workaround?

font:
@leseb Yes, we can just modify the `ceph.conf.<jinja|tmpl>` template directly or switch the template engine back to `sigil`.

PR #654, opened 2017-05-25
start mon: use -c instead of --inject-monmap

andrewschoen:
In a scenario where all mon hsots are rebooted some mons fail to come
back up, using -c which points at the configuration file is safer
because we can know that all monitors are defined there.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1455357

Signed-off-by: Andrew Schoen <aschoen@redhat.com>


leseb:
@andrewschoen can you fix this for kraken and luminous as well?

andrewschoen:
@leseb I've updated this for kraken and luminous as well

PR #656, opened 2017-05-30
Add tests/teardown.sh

zmc:
This is useful when executing tox outside a CI environment; if tests
fail nothing is cleaned up. While that might be desirable, let's provide
a way to manually tear everything down.

Signed-off-by: Zack Cerza <zack@redhat.com>


leseb:
I think we can ignore the lint bash for this one.

PR #659, opened 2017-05-31
DNM: Rewrite tox.sh in Python

zmc:


leseb:
I'm not really against that but I'd like to understand the rationale behind rewriting `tox.sh` in Python.
What's the added value?

Thanks.

tests/run_tox.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python
+"""
+Proxy script from tox. This is an intermediate script so that we can setup
+the environment properly then call ceph-ansible for testing, and finally tear
+down, while keeping tox features of simplicity and combinatorial confgiruation.
```
ktdreyer: confgiruation -> configuration

zmc: Hey man it's really important to preserve old typos ;)

(I'll fix it)

tests/run_tox.py
```diff
@@ -0,0 +1,167 @@
...
+import subprocess
+import time
+
+
+logging.basicConfig()
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
+
+
+def sh(cmd, cwd=None, ignore_errors=False, capture=False):
```
zmc: I suppose we could do that if we're ok with `check_call` merging `stdout` and `stderr`

zmc:
@ktdreyer, thanks for responding. @leseb, the idea behind this rewrite is to move towards enabling reuse and unit testing (if we want that), as @ktdreyer says. I wouldn't consider its current state as ready-to-merge yet, of course - hence the DNM. I mainly filed it early so I could test https://github.com/ceph/ceph-build/pull/736 ;)

tests/run_tox.py
```diff
@@ -0,0 +1,167 @@
...
+    sh("{TOXINIDIR}/tests/teardown.sh".format(**env),
+        cwd=env['CEPH_ANSIBLE_SCENARIO_PATH'])
+
+
+def main():
+    env = os.environ
+    setup(env)
+    try:
+        test(env)
+    except Exception:
```
zmc: Oh, yeah, that is certainly vestigial. As I mention [here](https://github.com/ceph/ceph-docker/pull/659#issuecomment-305571306) this PR isn't mature this very second. I left that stanza there because I was considering adding some error handling and/or reporting.

andrewschoen:
I agree with @ktdreyer here that switching to python could allow for more testable and loosely coupled code. However, if we are only switching tox.sh to tox.py we don't really get much of that benefit. I think those arguments are more valid if we decide to switch the entire project to python, but that's a much larger discussion.

All that being said (and I can't believe I'm saying this), I think that tox.sh is actually more readable than tox.py here. I had always thought that if tox.sh was rewritten I'd probably do it in ansible as I think the 'top down' way you read an ansible playbook makes sense here. Also because ansible has great support for package installs and running arbitrary commands. I had just never found a strong reason to port tox.sh as we typically don't have any issues with it.

leseb:
I agree with Andrew that with the introduction to Python, the script is definitely less readable and it also has a lot of `sh(` and `cmd`. It looks like a bash that has literally translated to Python with.
In the end, I don't much value in the current state as we are still using a lot of embedded shell command into this Python code.

It's just less readable and harder to hack I guess. I think bash was perfect for this.

zmc:
Closing as it's not really the direction we want to go in

ktdreyer:
I think this will be a great approach for other scripts in the tree

zmc:
Agreed and I may have time to contribute to that effort in a few weeks, but for now my priorities are taking me elsewhere :)

PR #660, opened 2017-06-01
daemon: do not run the monitor twice

leseb:
First we remove the initial monmap, this monmap is really dangerous
after the initial deployment. Thus it should be deleted so we are sure
it's used later.

Also prior to this patch we were running the monitor which would fail
instantly, then create a socket and the other process would not run.
Removing this useless start and let's just the monitor start by reading
the Ceph configuration file, which will contain the list of all
monitors.

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/start_mon.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/start_mon.sh
@@ -125,9 +125,19 @@ function start_mon {
 
     # Prepare the monitor daemon's directory with the map and keyring
     ceph-mon --setuser ceph --setgroup ceph --cluster ${CLUSTER} --mkfs -i ${MON_NAME} --inject-monmap $MONMAP --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
+
+    # Never re-use that monmap again, otherwise we end up with partitioned Ceph monitor
+    # The initial mon **only** contains the current monitor, so this is useful for initial bootstrap
+    # Always rely on what has been populated after the other monitors joined the quorum
+    rm -f $MONMAP
   else
     log "Existing mon, trying to rejoin cluster..."
-    ceph-mon --setuser ceph --setgroup ceph --cluster ${CLUSTER} -i ${MON_NAME} -c /etc/ceph/${CLUSTER}.conf --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
+    if [[ "$KV_TYPE" != "none" ]]; then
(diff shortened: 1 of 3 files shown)


alfredodeza:
could we try to avoid merging these with failing tests? if we really need to do that maybe add a note on why is that so that no one else is surprised when failures are ignored

ceph-releases/jewel/ubuntu/14.04/daemon/start_mon.sh
```diff
@@ -125,9 +125,19 @@ function start_mon {
...
+    # Always rely on what has been populated after the other monitors joined the quorum
+    rm -f $MONMAP
   else
     log "Existing mon, trying to rejoin cluster..."
-    ceph-mon --setuser ceph --setgroup ceph --cluster ${CLUSTER} -i ${MON_NAME} -c /etc/ceph/${CLUSTER}.conf --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
+    if [[ "$KV_TYPE" != "none" ]]; then
+      # This is needed for etcd or k8s deployments as new containers joining need to have a map of the cluster
+      # The list of monitors will not be provided by the ceph.conf since we don't have the overall knowledge of what's already deployed
+      # In this kind of environment, the monmap is the only source of truth for new monitor to attempt to join the existing quorum
+      ceph-mon --setuser ceph --setgroup ceph --cluster ${CLUSTER} -i ${MON_NAME} --inject-monmap $MONMAP --keyring $MON_KEYRING --mon-data "$MON_DATA_DIR"
```
andrewschoen: If we are doing ``rm -f $MONMAP`` above where are we getting the value of ``$MONMAP`` for ``--inject-monmap`` here?

leseb: When using k8s, you get the `$MONMAP` from the `get_mon_config` function, which will store the `$MONMAP` on the disk.

The `rm -f` is only triggered during the initial bootstrap. If you're not a new monitor you go in the `else` line 133 so the `$MONMAP` is present.

The `$MONMAP` is only needed for Ansible or manual deployment and only for initial bootstrap.

leseb:
@alfredodeza sorry about that, I tried this patch in the QE lab with success. Deadline is tomorrow so I'm trying to go as fast as possible. Obviously without neglecting test :).

ceph-releases/jewel/ubuntu/14.04/daemon/start_mon.sh
andrewschoen: Thanks for the explanation.

PR #668, opened 2017-06-13
daemon: improve debugging line

leseb:
When setting DEBUG=verbose, this only activates bash tracing with set
-x. This patch adds more info during the debug like the name of the
script and the line being executed. See:

$ ./a.sh
+./a.sh:6: echo a
a

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/debug.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/debug.sh
@@ -11,6 +11,9 @@ for option in $(comma_to_space ${DEBUG}); do
   case $option in
     verbose)
       echo "VERBOSE: activating bash debugging mode."
+      echo "To run Ceph daemons in debugging mode, pass the CEPH_ARGS variable like this:"
+      echo "-e CEPH_ARGS='--debug-ms 1 --debug-osd 10'"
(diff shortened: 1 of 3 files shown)


ceph-releases/jewel/ubuntu/14.04/daemon/debug.sh
```diff
@@ -11,6 +11,8 @@ for option in $(comma_to_space ${DEBUG}); do
   case $option in
     verbose)
       echo "VERBOSE: activating bash debugging mode."
+      DEBUG_MS=20
+      echo "VERBOSE: Ceph daemon will now run in debugging level $DEBUG_MS"
```
liewegas: This is unlikely to be the ceph debugging the user needs.  I would suggest instead having a different env var CEPH_ARGS that is passed through, independent of $VERBOSE?

leseb: What kind of debug options can we pass to CEPH_ARGS? Is that going to help running daemon with a higher debugging level?

liewegas:
anythin you pass on the command line (e.g., CEPH_ARGS="--debug-ms 1 --debug-osd 10")

leseb:
@liewegas alright thanks, will rework this then.

PR #671, opened 2017-06-16
daemon: improve ceph health check error

leseb:
Prior to this change if the command to get the ceph health failed we
didn't have any indication of what went wrong. This is now easier to
debug.


leseb:
CI was failing on the race condition again but most of the job ran well. Everything would have failed if this commits introduced an issue.

PR #674, opened 2017-06-22
daemon: fix array

leseb:
The double brackets are non-needed on an array and even if I wanted to
use them they were mispositionned.

```
creating /etc/ceph/ceph.client.admin.keyring
creating /etc/ceph/ceph.mon.keyring
creating {item//:/ }
bufferlist::write_file({item//:/ }): failed to open file: (2) No such
file or directory
could not write {item//:/ }
```


leseb:
@maxmanuylov

PR #677, opened 2017-06-23
daemon: fix array

leseb:
Problem: array=("${item//:/ }")

Quoting an array that contains multiple words (processed by bash)
doesn't treat the array with multiple fields, in the current case [0] is
'ceph.keyring Rgw' where it should only be 'ceph.keyring'.

Using a new IFS and read like suggested in
https://github.com/koalaman/shellcheck/wiki/SC2206 helps us.

That was though... One again thanks koalaman.

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/config.kv.etcd.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/config.kv.etcd.sh
@@ -60,7 +60,8 @@ function get_mon_config {
     ceph-authtool "$MON_KEYRING" --create-keyring --gen-key -n mon. --cap mon 'allow *'
 
     for item in ${OSD_BOOTSTRAP_KEYRING}:Osd ${MDS_BOOTSTRAP_KEYRING}:Mds ${RGW_BOOTSTRAP_KEYRING}:Rgw; do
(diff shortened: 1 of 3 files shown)


leseb:
@maxmanuylov I tested locally, I think we are good this time :)

leseb:
The CI doesn't test `config.kv.etcd.sh`and sloppy bash is happy so merging this.

leseb:
@maxmanuylov just launched a build on the docker hub, you should get a new version soon.

maxmanuylov:
@leseb works now, thanks!

maxmanuylov:
@leseb no, I'm wrong, actually there is still a problem. Mon is now OK, but osd fails:
```
common_functions.sh: line 132: /dev/sdb: Permission denied
```
Looks like [this line](https://github.com/ceph/ceph-docker/commit/628764b52ce0547d7bd6f47f128476b7f490aae0#diff-66a7c5bf7b7739b5739cefd9ec24ec07R132) tries to execute disk file ("/dev/sdb" in my case).

leseb:
@maxmanuylov looking into this.

leseb:
fixed in https://github.com/ceph/ceph-docker/pull/689

Issue #680, opened 2017-06-26
ceph-mgr "cannot open /etc/ceph/-i.conf"


AlessandroLorenzi:
I'm testing luminous ubuntu 16 container, i have a strange error: 

1. osds up:
```
  services:
    mon: 2 daemons, quorum srv1,srv2
    mgr: no daemons active
    osd: 1 osds: 1 up, 1 in
```
2. generating keys
```
  ceph  auth get-or-create mgr.srv1 mon 'allow *' -o /var/lib/ceph/mgr/ceph-srv1/keyring
```

3. running docker
```
/usr/bin/docker run \
        --net host -m 0b -e HOSTNAME=srv1 \
        -e MGR_NAME=srv1 \
        -e CEPH_PUBLIC_NETWORK=192.168.2.0/24 \
        -e CLUSTER=ceph \
        -v /var/lib/ceph/:/var/lib/ceph -v /etc/ceph:/etc/ceph  \
        --name ceph-mgr \
        ceph/daemon:tag-build-master-luminous-ubuntu-16.04 \
         mgr
[...]
ownership of '/var/lib/ceph/radosgw/karin-hannafin' retained as ceph:ceph
ownership of '/var/lib/ceph/radosgw/srv7' retained as ceph:ceph
ownership of '/var/lib/ceph/radosgw' retained as ceph:ceph
ownership of '/var/lib/ceph/tmp' retained as ceph:ceph
2017-06-26 15:30:08  /entrypoint.sh: static: does not generate config
2017-06-26 15:30:08  /entrypoint.sh: SUCCESS
##### entrypoint is ok here
##### running ceph-mgr command
2017-06-26 15:30:08.639144 7f548a704380 -1 did not load config file, using default settings.
2017-06-26 15:30:08.640069 7f548a704380 -1 Errors while parsing config file!
2017-06-26 15:30:08.640072 7f548a704380 -1 parse_file: cannot open /etc/ceph/-i.conf: (2) No such file or directory
2017-06-26 15:30:08.640073 7f548a704380 -1 parse_file: cannot open ~/.ceph/-i.conf: (2) No such file or directory
2017-06-26 15:30:08.640073 7f548a704380 -1 parse_file: cannot open -i.conf: (2) No such file or directory
2017-06-26 15:30:08.640445 7f548a704380 -1 Errors while parsing config file!
[...]

PR #684, opened 2017-06-27
Daemon: fix opts

AlessandroLorenzi:
Refs: #680


leseb:
Please fix master instead, I'll update the branch after that. Thanks!

PR #687, opened 2017-06-27
dockerfile: chmod +x is not needed

leseb:
Since we directly call the bash command, using chmod +x to make the file
executable is not needed. Moreover since we don't pass any argument to
the script generate_entrypoint.sh there is no need for '-c' option.

Somehow this resolves the issue during our build:

```
Step 16 : RUN chmod +x /generate_entrypoint.sh &&   bash -c
"/generate_entrypoint.sh" &&   rm -f /generate_entrypoint.sh &&   bash
-n /*.sh
 ---> Running in 6fa5c8d5735e
 bash: /generate_entrypoint.sh: /bin/bash: bad interpreter: Text file
 busy
```


leseb:
jenkins test ceph_ansible-luminous-centos7-dedicated_journal

djgalloway:
Thanks for this!  Looks much better.

Issue #681, opened 2017-06-26
Kubernetes / CoreOS : ceph-mon stuck in  Waiting: CrashLoopBackOff


fjudith:
## Environment

```
Kubernetes: 1.6.6
CoreOS: 1437.2.0
FsType: xfs
```
Hello,

I've followed the kubernetes example after mounting xfs disks in the osd-directory (e.g. `/home/core/data/ods/`). But ceph-mon pods remain is the state `Waiting: CrashLoopBackOff`.

#### logs
```
2017-06-26T19:37:06.550463463Z ++start_mon.sh:29: get_ip(): ip -4 -o a s eth0
2017-06-26T19:37:06.550624411Z ++start_mon.sh:29: get_ip(): awk '{ sub ("/..", "", $4); print $4 }'
2017-06-26T19:37:06.551614745Z ++start_mon.sh:29: get_ip(): grep 'scope global'
2017-06-26T19:37:06.551901869Z +start_mon.sh:94: start_mon(): MON_IP=10.2.62.8
2017-06-26T19:37:06.552750828Z ++start_mon.sh:95: start_mon(): get_network eth0 4
2017-06-26T19:37:06.552760035Z ++start_mon.sh:47: get_network(): local nic=eth0
2017-06-26T19:37:06.552764151Z ++start_mon.sh:49: get_network(): local ip_version=4
2017-06-26T19:37:06.552768007Z ++start_mon.sh:51: get_network(): case "$ip_version" in
2017-06-26T19:37:06.552771988Z ++start_mon.sh:69: get_network(): is_available ip
2017-06-26T19:37:06.55277591Z ++common_functions.sh:46: is_available(): command -v ip
2017-06-26T19:37:06.552779766Z ++start_mon.sh:70: get_network(): ip -4 route show dev eth0
2017-06-26T19:37:06.55278361Z ++start_mon.sh:70: get_network(): grep proto
2017-06-26T19:37:06.553383491Z ++start_mon.sh:70: get_network(): grep -v default
2017-06-26T19:37:06.554373281Z ++start_mon.sh:70: get_network(): awk '{ print $1 }'
2017-06-26T19:37:06.554385444Z ++start_mon.sh:70: get_network(): grep /
2017-06-26T19:37:06.554575871Z +start_mon.sh:95: start_mon(): CEPH_PUBLIC_NETWORK=10.2.62.0/24
2017-06-26T19:37:06.554584469Z +start_mon.sh:96: start_mon(): ip_version=4
[...]

PR #688, opened 2017-06-28
daemon: fix array for creating monmap in k8s

leseb:
Quoting an array that contains multiple words (processed by bash)
doesn't treat the array with multiple fields.

Using a new IFS and read like suggested in
https://github.com/koalaman/shellcheck/wiki/SC2206 helps us.

Fixes: https://github.com/ceph/ceph-docker/issues/681


leseb:

PR #689, opened 2017-06-28
daemon: use a null operator to declare a variable

leseb:
Simply using ${OSD_DEVICE:-none} will let the shell execute its content.
To avoid this we use the no-op :, which evaluates its
argument and then throws it away, rather than executing it.

### DIFF
--- a/ceph-releases/jewel/ubuntu/14.04/daemon/common_functions.sh
+++ b/ceph-releases/jewel/ubuntu/14.04/daemon/common_functions.sh
@@ -122,7 +122,7 @@ function dev_part {
 }
 
 function osd_trying_to_determine_scenario {
-  ${OSD_DEVICE:-none}
+  : "${OSD_DEVICE:-none}"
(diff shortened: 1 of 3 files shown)


ceph-releases/jewel/ubuntu/14.04/daemon/common_functions.sh
```diff
@@ -122,7 +122,7 @@ function dev_part {
 }
 
 function osd_trying_to_determine_scenario {
-  ${OSD_DEVICE:-none}
+  : "${OSD_DEVICE:-none}"
```
maxmanuylov: This line does not change the OSD_DEVICE value. If you want this var to get "none" value in case it is not set you should do `OSD_DEVICE="${OSD_DEVICE:-none}"`.

leseb: Fixed in https://github.com/ceph/ceph-docker/pull/690

PR #682, opened 2017-06-27
daemon: rework dmcrypt implementation

leseb:
Open LUKS devices when necessary, do not wait for ceph-disk to do it for
us.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #701, opened 2017-07-03
Workaround for ceph-disk not unlocking block device for bluestore

Blackclaws:
This is a temporary workaround until this is implemented upstream.

ceph-disk does not unlock the blockdevice for bluestore when using --dmcrypt with activate.

It can be instructed to unlock the blockdevice with ceph-disk activate-block --dmcrypt however I'm not sure thats a good idea.


leseb:
I've started something here already: https://github.com/ceph/ceph-docker/pull/682

Blackclaws:
Ah I see. I used python in my version to 1:1 copy what ceph-disk is doing internally with the dmcrypt keys. But your version looks more general and probably fixes the problem at a more root location.

leseb:
Yes thanks, I didn't know ceph-disk wasn't re-opening the block partition on activate. Do we have the same problem for WAL too?

Blackclaws:
I think ceph-disk isn't opening anything other than the data partition. However at least in standard configuration there isn't a seperate WAL set up. I could imagine the problem exists there as well.

PR #702, opened 2017-07-03
Fix for mgr not starting because of missing cluster name. 

Blackclaws:
Array expansion was missing. Hence only --cluster was added there. It then tried to open /etc/ceph/auth.conf which didn't exist and even if it did would have done the wrong thing.

Hence starting managers does not work in docker containers right now. They are needed for luminous however.

We could think about always starting managers with monitors as that is a recommended suggestion anyway. Would however reduce user control.


leseb:
This was fixed this morning by https://github.com/ceph/ceph-docker/pull/698

Blackclaws:
Ah sorry. Didn't see that.

leseb:
No worries.

PR #705, opened 2017-07-04
Update Docker to CMD

hailong226:
using calico networking, on stage entrypoint, cannot get ip. on stage cmd healthy, 
it looks like we have not been able to discover the network settings


leseb:
Can you explain why your commit is solving your problem?

leseb:
Closing this due to non-activity, feel free to re-open.

PR #715, opened 2017-07-10
Fix forego breaking docker build

kpcyrd:
`ADD` appears to download the file and also automatically untar the archive if it ends in .tgz.

The build started to fail because /forego.tgz is a directory.

I'm not sure how many docker versions the file needs to support, if this breaks things for somebody on older versions I'd recommend to remove `ADD` all together and wget it explicitly. This would also slightly reduce the image size if you get+remove the archive within the same layer.

### DIFF
--- a/ceph-releases/jewel/centos/7/base/Dockerfile
+++ b/ceph-releases/jewel/centos/7/base/Dockerfile
@@ -12,7 +12,7 @@ ENV CONFD_VERSION 0.10.0
 RUN rpm --import 'https://download.ceph.com/keys/release.asc'
 RUN rpm -Uvh http://download.ceph.com/rpm-${CEPH_VERSION}/el7/noarch/ceph-release-1-1.el7.noarch.rpm
 RUN yum install -y epel-release && yum clean all
-ARG PACKAGES="unzip ceph-mon ceph-osd ceph-mds ceph-base ceph-common ceph-radosgw rbd-mirror device-mapper sharutils etcd kubernetes-client e2fsprogs"
+ARG PACKAGES="unzip ceph-mon ceph-osd ceph-mds ceph-base ceph-common ceph-radosgw rbd-mirror device-mapper sharutils etcd kubernetes-client e2fsprogs wget"
 RUN yum install -y $PACKAGES && rpm -q $PACKAGES && yum clean all
 
 # Install confd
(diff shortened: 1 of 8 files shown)


guits:
jenkins test this please

guits:
@kpcyrd can you please rebase your branch ?

guits:
jenkins test this please

guits:
@kpcyrd if you take a look at the jobs failure, it looks your fix is not working.
I wonder about the fix you suggested in your first comment.

kpcyrd:
@guits I had issues reading the jenkins output, it appears a failing docker build doesn't fail the build directly. I've pushed a new version that avoids the whole magic `ADD` all together.

kpcyrd:
There's currently one failing build on jewel-centos7, I've only touched `luminous/ubuntu` so far. If you're ok with the change (and we figure out what's wrong with the failing build). I'd apply the change to all affected docker files.

guits:

kpcyrd:
@guits yes, the build stops failing on recent docker versions :)

`ADD` vs `COPY` is sort of controversal, `COPY A B` is strictly "copy A from outside to B inside the image". `ADD A B` is magic and it appears it's behaviour changed very recently. It's mostly the same as COPY (and often used like that), except when:

- A is an url: the url is fetched and added to the image
- A is an .tar.gz: the archive is extracted into the image (this is how base images are created)

It appears that very recently the behaviour was changed, so `ADD`ing an url _that ends with .tgz_ means download _and_ extract the archive. As a result `/forego.tgz` was either a directory or a file, depending on your docker version. Using wget directly is less ambiguous. You should be able to reproduce this by updating the docker version on your jenkins.

The image should be about 1.3M smaller as well, but I think that's neglectable.

guits:
Thx for this explanation @kpcyrd my question was more about the final image size you mentionned but even if as you said this is neglectable, lets do it the way you are suggesting since it will avoid ambiguous behavior. If you dont mind can you please apply this patch for other releases?

guits:
@kpcyrd can you squash your commits. There is no need to have several commits for this.

leseb:
jenkins test ceph_ansible-jewel-centos7-cluster

leseb:
jenkins test ceph_ansible-luminous-centos7-dedicated_journal

Issue #711, opened 2017-07-08
bluestore on /dev/loop0 device osd_disk_activate use wrong partition


osixia:
Hello,
I just tried to run an osd with bluestore on a virtual disk at /dev/loop0 and osd_disk_activate use wrong partition.

ceph-disk prepare created partition like /dev/loop0p1, /dev/loop0p3 ... 
but some partitions are partially hard coded in osd_disk_activate to use  /dev/loop01 and  /dev/loop03 and fail.
```
DATA_UUID=$(get_part_uuid "${OSD_DEVICE}"1)
...
LOCKBOX_UUID=$(get_part_uuid "${OSD_DEVICE}"3 || true)
```
also the following commands fail for bluestore
```
JOURNAL_PART=$(ceph-disk list "${OSD_DEVICE}" | awk '/ceph journal/ {print $1}') # This is privileged container so 'ceph-disk list' works
JOURNAL_UUID=$(get_part_uuid "${JOURNAL_PART}")
```

PR #720, opened 2017-07-12
fix device partition name for device like /dev/loop0

osixia:
Related issue: #711


guits:
@osixia you should have 1 commit, can you please rebase on master?

guits:
@osixia it's still not well rebased on master, it's even worse actually, you have now 6 commits

PR #740, opened 2017-08-03
Update teardown script

HardikDR:
PVs are not namespaced in kubernetes. This script can potentially delete all the other PVs in the cluster. Thus namespacing the deletion of the PVs. Also it would be a good practice to delete the namespace at last.


HardikDR:
Sorry for goofing up. I have created new PR to delete specific set of PVs and not all.
Basically `-n ceph` wont apply for PVs in kubernetes.

PR #741, opened 2017-08-03
Update teardown script further

HardikDR:
PVs are not namespaced , so we should pick up PVs only from ceph namespace.

### DIFF
--- a/examples/kubernetes/delete_ceph_cluster.sh
+++ b/examples/kubernetes/delete_ceph_cluster.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 kubectl delete secret ceph-secret-admin --namespace kube-system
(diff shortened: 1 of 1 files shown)


examples/kubernetes/delete_ceph_cluster.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 kubectl delete secret ceph-secret-admin --namespace kube-system
+kubectl get pv | grep 'ceph/' | awk {'print $1'} | xargs -I{} kubectl delete pv {}
```
leseb: no need to grep and then awk, awk can do both: `awk '/ceph/ {print $1}'`

HardikDR:
Thanks , that was a nice catch !!

PR #754, opened 2017-08-22
fix mds stat hang command

leseb:
For some reasons, when the pool does not exist, the stat hangs.
Adding a timeout to the command fixes the issue.


leseb:

leseb:

PR #760, opened 2017-08-28
daemon: do not log inside the container

leseb:
Logging inside the container is not useful since it writes to the
overlayfs partition, resulting in potential performance degradation on
the container.

If you need to check the logs, just look at journald.


BlaineEXE:
With this PR, will the logs should still be viewable by `docker logs <container>`, correct?

leseb:
@BlaineEXE yes, `docker logs` are fetched from `stdout`, we still log to `stdout`.

BlaineEXE:
@leseb: Great! Thank you!

PR #774, opened 2017-09-12
contrib: add a upstream/downstream resync bug

leseb:
will ease rhcs downstream resyncs


leseb:
From @ktdreyer https://pypi.python.org/pypi/gitchangelog

PR #790, opened 2017-09-25
osd_disk_activate: fix wrong catching of journal part

guits:
`ceph-disk list <device>` returns:

```
/dev/sdb :
 /dev/sdb1 ceph data, prepared, cluster test, osd.1, journal /dev/sdc2
 ```

 therefore, `awk '/ceph journal/ {print $1}'` won't catch the actual
 associated journal partition.

This commit fixes it by using sed instead.

### DIFF
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_activate.sh
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_activate.sh
@@ -17,7 +17,7 @@ function osd_activate {
     else
       CLI+=("${OSD_DEVICE}")
     fi
-    JOURNAL_PART=$(ceph-disk list "${CLI[@]}" | awk '/ceph journal/ {print $1}') # This is a privileged container so 'ceph-disk list' works
+    JOURNAL_PART=$(ceph-disk list "${CLI[@]}" | grep journal | sed -r 's/^.*\s([^ ]+)$/\1/') # This is a privileged container so 'ceph-disk list' works
     JOURNAL_UUID=$(get_part_uuid "${JOURNAL_PART}" || true)
   fi
 

### CODE REVIEW

File: ceph-releases/luminous/ubuntu/16.04/daemon/osd_scenarios/osd_disk_activate.sh [hunk 1]
```diff
@@ -30,7 +30,7 @@ function osd_activate {
   if [[ ${OSD_DMCRYPT} -eq 1 ]] && [[ ${OSD_FILESTORE} -eq 1 ]] && [[ ${OSD_BLUESTORE} -eq 0 ]]; then
     DATA_UUID=$(get_part_uuid "$(dev_part "${OSD_DEVICE}" 1)")
     LOCKBOX_UUID=$(get_part_uuid "$(dev_part "${OSD_DEVICE}" 5)")
-    JOURNAL_PART=$(ceph-disk list "${OSD_DEVICE}" | awk '/ceph journal/ {print $1}') # This is a privileged container so 'ceph-disk list' works
+    JOURNAL_PART=$(ceph-disk list "${OSD_DEVICE}" | grep journal | sed -r 's/^.*\s([^ ]+)$/\1/') # This is a privileged container so 'ceph-disk list' works.
```
leseb: 👎  when using dmcrypt the output is different... so this command is correct;

guits: let's use json output then

leseb: or `awk`? :p

alfredodeza:
At some point it would be useful to start porting these to something that can be tested or re-used like in Python. For example, in this particular case, ceph-deploy has already done this work.

Specifically on this case, it detects correctly the journal if present:

https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/osd.py#L521

This merged solution is OK. But looking forward, the broad use of shell scripting needs to be reduced and ported to Python so that it can:

* be unit tested
* get re-used
* improve  readability

PR #795, opened 2017-09-26
Add openSUSE 42.3 flavor for luminous

BlaineEXE:
I'd like to have the option to build openSUSE flavor containers for Luminous. I've tested that this gets a basic cluster up and running successfully.

### DIFF
--- a/ceph-releases/luminous/opensuse/42.3/daemon/Dockerfile
+++ b/ceph-releases/luminous/opensuse/42.3/daemon/Dockerfile
@@ -1,0 +1,63 @@
+# CEPH DAEMON IMAGE
+# CEPH VERSION: Luminous
+# CEPH VERSION DETAIL: 12.x.x
+
+FROM opensuse:42.3
+MAINTAINER Blaine Gardner "blaine.gardner@suse.com"
+
+# OS version string for repos
+ARG OS_VERSION='openSUSE_Leap_42.3'
+# Prevent upgrades to a newer major release
+ARG CEPH_VERSION_NUM_UPPERBOUND='13'
+ENV CEPH_VERSION 'luminous'
+ARG CONFD_VERSION='0.13.0'
+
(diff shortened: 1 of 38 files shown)


ceph-releases/luminous/opensuse/42.3/daemon/ganesha.repo
```diff
@@ -0,0 +1,5 @@
+[ganesha]
```
leseb: not a symlink?

BlaineEXE: Hmm ... I just copied the structure directly from the CentOS 7. I missed this. What is this file used for? It doesn't look like it gets added to the container in the Dockerfile. Perhaps this is unnecessary.

leseb: hum this is used here: https://github.com/ceph/ceph-docker/blob/43ee6f104119eaa309248501d4df31aeacdf4ae5/ceph-releases/luminous/centos/7/daemon/Dockerfile#L15

If it's not a symlink, this might mean there is only a package for centos, at least only rpm.
Is Suse rpm based?

BlaineEXE: It is rpm based, but I'd prefer to get ganesha builds from a SUSE repository since it should be tested alongside whatever the current Ceph version is. I'll have to do a little digging and update this. Good catch.

leseb: sure, let me know when you're ready.

BlaineEXE: Will do. Thanks!

BlaineEXE:
Okay. This should be ready to merge. I removed the `ganesha.repo` file and added NFS Ganesha components relevant for Ceph.

PR #798, opened 2017-09-27
daemon: do not run umount lockbox

leseb:
Even if it runs with || true and won't fail this is really confusing for
users.


leseb:

PR #813, opened 2017-10-11
Fix generate-dev-env returning failure on success

BlaineEXE:
The generate-dev-env.sh script should not return failure on the final
`popd` if a preceding `pushd` was never executed.

### DIFF
--- a/generate-dev-env.sh
+++ b/generate-dev-env.sh
@@ -74,4 +74,4 @@ case "$1" in
       esac
     ;;
 esac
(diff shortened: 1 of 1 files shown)


BlaineEXE:
I'm using the `generate-dev-env.sh` script in some local automation and don't want it to return a false failure.

BlaineEXE:
I'm using the `generate-dev-env.sh` script in some local automation and don't want it to return a false failure.

PR #792, opened 2017-09-25
do not use hard coded label selector in kubectl

rootfs:
- not use hard coded label selector in kubectl
- don't auth add if keyring is missing
- init luminous osd directory

PR #819, opened 2017-10-17
luminous: Use the admin keyring to add OSD keys

adamwg:
In luminous the bootstrap-osd profile is not allowed to add keys. Use the admin keyring to add OSD keys, allowing OSDs to start.


liewegas:
the fix is to use the new 'osd new' command, which does all teh key and other creaetion stuff in one go.

liewegas:
(although i don't know what this scenario is trying to do in general; maybe client.admin is supposed to be there and that is fine too.)

leseb:
@liewegas when using `osd new` is the admin key needed or is `client.bootstrap-osd` sufficient?

liewegas:
'osd new' is the *only* command the new bootstrap-osd can do. :)

alfredodeza:
This is how ceph-volume does it: https://github.com/ceph/ceph/blob/master/src/ceph-volume/ceph_volume/util/prepare.py#L37-L57

adamwg:
Looks like the switch to `osd new` was made in #792. I'll close this PR and see if the latest is working for me.

Thanks all for the discussion.

PR #823, opened 2017-10-20
Fix mon id where name in monmap doesn't match monitor name

JrCs:
This fix change the mon id in monmap in case the monitor name in the monmap doesn't match the hostname of the monitor. 
This can be the case when using k8s to create the monmap from the pods labels and the hostname of the monitors are full qualified while the hostname of the servers is not full qualified.


Raboo:
Can you make it so it works with jewel as well?

leseb:
@JrCs, @Raboo please reach an agreement when it comes to who/how this should be fixed but please for both Jewel and Luminous :). Thanks

JrCs:
Backport to jewel too.

Raboo:
Thanks @JrCs

PR #831, opened 2017-11-02
Gracefully unmount osd

BlaineEXE:
I revisited PR #694, still starting from @leseb's `exit.sh` script and converted it from an exit hook to a script called `osd-exit.sh` which can be called before destroying an OSD container. I added code to the kubernetes example to start testing it. I'd like to do a couple more tests, but my first was successful.


BlaineEXE:
I believe @liewegas, @alram, and @dmick might have interest in looking at this.

BlaineEXE:
Per comment https://github.com/ceph/ceph-container/pull/694#issuecomment-341528656, these changes are unnecessary. Cosing this PR.

PR #837, opened 2017-11-07
Fix bad selection of monitor(s) from k8s

JrCs:


ceph-releases/jewel/ubuntu/14.04/daemon/config.k8s.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+
```
leseb: ?

JrCs: It is just for the readiness

JrCs: Do you want that i remove the empty line ?

leseb:
jenkins test this please

osixia:
Why did you remove the namespace ?!

JrCs:
Namespace is not necessary because the pods are start on the right namespace and inside the containers of the pods the default namespace is the namespace of the pods

osixia:
Thanks,
but should **check_zombie_mons.py** use **K8S_MON_SELECTOR** variable too ?

JrCs:
@ghost you're right.
Add a new PR #860.

PR #840, opened 2017-11-14
daemon: remove OSD_FORCE_ZAP option

leseb:
It is too dangerous to have this kind of capability, especially if there
is automation in place that uses this option.
Prior to deploy an OSD you must make sure the device is in a proper
shape. Also, ceph-disk prepare will create a GPT label if there is no
label on the drive.


leseb:
jenkins test ceph_ansible-jewel-xenial-filestore_osds_container

leseb:
jenkins test ceph_ansible-luminous-centos7-bluestore_osds_container

PR #843, opened 2017-11-15
luminous: Don't provide a keyring argument to rbd-mirror

adamwg:
rbd-mirror needs to use multiple keyrings to connect to multiple
clusters. Passing the -k argument prevents it from doing so. It finds
the correct keyring on its own via the -n argument, so the -k is
unnecessary anyhow.


fitschn:
+1

PR #859, opened 2017-12-07
Enable Ceph Manager restful module port for the Ubuntu 16.04 demo

p-na:
This is just a suggestion, so that one doesn't need to find out the IP of the Docker container to be able to access the API of the Ceph Manager.

Signed-off-by: Patrick Nawracay <pnawracay@suse.com>


leseb:
@p-na do you like this script? Because I'm about to kill it in https://github.com/ceph/ceph-container/pull/862
cn is the replacement for this but it does not (and won't likely) expose ceph-mgr modules.

Is that an issue for you?

p-na:
I've found out that it's the old API (on port 5000). Therefore I don't need this change.

PR #863, opened 2017-12-11
zap: fix issue on AWS environment

guits:
devices like : `/dev/xvda` don't match the current regexp, therefore the
purge fails on AWS environment.


alfredodeza:
This is going to keep happening because it is incredibly difficult to correctly match any/all combinations for device names. A regular expression is not the right way to solve this, and we will continue to match devices that aren't correct (false positives) or correct devices that aren't matched at all (false negatives).

In ceph-volume we use `lsblk` and `blkid` to corroborate input for devices and partitions. Please consider an alternative solution to detect/match devices vs. trying to fix this (and other) regular expressions

PR #877, opened 2018-01-09
Cleaning container

ErwanAliasr1:
The actual container is running a pretty classic ubuntu container full of useless stuff.
As a result the container is pretty big on the filesystem.

This patch aims at gaining disk's space by using the following techniques:
- stripping raw binaries/libraries
- removing useless data files like documentation
- removing useless binaries
- removing useless packages like perl
- removing testing code/binaries
- removing pre-compiled python files as we nearly don't use them
- compressing debugging tools
- clearing log files

Note the stripping process is done in two passes :
- first one targets "easy" stuff
- second one targets more tricky such that may be discussed if some
  issue occurs

A typical output looks like the following :
  Step 18 : RUN bash "/clean_container.sh"
   ---> Running in 6a48dec89816
  Stripping process: Step 1 removed 137MB
  [...]
  Stripping process: Step 2 removed 212MB
  [...]
  Stripping process saved 349MB and dropped container size from 851MB to 502MB

### DIFF
--- a/ceph-releases/luminous/centos/7/daemon/Dockerfile
+++ b/ceph-releases/luminous/centos/7/daemon/Dockerfile
@@ -8,29 +8,32 @@ MAINTAINER Sébastien Han "seb@redhat.com"
 ENV CEPH_VERSION luminous
 ENV CONFD_VERSION 0.10.0
 
-# Install Ceph
-RUN rpm --import 'https://download.ceph.com/keys/release.asc'
-RUN rpm -Uvh http://download.ceph.com/rpm-${CEPH_VERSION}/el7/noarch/ceph-release-1-1.el7.noarch.rpm
-RUN yum install -y epel-release
-ADD ganesha.repo /etc/yum.repos.d/
-RUN yum clean all
-ARG PACKAGES="unzip ceph-mon ceph-osd ceph-mds ceph-mgr ceph-base ceph-common ceph-radosgw rbd-mirror device-mapper sharutils etcd kubernetes-client e2fsprogs s3cmd wget nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw"
-ARG PURGES="/tmp/* /var/tmp/* /usr/lib/{dracut,locale,systemd,udev} /usr/bin/hyperkube /usr/bin/etcd /usr/bin/systemd-analyze /etc/{udev,selinux} /usr/lib/{udev,systemd}"
-RUN yum install -y $PACKAGES && rpm -q $PACKAGES && yum clean all && rm -rf $PURGES
-
-# Install confd
-ADD https://github.com/kelseyhightower/confd/releases/download/v${CONFD_VERSION}/confd-${CONFD_VERSION}-linux-amd64 /usr/local/bin/confd
-RUN chmod +x /usr/local/bin/confd && mkdir -p /etc/confd/conf.d && mkdir -p /etc/confd/templates
-
-# Download forego
-RUN wget -O /forego.tgz 'https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-amd64.tgz' && \
-  cd /usr/local/bin && tar xfz /forego.tgz && chmod +x /usr/local/bin/forego && rm /forego.tgz
-
-# Add s3cfg file
 ADD s3cfg /root/.s3cfg
 
 # Add bootstrap script, ceph defaults key/values for KV store
 ADD *.sh ceph.defaults check_zombie_mons.py ./osd_scenarios/* entrypoint.sh.in disabled_scenario /
+
+ARG PACKAGES="unzip ceph-mon ceph-osd ceph-mds ceph-mgr ceph-base ceph-common ceph-radosgw rbd-mirror device-mapper sharutils etcd kubernetes-client e2fsprogs s3cmd wget nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw"
+
+# Install Ceph
+ADD ganesha.repo /etc/yum.repos.d/
+
+RUN rpm --import 'https://download.ceph.com/keys/release.asc' && \
+  rpm -Uvh http://download.ceph.com/rpm-${CEPH_VERSION}/el7/noarch/ceph-release-1-1.el7.noarch.rpm && \
+  yum install -y epel-release && \
+  yum clean all && \
+  yum install -y $PACKAGES && \
+  rpm -q $PACKAGES && \
+  yum clean all && \
+  \
+  # Install confd
+  wget -O /usr/local/bin/confd "https://github.com/kelseyhightower/confd/releases/download/v${CONFD_VERSION}/confd-${CONFD_VERSION}-linux-amd64" && \
+  chmod +x /usr/local/bin/confd && mkdir -p /etc/confd/conf.d && mkdir -p /etc/confd/templates && \
+  \
+  # Download forego
+  wget -O /forego.tgz 'https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-amd64.tgz' && \
+  cd /usr/local/bin && tar xfz /forego.tgz && chmod +x /usr/local/bin/forego && rm /forego.tgz && \
+  bash /clean_container.sh && rm /clean_container.sh
 
 # Modify the entrypoint
 RUN bash "/generate_entrypoint.sh" && \
--- a/ceph-releases/luminous/centos/7/daemon/clean_container.sh
+++ b/ceph-releases/luminous/centos/7/daemon/clean_container.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+INITIAL_SIZE="$(du -sm / 2>/dev/null | awk '{ print $1 }')"
+DEBUG=
+
+PURGES="/tmp/* /var/tmp/* /usr/lib/{dracut,locale,systemd,udev} /usr/bin/hyperkube /usr/bin/etcd /usr/bin/systemd-analyze /etc/{udev,selinux} /usr/lib/{udev,systemd}"
+PURGES="$PURGES /usr/share/hwdata/{iab.txt,oui.txt}"
+# As we need to keep every PURGES as an arguement to rm, let's ignore SC2086
+# shellcheck disable=SC2086
+rm -rf $PURGES
+
+PURGE_PKGS="groff-base e2fsprogs sharutils shadow-utils yum-utils python-gobject-base python-kitchen passwd bind-license policycoreutils-python epel-release"
+PURGE_PKGS="$PURGE_PKGS libselinux-python libss libcgroup dbus-python yum-plugin-ovl rbd-mirror libxml2-python python-IPy checkpolicy libsemanage-python setools-libs"
+PURGE_PKGS="$PURGE_PKGS systemd-sysv dbus-glib gobject-introspection audit-libs-python"
+
+# Step 1
+# Let's remove easy stuff
+strip -s /usr/local/bin/{confd,forego}
+
+# Compressing those very big binaries
+# As we don't run them often, the peformance penalty isn't huge
+for binary in /usr/bin/{etcdctl,kubectl} /usr/local/bin/{forego,confd}; do
+  gzexe $binary && rm ${binary}~
+done
+
+rm -rf /usr/share/{doc,info,locale,man}/*
+
+# Some logfiles are not empty, there is no need to keep them
+find /var/log/ -type f -exec truncate -s 0 {} \;
+
+STEP1_SIZE=$(du -sm / 2>/dev/null | awk '{print $1}')
+REMOVED_STEP1_SIZE=$((INITIAL_SIZE - STEP1_SIZE))
+echo "Stripping process: Step 1 removed ${REMOVED_STEP1_SIZE}MB"
+
+# Step 2
+# Let's remove stuff than can be discussed
+
+rm -f /usr/lib/ceph/mgr/dashboard/static/AdminLTE-*/plugins/datatables/extensions/TableTools/images/psd/* # Photoshop files inside a container ?
+
+# Let's remove all the pre-compiled python files, if needed they will be rebuilt
+find  / -xdev -name "*.pyc" -exec rm {} \; # 23MB of pyc
+
+# ceph-dencoder is only used for debugging, compressing it saves 10MB
+# If needed it will be decompressed
+xz /usr/bin/ceph-dencoder
+
+# Timezone is not configured so let's remove the zoneinfo (8MB)
+rpm -e tzdata --nodeps
+
+# Removing perl as we don't need it
+# perl is required by libibverbs (Infiniband) which is required by ceph in general
+# the ceph daemons are linked against the lib but we don't care about perl itself
+#	perl(Getopt::Long) is needed by (installed) libibverbs-13-7.el7.x86_64
+#	perl(File::Basename) is needed by (installed) libibverbs-13-7.el7.x86_64
+#	perl(strict) is needed by (installed) libibverbs-13-7.el7.x86_64
+#	perl(warnings) is needed by (installed) libibverbs-13-7.el7.x86_64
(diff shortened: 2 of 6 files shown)


leseb:
@bassam fyi

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
```diff
@@ -0,0 +1,58 @@
...
+
+STEP1_SIZE=$(du -sm / 2>/dev/null | awk {'print $1}')
+REMOVED_STEP1_SIZE=$(($INITIAL_SIZE - $STEP1_SIZE))
+echo "Stripping process: Step 1 removed ${REMOVED_STEP1_SIZE}MB"
+
+# Step 2
+# Let's remove stuff than can be discussed
+
+# Let's strip the ceph libraries
+strip -s /usr/lib/ceph/erasure-code/*
```
leseb: @liewegas thoughts?

leseb: @liewegas thoughts?

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
```diff
@@ -0,0 +1,58 @@
...
+strip -s /usr/lib/rados-classes/*
+strip -s /usr/lib/python2.7/dist-packages/{rados,rbd,rgw}.x86_64-linux-gnu.so
+
+rm -f /usr/lib/ceph/mgr/dashboard/static/AdminLTE-*/plugins/datatables/extensions/TableTools/images/psd/* # Photoshop files inside a container ?
```
leseb: @jcsp, is it something we can remove in ceph?

ceph-releases/luminous/ubuntu/16.04/daemon/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y wget unz
     wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add - && \
     echo "deb http://download.ceph.com/debian-$CEPH_VERSION/ xenial main" | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list && \
     echo "deb http://download.ceph.com/nfs-ganesha/deb-V2.5-stable/luminous/ xenial main" | tee /etc/apt/sources.list.d/nfs-ganesha.list && \
-    apt-get update && apt-get install -y --force-yes $PACKAGES && \
+    apt-get update && apt-get install -y  --no-install-recommends --force-yes $PACKAGES && \
     dpkg -s $PACKAGES && \
     apt-get clean && rm -rf $PURGES && \
```
leseb: can you include this part in your clean up script?

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
liewegas: I think this should be dropped and that this is the real problem: https://github.com/ceph/ceph/pull/19870

ceph-releases/luminous/ubuntu/16.04/daemon/Dockerfile
ErwanAliasr1: I don't think that's a good idea since that would make that layer bigger than expected

leseb: then let's put everything in one `RUN` statement :)

ErwanAliasr1: yes !

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
ErwanAliasr1: Thanks for taking it upstream. I keep this change as it could be useful on existing & older releases and it really save space. Once your patch will be in a release, that line will be useless but without any impact.

liewegas: We'll be able to backport the packaging change to luminous, which means this is temporary.. I'd lean towards avoiding adding it in the first place.

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+INITIAL_SIZE=$(du -sm / 2>/dev/null | awk {'print $1}')
+DEBUG=
+
+# Step 1
+# Let's remove easy stuff
+strip -s /usr/local/bin/{confd,forego,kubectl}
+strip -s /usr/bin/{crushtool,monmaptool,osdmaptool}
```
ktdreyer: I'm surprised `/usr/bin/{crushtool,monmaptool,osdmaptool}` is not stripped already. What OS / Ceph version were you using where this makes a difference?

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# ceph-dencoder is only used for debugging, compressing it saves 10MB
+# If needed it will be decompressed
+gzip -9 /usr/bin/ceph-dencoder
+
+# Timezone is not configured so let's remove the zoneinfo (8MB)
+dpkg --purge --force-all tzdata
+
+# Removing perl as we don't need it
+apt-get remove -y --auto-remove perl
```
ktdreyer: What is pulling in `perl` to begin with? This would be great to document the reason in the comments.

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
```diff
@@ -0,0 +1,58 @@
...
+# Let's remove stuff than can be discussed
+
+# Let's strip the ceph libraries
+strip -s /usr/lib/ceph/erasure-code/*
+strip -s /usr/lib/rados-classes/*
+strip -s /usr/lib/python2.7/dist-packages/{rados,rbd,rgw}.x86_64-linux-gnu.so
+
+rm -f /usr/lib/ceph/mgr/dashboard/static/AdminLTE-*/plugins/datatables/extensions/TableTools/images/psd/* # Photoshop files inside a container ?
+
+# Let's remove all the pre-compiled python files, if needed they we'll be rebuilt
```
ktdreyer: `they we'll` -> `they will`

This means we're rebuilding the python-rados files any time we run `/usr/bin/ceph`, right?

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
ErwanAliasr1: System base is pulling it

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
ErwanAliasr1: Get:131 http://download.ceph.com/debian-luminous xenial/main amd64 ceph-osd amd64 12.2.2-1xenial [12.9 MB]
Get:132 http://download.ceph.com/nfs-ganesha/deb-V2.5-stable/luminous xenial/main amd64 nfs-ganesha amd64 2.5.4-0-g9e93d73-1xenial [573 kB]

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
ErwanAliasr1: Only the first time we'll call the python files.

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
ErwanAliasr1: Added a note for that. thx.

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
ErwanAliasr1: very nice @liewegas

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
ktdreyer: I wonder if it's something else that pulls in "perl"?

Running `apt-get remove -y --auto-remove perl` in a vanilla `ubuntu:xenial` container says the package is not installed. But perl-base is installed, and many things like "apt" depend on it.

ErwanAliasr1: Look at my comment for perl-base at line 45.

ceph-releases/luminous/centos/7/daemon/Dockerfile
```diff
@@ -8,29 +8,32 @@ MAINTAINER Sébastien Han "seb@redhat.com"
...
-ADD s3cfg /root/.s3cfg
-
-# Add bootstrap script, ceph defaults key/values for KV store
-ADD *.sh ceph.defaults check_zombie_mons.py ./osd_scenarios/* entrypoint.sh.in disabled_scenario /
+wget -O /forego.tgz 'https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-amd64.tgz' && \
+cd /usr/local/bin && tar xfz /forego.tgz && chmod +x /usr/local/bin/forego && rm /forego.tgz && \
+bash /clean_container.sh && rm /clean_container.sh
 
 # Modify the entrypoint
 RUN bash "/generate_entrypoint.sh" && \
```
leseb: can we merge this RUN with the previous one as well?

ceph-releases/luminous/centos/7/daemon/clean_container.sh
```diff
@@ -0,0 +1,76 @@
...
+PURGE_PKGS="$PURGE_PKGS systemd-sysv dbus-glib gobject-introspection audit-libs-python"
+
+# Step 1
+# Let's remove easy stuff
+strip -s /usr/local/bin/{confd,forego}
+
+# Compressing those very big binaries
+# As we don't run them often, the peformance penalty isn't huge
+for binary in /usr/bin/{etcdctl,kubectl} /usr/local/bin/{forego,confd}; do
+   gzexe $binary && rm ${binary}~
```
leseb: indent 2

ceph-releases/luminous/ubuntu/16.04/daemon/Dockerfile
```diff
@@ -23,26 +25,24 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y wget unz
...
-# Add bootstrap script, ceph defaults key/values for KV store
-ADD *.sh ceph.defaults check_zombie_mons.py ./osd_scenarios/* entrypoint.sh.in disabled_scenario /
+   wget -O /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" && \
+   chmod +x /usr/local/bin/kubectl && \
+\
+# Cleaning container
+   bash /clean_container.sh && rm /clean_container.sh
 
 # Modify the entrypoint
 RUN bash "/generate_entrypoint.sh" && \
```
leseb: ditto

ceph-releases/luminous/ubuntu/16.04/daemon/clean_container.sh
leseb: corresponding PR: https://github.com/ceph/ceph/pull/19925

ceph-releases/luminous/centos/7/daemon/Dockerfile
ErwanAliasr1: It doesn't provides much gain. The main intent was to save space.

I do prefer keeping that entrypoint stuff splitted.

ceph-releases/luminous/centos/7/daemon/clean_container.sh
ErwanAliasr1: I fixed all the indent 2 issues and refactored the Dockerfile to get that very long RUN entry much more readable.

bassam:
It would be great if the cleaning/minification of the base OS was kept separate from Ceph itself. The Dockerfile and scripts could benefit from this factoring -- OS minified --> Ceph --> Others (Ceph-Container, Rook etc.)

Is `strip -s` going to prevent good backtraces? Also `strip` on golang binaries is not recommended or supported, I've seen it work too.

ceph-releases/luminous/ubuntu/16.04/daemon/Dockerfile
```diff
@@ -9,41 +9,36 @@ ENV CEPH_VERSION luminous
...
+  wget -O /usr/local/bin/confd "https://github.com/kelseyhightower/confd/releases/download/v${CONFD_VERSION}/confd-${CONFD_VERSION}-linux-amd64" && \
+  chmod +x /usr/local/bin/confd && mkdir -p /etc/confd/conf.d && mkdir -p /etc/confd/templates && \
+  \ # Install forego
+  wget -O /forego.tgz 'https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-amd64.tgz' && \
+  cd /usr/local/bin && tar xfz /forego.tgz && chmod +x /usr/local/bin/forego && rm /forego.tgz && \
+  \ # Install kubectl
+  wget -O /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" && \
+  chmod +x /usr/local/bin/kubectl && \
+  \ # Cleaning container
+  bash /clean_container.sh && rm /clean_container.sh
```
bassam: Running `clean_container` here does not remove some of unneeded files from the base Ubuntu image. The base image is plain Ubuntu:16.04 and its going to bring in ~100MB of files even if you end up removing them in this step.

Rook solved this by slimming the base ubuntu image and exporting it to a base tarball and back in as a new base. See https://github.com/rook/rook/blob/master/images/base/Makefile. New docker daemon now support multi-stage builds which make this work much easier but its widely deployed yet.

bassam:
This work does seem duplicative with Rook's base images. I captured some discussion in https://github.com/rook/rook/issues/1404.

ceph-releases/luminous/ubuntu/16.04/daemon/Dockerfile
leseb: @bassam, just a FYI @ErwanAliasr1 has a commit for that. We tested it together, pushing it soon.

ErwanAliasr1:
Poke @leseb I'm ready

ceph-releases/luminous/centos/7/daemon/Dockerfile
```diff
@@ -8,30 +8,33 @@ MAINTAINER Sébastien Han "seb@redhat.com"
...
+  wget -O /usr/local/bin/confd "https://github.com/kelseyhightower/confd/releases/download/v${CONFD_VERSION}/confd-${CONFD_VERSION}-linux-amd64" && \
+  chmod +x /usr/local/bin/confd && mkdir -p /etc/confd/conf.d && mkdir -p /etc/confd/templates && \
+  \
+  # Download forego
+  wget -O /forego.tgz 'https://bin.equinox.io/c/ekMN3bCZFUn/forego-stable-linux-amd64.tgz' && \
+  cd /usr/local/bin && tar xfz /forego.tgz && chmod +x /usr/local/bin/forego && rm /forego.tgz && \
+  bash /clean_container.sh && rm /clean_container.sh
+
 # Modify the entrypoint
 RUN bash "/generate_entrypoint.sh" && \
```
leseb: can't we include this  `RUN` in the previous `RUN`?

ErwanAliasr1: No, there is no particular gain in space and for the semantic, I do prefer keeping the install/clean in a single run and the add of the entrypoint in a separate one.

Issue #884, opened 2018-01-24
ceph osd: Cannot use NVMe device as an OSD when running Ceph in Containers.


Deepthidharwar:
Using the latest ceph-ansible, to deploy Ceph in Containers. Ceph cluster comprises of 1 MON, 1 MGR and 1 OSD - NVMe. Currently when we try to deploy the same, the OSD container keeps restarting as it fails to execute entrypoint.sh disk_list.sh.

When we use NVMe, the partitions are /dev/nvme0n1p1 and /dev/nvme0n1p2. 
Looks like it is appending 1 to the OSD_DEVICE when trying to mount ceph-data,  so it is searching for /dev/nvme0n11 instead.

The device is prepared properly but when activating the disk, it fails.
There are no errors on the ceph-ansible side as it has deployed the container.

Manually running the script:

# /usr/share/ceph-osd-run.sh nvme0n1
mount: special device /dev/nvme0n11 does not exist
Error response from daemon: No such container: expose_partitions_nvme0n1
2018-01-24 03:35:22  /entrypoint.sh: static: does not generate config
mount: special device /dev/nvme0n11 does not exist

disk-list.sh 

function mount_ceph_data () {
  if is_dmcrypt; then
	mount /dev/mapper/"${data_uuid}" "$tmp_dir"
  else
	if is_loop_dev "${OSD_DEVICE}"; then
	  mount "${OSD_DEVICE}p1" "$tmp_dir"
	else
	  mount "${OSD_DEVICE}1" "$tmp_dir"
	fi
  fi

As seen above, the check is only for loop-dev and not for NVMe.  So it fails to mount when using NVMe OSD device.

alfredodeza:
The referenced PR (#885 ) doesn't seem to address some of the issues described here:

* A '1' is appended to the nvme device
* There are no errors reported when this fails

The referenced PR is addressing detection of dmcrypt devices which this issue doesn't even mention.

guits:
the error reported about the '1' appended was because of [this](https://github.com/ceph/ceph-container/pull/885/files#diff-46e2ba49b269379fcfc424702c34ac82L47)
there is no need to rely on `is_loop_dev` which was by the way used by `is_dmcrypt` since i  rely on `blkid` to find and mount the ceph data partition.
The main goal of the referrenced PR was to drop the `is_loop_dev` function and not detection of dmcrypt devices.

PR #885, opened 2018-01-24
disk_list.sh: improve 'ceph data' mount partition

guits:
We can assume the device is dmcrypt as soon as we find `TYPE=crypto_LUKS`
on `${OSD_DEVICE}`.
There is no need to try to predict what will be the name of the 'ceph data'
partition, therefore, `is_loop_dev` is no longer needed.
Let's find the `ceph data` uuid partition with `blkid` and mount it.

Fixes: #884


alfredodeza:
This is not the correct approach, because "crypto_LUKS" just gives *one* of the two ways that Ceph supports encryption. One of them is LUKS, and the other one is PLAIN. This method will not work with PLAIN encryption.

PR #886, opened 2018-01-29
Add sudo and lvm packages for ceph-volume

wedged:
ceph-volume requires sudo, and supports LVM.  This change adds those packages to the image so that we can make progress toward deprecating ceph-disk.

### DIFF
--- a/ceph-releases/luminous/ubuntu/16.04/daemon/Dockerfile
+++ b/ceph-releases/luminous/ubuntu/16.04/daemon/Dockerfile
@@ -10,7 +10,7 @@ ENV CONFD_VERSION 0.10.0
 ENV KUBECTL_VERSION v1.6.0
 
 # Packages list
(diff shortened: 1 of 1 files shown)


alfredodeza:
This was fixed in http://tracker.ceph.com/issues/22282 and got backported to Luminous. We are just waiting on a new Ceph release, which unfortunately, means a few more months. Maybe April 2018?

alfredodeza:
The `sudo` issue got fixed, but `lvm2` will still be required. It is now required in Ceph, but not sure how this project handles that.

leseb:
I'm fine having `lvm2` since we will need it for ceph-volume.

leseb:
@wedged please squash your commits.

wedged:
done, sorry 'bout that.  my git etiquette is poor, as you can see.

PR #891, opened 2018-01-30
[DNM] Updating the build process to use rook's makefile

ErwanAliasr1:
The idea is too remove the actual generate-dev-env.sh by a makefile taking care of dependencies & paralellism.


leseb:
Please rebase on top of https://github.com/ceph/ceph-container/pull/886
Thanks.

ErwanAliasr1:
@leseb done

bassam:
This is a great step forward!

I believe there is more room for simplification here, specifically, I think you can reduce the level of symlinking and duplication further.

What if the Dockerfiles/makefiles for daemon, daemon-base, etc. were not specific to a Ceph release? Instead, the Makefile knows how to copy ceph release specific scripts into the image being built. For example,

```
ceph-releases/
   luminous/
        <scripts and entrypoints>. No dockerfiles
images
   base
   daemon-base
   daemon
        Makefile (this copies the scripts from ceph-releases/ as appropriate while building the image)
```

I think with this approach you wouldn't need any symlinking.

Also you can bring `base` back into the images directory.

ceph-releases/jewel/centos/7.4.1708/daemon-base/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# CEPH VERSION: Jewel
+# CEPH VERSION DETAIL: 10.2.x
+
+FROM BASEIMAGE
+MAINTAINER Sébastien Han "seb@redhat.com"
+
+ENV CEPH_VERSION jewel
+
+# Install Ceph
+ARG PACKAGES="unzip ceph-mon ceph-osd ceph-mds ceph-base ceph-common ceph-radosgw rbd-mirror device-mapper sharutils etcd kubernetes-client e2fsprogs wget"
```
bassam: daemon base should not include install wget unzip etc.

ceph-releases/kraken/ubuntu/16.04/daemon-base/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+FROM BASEIMAGE
+MAINTAINER Sébastien Han "seb@redhat.com"
+
+ENV CEPH_VERSION kraken
+
+ADD clean_container.sh /
+
+# Packages list
+ARG PACKAGES="ceph-mon ceph-osd ceph-mds ceph-mgr ceph-base ceph-common radosgw rbd-mirror sharutils etcd"
```
bassam: daemon-base does not need etd.

ceph-releases/luminous/ubuntu/16.04/daemon-base/clean_container.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Let's remove all the pre-compiled python files, if needed they will be rebuilt
+find  / -xdev -name "*.pyc" -exec rm {} \; # 23MB of pyc
+
+# ceph-dencoder is only used for debugging, compressing it saves 10MB
+# If needed it will be decompressed
+gzip -9 /usr/bin/ceph-dencoder
+
+# Timezone is not configured so let's remove the zoneinfo (8MB)
+dpkg --purge --force-all tzdata
```
bassam: should this happen in base cleanup?

ceph-releases/luminous/ubuntu/16.04/daemon-base/clean_container.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Removing perl as we don't need it
```
bassam: not sure this is worth it since most deb tools rely on perl. Can it be moved to base?

ceph-releases/luminous/ubuntu/16.04/daemon-base/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# install prerequisites
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y wget unzip uuid-runtime python-setuptools udev dmsetup && \
+  # Install ceph, ganesha and etcd \
+  wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add - && \
+  echo "deb http://download.ceph.com/debian-$CEPH_VERSION/ xenial main" | tee /etc/apt/sources.list.d/ceph-$CEPH_VERSION.list && \
+  echo "deb http://download.ceph.com/nfs-ganesha/deb-V2.5-stable/luminous/ xenial main" | tee /etc/apt/sources.list.d/nfs-ganesha.list && \
+  apt-get update && apt-get install -y  --no-install-recommends --force-yes $PACKAGES && \
+  dpkg -s $PACKAGES && \
+  apt-get clean && \
+  bash /clean_container.sh && rm -f /clean_container.sh
```
bassam: ideally this cleanup here would be specific to ceph, and not the whole OS image. OS specific clean should go in base.

bassam: nit: I find the inline cleanup more readable than another script: https://github.com/ceph/ceph-container/blob/master/images/daemon-base/Dockerfile.luminous-ubuntu#L46

ErwanAliasr1:
@bassam I tried this approach of having the daemon-base into the ceph-release directory as we don't have the same set of packages to install across the various ceph releases.

I agree the symlinking is a pain.

ceph-releases/jewel/centos/7.4.1708/daemon-base/Dockerfile
ErwanAliasr1: ack

ceph-releases/kraken/ubuntu/16.04/daemon-base/Dockerfile
ErwanAliasr1: ack

ceph-releases/luminous/ubuntu/16.04/daemon-base/Dockerfile
ErwanAliasr1: Agreed about the specific cleanup to ceph, I have to adjust that.
Disagree about having an inline cleanup, that could be too long & noisy.

ceph-releases/luminous/ubuntu/16.04/daemon-base/clean_container.sh
ErwanAliasr1: Neither the gzip of ceph-dencoder ;)
I've been reusing the same cleanup as per my initial PR, this have to be adjusted.

BlaineEXE:
I have concerns about base container squashing. Having smaller base containers is great, but squashing in this way obfuscates where the container comes from, and for sysadmins this is user-unfriendly. (Source: My roommate is a kubernetes sysadmin) Additionally, for any users who want to vet containers with security audits, an unknown-os base image from a tarball is a risk, and it would need its own vetting process. There is no guarantee that one of the binaries hasn't been modified malicously. I don't think saving 40MB of disk space is worth the user/admin troubles and security concerns this causes.

Squashing the base image also means that if multiple other containers are using the same base image, which is a likely occurrence, squashing the base image runs the risk of increasing the total disk usage on hosts.

The more important things we should be doing to make sure the container images are as small as possible is to avoid modifying the same file in multiple layers. The current ubuntu and rhel cleanup is done in the same layer as package installation, for example, which is a great way to do this.

ErwanAliasr1:
@BlaineEXE About base, that's a good argument.

BlaineEXE:
@ErwanAliasr1 Where is this branch hosted? I can't seem to find it to play around with the new format.

ErwanAliasr1:
@BlaineEXE Search for evelu-makefile branch. I do not guarantee a full stability.

bassam:
> Additionally, for any users who want to vet containers with security audits, an unknown-os base image from a tarball is a risk, and it would need its own vetting process. 

@BlaineEXE I'm not following why its a security risk to squash a container. My understanding is that security vetting tools should still be able to check for vulnerabilities etc.

> The current ubuntu and rhel cleanup is done in the same layer as package installation, for example, which is a great way to do this.

this is not a space saving since the base image would still contain the removed files, right?

BlaineEXE:
@bassam 

As to image size, I'm not referring to the raw size of the image built here. I'm referring to inter-image deduplication by image sharing. Consider the following. Sizes are rough.

Currently:
```
base (based on ubuntu:xenial but squashed) 70MB base
daemon-base (based on base) 80MB layer
daemon (based on daemon-base) 20MB layer
rook (based on daemon-base) 40MB layer
user-application (based on ubuntu:xenial) - 20MB
ubuntu:xenial - 110MB
```
total: 340MB

Proposed:
```
daemon-base (based on ubuntu:xenial) 80MB layer
daemon (based on daemon-base) 20MB layer
rook (based on daemon-base) 40MB layer
user-application (based on ubuntu:xenial) 20MB
ubuntu:xenial - 110MB
```
total: 270
In this case, the total disk usage is reduced because a user application also uses a common base to ceph-container/rook, saving 70MB on disk. The disk savings improve when the base layer gets updated also. Only the ubuntu:xenial layer needs updated instead of both the ubuntu:xenial layer AND the squashed base layer. Thus, I don't think squashing the base layer to save something like 40MB is worth breaking inter-image deduplication savings.


As to the security concern, I don't think it's a risk as much as it makes using this more annoying for sysadmins. The history of how a container is built is important to sysadmins. An obfuscated history is concerning, because the sysadmin doesn't know where the base came from when it shows up just as an arbitrary tarball. There could be compromised binaries that claim to be a particular thing but are actually malicious. It is easier to vet the image for security concerns if the history says "from some known and already vetted base" rather than "from tarball". If the sysadmins are security-minded, which they should be, they can no longer assume the base safe because they've already vetted it. They have to re-vet that base, which might be much more involved than simply running something like clair. In these respects, I don't think that making the sysadmin's life harder is worth saving about 40MB of disk space. I know things add up, but disk space is pretty cheap nowadays. Sysadmins are our end user, and their time is much more valuable than disk space.

BlaineEXE:
As a further aside to the space considerations, ubuntu minimal base images are getting progressively smaller and smaller, and as they continue to do so, the space savings gained by the export-import squash will become smaller and smaller.

http://blog.dustinkirkland.com/2018/02/rfc-ubuntu-1804-lts-minimal-images.html

bassam:
@BlaineEXE  +1 for using Bionic *instead* of our own base :-)

BlaineEXE:
@bassam I think that sounds like a good idea. That also leaves time to transition to using an unsquashed upstream base.

leseb:
@ErwanAliasr1 should we close this in favour of https://github.com/ceph/ceph-container/pull/899?

ErwanAliasr1:
Closing in favor of #905

PR #895, opened 2018-02-05
mon: use `uname -n` to set the mon name when starting

guits:
sometimes it's possible to face an issue with "wait for monitor socket
to exist" task in ceph-mon role according to the hostname configuration
on the host.

Prior this commit, the mon nodes used to create the socket name with the
`$HOSTNAME` variable which is defined with the content of `/etc/hostname` by
default whereas the failing check is using the ansible fact `{{
  ansible_hostname }}` which actually uses `uname -n` command.

We should force the mon to start using the result of `uname -n` to be
sure we will satisfy the check. This way, it means that if we use the
result of `uname -n` to start the mon and then we use the same command
to check whether the socket exists we ensure we won't have a mismatch.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1541303


guits:
@leseb comment addressed

PR #899, opened 2018-02-08
[DNM] First iteration of tree-pruning prototype

BlaineEXE:
I have been working with @ErwanAliasr1 on an idea for how to get rid of the symlink bloat and prune the trees down. The overall idea is to have a core set of code that is shared by all versions, and specific versions can override these files (or include extra files) individually. If the files are not overridden, the core/shared files are used.

This is a barebones prototype of the idea. Some changes needed (beyond
refactoring the code into the prototype format, if we decide to move
forward with the proto format) include (but are not limited to):
 - `__DOCKERFILE_BASE_IMAGE__` should be replaced by the makefile
 - With above, staging should also include the architecture
 - a makefile clean.all that cleans any/all build artifacts

P.S., A decision I made was to collect files for each container build into a staging directory. The makefile also has a `stage` target. The benefit of this is that services like quay.io and dockerhub can build files in staged directories directly. This has also been good for my development by verifying that the files being built are overridden correctly, and it should help other devs building containers as well when they can see what is output into staging from their attempted build.

## Iteration 2

- Rewrite stage.sh into python stage.py
- Allow `__VARIABLE__` replacements in any file
- Allow recursive `__VARIABLE__` replacements
- Build daemon-base and daemon
- Add all_ceph_releases/ directory
  TODO: This is pretty much synonymous with core. Do we want both?
- The luminous/ dir is now unnecessary, as files in all_ceph_releases/OS
  are able to provide the necessary differences. The code still supports
  `<specific ceph release>/` directories.
- Added basic doc to note documentation that needs updated

P.S., I included the Dockerfiles generated for daemon and daemon-base for both ubuntu and opensuse so reviewers can see what my code generates. These files wouldn't normally be versioned. All 4 containers build, but I haven't tested to see if they work properly.

## Iteration 3
Add ability to blacklist dirs/files

Files or directories can be blacklisted selectively by specifying the
build flavor environment variables that match what needs to be
blacklisted.

e.g., `path/to/be/blacklisted CEPH_VERSION=jewel BASEOS_REG=ubuntu`
      will blacklist the path for `jewel` builds on `ubuntu` bases

P.S., This update has several extraneous files that do nothing, but they exist to verify files are blacklisted appropriately. These would be removed for a master pull.

## Iteration 4
Erwan and I worked to improve the `stage.log` output from the staging step.


BlaineEXE:
Additional thought: I think we should consider using tini for ceph-container as Rook does.

ErwanAliasr1:
I'd like to use the branch directly, can I have the link to it please ?

ErwanAliasr1:
It would be nice to make 1 commit with your actual work and 1 commit for the importing the existing files. That would ease the review of your work. The copy does pollute the reading.
I had the similar issue on my PR.

ErwanAliasr1:
The Dockerfile in core/ seems not in an accurate please as one for the core/{base|daemon|daemon-base} should have one. Having one in core seems strange.

src/stage.sh
```diff
@@ -0,0 +1,72 @@
...
+  done
+}
+
+function stage_shallower_files () {
+  images_base_dir="${1%/}" # remove trailing '/'
+  # Then for all images that we're staging, copy the files up to the image
+  # base but not within the image base
+  image_staging_dirs="$(find "${STAGING_DIR}/" -maxdepth 1 -mindepth 1 -type d)"
+  override_dir="${images_base_dir}" # start with deepest dir
+  while [ ! -z "${override_dir}" ]; do
```
ErwanAliasr1: while [ -n 
"! -z" is "-n"

src/stage.sh
```diff
@@ -0,0 +1,72 @@
...
+    echo "Populating all staging images from ${override_dir}"
+    find -D exec "${override_dir}/" -depth -maxdepth 1 -type f | \
+      while read file; do copy_to_multiple "${file}" "${image_staging_dirs}"; done
```
ErwanAliasr1: This line triggers https://github.com/koalaman/shellcheck/wiki/SC2162.
Please fix or annotate to remove the warning.

BlaineEXE:
I wasn't as concerned with keeping good commits for a prototype that won't end up being merged into the project main. And I do indeed need to fix the Dockerfile location.

src/luminous/__DOCKERFILE_CEPH_VERSION__
```diff
@@ -0,0 +1 @@
+luminous
```
ErwanAliasr1: Can't we extract that information for the makefile variable ? That would avoid a new file.

ErwanAliasr1:
Would it be possible to get the modification of the rook's makefile to be splitted in another commit ? That would help tracking those changes.

src/stage.sh
BlaineEXE: Next version will by Python linted with flake8 and line length 99.

leseb:
@BlaineEXE is `src/logging` a leftover?

BlaineEXE:
@ErwanAliasr1 RE: logging, yes. I'm not sure what this file is. It may be an auto-generated python file.

ErwanAliasr1:
@BlaineEXE I improved the logging in this branch : https://github.com/ErwanAliasr1/ceph-docker/commits/evelu-poc-prune-tree

Can you integrate it ? That greatly improve the readability.

BlaineEXE:
@ErwanAliasr1 - I think your changes look good. I'll pull your commits in soon.

ErwanAliasr1:
Closing in favor of #905

PR #905, opened 2018-02-14
[skip ci] Transition to new project structure from POC #899

BlaineEXE:
I request feedback on https://github.com/ceph/ceph-container/pull/905#issuecomment-367172626

TODO List. This is a WIP PR, please do not comment about needing things that are already on this list unless you have specific improvements to suggest:
- [x] Documentation
  - [x] Updates to readme
  - [x] Python public methods have docstrings
  - [x] If the `__VAR__` file is supposed to contain actions that need done, a no-effect override needs to have the content `/bin/true` so that the `... && __VAR__ && ...` and-ing will proceed correctly.
- [x] Remove accidental modification to `examples/kubernetes/rgw_s3_client.py`
- [x] Pull Rook's Makefiles to show our modifications to them
- [x] Get rid of 'merge turds' in first commit
- [x] Cleanup of tzdata isn't working so well <-- commented in code as BUG
- [x] Clean up commits before merge
- [x] Rebase against master before merge
- [x] Retire old workflow scripts.


stage.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
+
+import os
+import logging
+import shutil
+import sys
+import time
+
+from stagelib.envglobals import *  # noqa: F403
```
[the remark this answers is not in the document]
ErwanAliasr1: from stagelib.envglobals import printGlobal, CEPH_VERSION, BASEOS_NAME, BASEOS_TAG, BASEOS_REG, ARCH, IMAGES_TO_BUILD, STAGING_DIR

This ^ change avoid many many warnings.

.gitignore
```diff
@@ -8,3 +8,8 @@ ceph.osd.keyring
 ceph.rgw.keyring
 client-admin-key
 ceph-releases/devel/*
+
+__pycache__
+*.pyc
+*.pyo
+*.log
```
ErwanAliasr1: +staging-*

stagelib/filecopytools.py
```diff
@@ -0,0 +1,59 @@
...
+def copy_files(filenames, src_path, dst_path, blacklist):
+    """
+    Copy a list of filenames from src to dst. Will overwrite existing files.
+    If any files are in the blacklist, they will not be copied.
+    If the src directory is in the blacklist, the dest path will not be created, and the files will
+      not be copied or processed.
+    """
+    # Adding a trailing "/" if needed to improve output coherency
+    dst_path = os.path.join(dst_path, '')
+    logging.debug('      {:<80}'.format(os.path.join(src_path, '')))
```
ErwanAliasr1: The actual 80 aligment makes an horrible alignment on the test files.

Makefile
```diff
@@ -0,0 +1,95 @@
...
+clean.image.%: daemon.%
+	rm -rf $(STAGING_DIR)
+
+stage: $(foreach p, $(FLAVORS_TO_BUILD), stage.$(p)) ;
+build: $(foreach p, $(FLAVORS_TO_BUILD), stage.$(p) do.image.$(p)) ;
+push: $(foreach p, $(FLAVORS_TO_BUILD), do.image.$(p)) ;
+clean: $(foreach p, $(FLAVORS_TO_BUILD), clean.image.$(p))
+
+stage-test:
+	DEBUG=1 tests/stage-test/test_staging.sh
```
ErwanAliasr1: Shall we enforce DEBUG for the stage-test ?

Makefile
```diff
@@ -0,0 +1,95 @@
...
+# [ceph release name]@[arch]@[base os name]@[base os registry]@[base os repo]@[base os tag]
+# e.g., luminous@amd64@ubuntu@_@ubuntu@16.04
+#       luminous@arm64@ubuntu@arm64v8@ubuntu@16.04
+
+#TODO: Staging should also stage the architecture, and the Dockerfile "FROM"
+#      replacement should be done from the baseos_reg/baseos_repo:baseos_tag
+#      rather than a __VARIABLE__ file.
+
+FLAVORS_TO_BUILD ?= \
+	luminous@amd64@ubuntu@_@ubuntu@16.04 \
```
ErwanAliasr1: Should be removed since this is not supported in that PR.

stagelib/filecopytools.py
BlaineEXE: Yeah. I noticed that.

Makefile
BlaineEXE: It isn't strictly necessary, but I think it makes sense for tests to have as verbose logging as possible.

Makefile
BlaineEXE: There is a lot in the Makefile that needs removed if it's not supported in this PR including all build/clean/push/daemon/...

stagelib/filecopytools.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    if not os.path.isdir(src_path):
+        return
+    for dirname, subdirs, files in os.walk(src_path, topdown=True):
+        # Remove src_path (and '/' immediately following) from our dirname
+        if os.path.join(dirname, '') in blacklist:
+            logging.info('\t{:<80}     [DIR BLACKLISTED]'.format(dirname))
+            subdirs[:] = []
+            continue
+        dst_path_offset = dirname[len(src_path)+1:]
```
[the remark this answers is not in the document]
BlaineEXE: Are you using a different linter than flake8? I lint all my code (though a couple things have slipped through), and I don't see this warning.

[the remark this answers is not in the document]
ErwanAliasr1: That is coming from https://pypi.python.org/pypi/pycodestyle in Atom

Makefile
ErwanAliasr1: I mean we just need to diff the actual makefile & the original one and put that diff into a separated commit. That helps understanding what are the differences you introduced.

stagelib/replace.py
```diff
@@ -0,0 +1,106 @@
...
+        logging.debug('\t{:<80}     [No STAGE_REPLACE_WITH_VARs to replace]'.format(file_path))
+        return template_text
+    for var in variable_matches:
+        var_name = var[len('STAGE_REPLACE_WITH_'):]  # strip STAGE_REPLACE_WITH_
+        logging.info('\t{:<80}  :: {:<38} <- {}'.format(file_path, var_name, globals()[var_name]))
+        template_text = _replace_global_in_text(var, template_text)
+    return template_text
+
+
+def do_variable_replace(replace_root_dir):
```
alfredodeza: this looks like a template engine. Why this PR is not proposing to use a well-established templating engine to do this work instead of adding untested (and undocumented) code?

stagelib/envglobals.py
```diff
@@ -0,0 +1,19 @@
+import logging
```
alfredodeza: nothing in this file will ever work if one of these environment variables is undefined. The `IMAGES_TO_BUILD` will break in split if that is unset, the `pringGlobal` is not necessary because there is no need to ever check the `globals()` in Python to see if a variable is defined in a module that is coming from an environment variable. That function does not need to exist, it can only be logged or printed directly from looking at the environment variables.

stagelib/replace.py
ErwanAliasr1: @alfredodeza What do you have in mind ?
We have some requirements like keeping a file-based mecanism and we don't have any yaml/xml stuff.

stagelib/envglobals.py
BlaineEXE: This is intended. The file explicitly should not work if any of the environment variables are undefined.

stagelib/replace.py
BlaineEXE: I had a discussion about this with @ErwanAliasr1. I'm open to suggestions, but I don't know of any templating engines that meet our needs of being able to have inputs defined by files that can be overridden by more specific files. In our discussion, we reasoned that we are reinventing a wheel to some perspective, but it is a lightweight engine that is very straightforward with no need to hack around forcing our override system to support inputs to an outside template engine. We can simply put text in a file with no extra boilerplate, and this will be easier for newcomers.

BlaineEXE:
@ErwanAliasr1 RE: Exception handling. I realize that I am not handling exceptions. I do intend that in every instance where you mention an exception, I do expect the staging engine to fail. In all cases, I would merely be catching the exception and raising my own.

stagelib/filecopytools.py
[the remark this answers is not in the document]
BlaineEXE: I'm not using pycodestyle as a linter. Flake8 is updated more frequently, is based on the pep8 standard, and also includes other features like inspection of cyclomatic complexity.

BlaineEXE:
Topic for discussion:

These scripts copy files. They do not hardlink. Moving the project structure to the one here will mean we can no longer use the `./generate-dev-env.sh ...` script as has been done before. 

Options to my mind:
1. Do away with the notion of working on hardlinked files
2. Hardlink files instead of copying when staging. One can then work in the staged dir; however, if a file has a `.bak` version, only the `.bak` version will be hardlinked, meaning devs will have to be aware of that.

stagelib/filecopytools.py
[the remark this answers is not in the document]
BlaineEXE: I did some digging, and flake8 is actually correct here. There was an update to pep8 Apr 2012 that pycodestyle doesn't implement correctly. https://hg.python.org/peps/rev/37af28ad2972

ErwanAliasr1:
+1 for removing generate-dev-env.sh, I don't see the real benefit of hardlinking

stagelib/replace.py
alfredodeza: @ErwanAliasr1 we've advocated in the past for a simplification of the current multi-distro multi-symlink process. The idea of trying to simplify is correct. What we've asked many times before is to *not* support a matrix of distributions because there is almost no value and it generates a tremendous burden in maintenance. The complexity in this PR to make it simpler to keep the matrix around should be another strong point in favor of just not supporting multi-distro generation.

This also does not make it any simpler for any downstream process.

stagelib/envglobals.py
alfredodeza: There are many ways code can be made to stop executing if the desire is to not continue when a certain condition is not met. Defining environment variables in this particular way, inspecting `globals()` and calling `.split()` on something that might not work is not intuitive, does not help debugging, and will not explicitly say what was required or what was missing.

It will raise a KeyError on the first item. What if you are missing 3 out of 4, or more than one? It will not tell you.

stagelib/replace.py
alfredodeza: @BlaineEXE this will absolutely not be easier for newcomers, as it is highly undocumented, tested with a non-standard bash script, and it tries to solve a very complicated arrangement of multi-distro building for the project. This is not straightforward at all

stagelib/envglobals.py
BlaineEXE: That can be improved.

stagelib/replace.py
BlaineEXE: Documentation can indeed be improved. Feel free to help with that.

As for not supporting multiple distros, I'd like to see what @leseb, @liewegas, and @dmick think. I agree that not supporting multiple distros will make it easier to support. It will also make it harder to support users who would like to build for other distros.

ErwanAliasr1: @alfredodeza that project does improve the way that downstream works as it put a real split between what is common and what is different from every distro. A downstream should only work in its distro directory.
And I'm sorry but I don't understand your comment here. We have to support several ceph releases, on several distro and several versions of those. That makes a matrix in anyway.  This solution clearly expose what is common and what is different between flavors. There is not a single link avoiding then colateral damage when you edit a file. That also kills code duplication we had.
So yes, that PR needs to be more documented, and we'll do but it does answer of maintainability purpose.  I can hear that's not so obvious for now, we'll work on clarifying that.

stagelib/envglobals.py
ErwanAliasr1: I agree that showing a traceback is rarely a good option and that's why I suggested using explicit error handling to clarify that.
About the variables, they come from the Makefile (which is the only way to use it) which populates them.

stage.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Copyright (c) 2017 SUSE LLC
+
+import os
+import logging
+import shutil
+import sys
+import time
+
+from stagelib.envglobals import (printGlobal, CEPH_VERSION, BASEOS_NAME, BASEOS_TAG, BASEOS_REG,
```
ErwanAliasr1: The parenthesis generates an invalid syntax in pyflakes

Makefile
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2017 SUSE LLC
```
ErwanAliasr1: why did you updated the Makefile this way ? I wanted to start a PR to prepare the migration but this makefile can't build anything :(

stagelib/replace.py
leseb: @alfredodeza If you're interested in learning more about multi-distro support, you should read https://github.com/ceph/ceph-container/issues/247. It gives the whole perspective about distro management and why it's essential/crucial.

We support multi-distro for builds **only**, we are not testing the code on all the distributions.

stage.py
```diff
@@ -0,0 +1,94 @@
...
+            staging_path = os.path.join(STAGING_DIR, image)
+            mkdir_if_dne(staging_path, mode=0o755)
+            copy_files(src_files, src_path, staging_path, blacklist)
+            recursive_copy_dir(src_path=os.path.join(src_path, image), dst_path=staging_path,
+                               blacklist=blacklist)
+        do_variable_replace(replace_root_dir=os.path.join(STAGING_DIR, image))
+
+
+if __name__ == "__main__":
+    main(CORE_FILES_DIR, CEPH_RELEASES_DIR, BLACKLIST_FILE)
```
ErwanAliasr1: LOG_FILE should be the last argument of main()

ErwanAliasr1:
The PR is missing an empty flavor-blacklist.txt

stage.py
```diff
@@ -0,0 +1,94 @@
...
+import sys
+import time
+
+from stagelib.envglobals import (printGlobal, CEPH_VERSION, BASEOS_NAME, BASEOS_TAG, BASEOS_REG,
+                                 ARCH, IMAGES_TO_BUILD, STAGING_DIR)
+from stagelib.filecopytools import list_files, mkdir_if_dne, copy_files, recursive_copy_dir
+from stagelib.replace import do_variable_replace
+from stagelib.blacklist import get_blacklist
+
+CORE_FILES_DIR = "src"
```
ErwanAliasr1: I have to admit I preferred core/ as you did it in the POC.

alfredodeza:
This PR tries to simplify a complex project structure without removing the complexity *that comes from the structure* but rather, by keeping the same objectives and attempting to simplify with a Makefile.

I am going to bring up the same points from a team discussion from early in 2017, that are unfortunately still present:

* Linking makes it hard for anyone to understand what is going and what the real source of the project is. For example the Leap42.2 entrypoint links to Ubuntu14.04. 
* The naming convention for distributions is unclear: why is there a RHEL 7.3 ? Is 7.4 not supported? The Dockerfile says it is "FROM rhel7:7-released" so which one is it?
* The matrix of supported distro, distro versions, and Ceph releases makes any work or understanding of the project tremendously difficult (as exemplified by this pull request)
* We keep questioning the value of supporting a matrix of multi-distros vs. a single one, as related to number of lines of code to maintain, number of builds to produce, and scenarios to test. For example Leap42.2 is not tested anywhere in our CI. Does that mean we should be responsible for building/releasing/supporting it? Adding just that one distro would make our test number explode.

These are the problems that we've seen in the past, that are still not addressed, and will not get any easier with this PR. 

I am -1 on merging this work since it will solidify the premise that the layout needs to be kept around, while not helping reduce any of the issues we have raised.

ErwanAliasr1:
@alfredodeza I'm sorry, you are so wrong here .... You are not looking in the right direction.

This PR is just about to introduce the building tool, I'm currently working on a new PR to perform the migration from the actual system (that you read) and this new architecture. 

That PR is an empty shell for now and doesn't implement any OS.

> Linking makes it hard for anyone to understand ....
If you paid attention to this PR, that's exactly the goal of it.
There is no symlinks at all and all our structure is about removing them to increase the maintainability. Every single dir is running is own files and prevent any collateral damage by editing any distro specific file.

> For example Leap42.2 is not tested anywhere in our CI. Does that mean we should be responsible for building/releasing/supporting it?
At least, we should be in position to offer a reliable build mechanism for every downstream. About the responsibility ? It's up to every downstream to handle it. If we do offer a proper way of handling distros (and I do think we do) then we'll get more contributions (and @BlaineEXE  is a good example of it) & a better product.

I'm not answering the other topics you point as they are obviously a misunderstanding of the actual goal of this PR.

If you want to get a better overview of how the matrix will be implemented, look at this directory https://github.com/SUSE/ceph-container/tree/poc-prune-tree/src and only read the core & all_ceph_release directory.

stagelib/envglobals.py
BlaineEXE: I have added exception handling to the `envglobals.py` module to fix this issue.

BlaineEXE:
@alfredodeza: I understand that building different flavors of containers makes the project structure complicated. The project is not as simple as I would like, but I am working to make it as simple as I can. I am open to suggestions for ways to improve upon the design. As to the points you have raised:

---

> Linking makes it hard for anyone to understand what is going and what the real source of the project is. For example the Leap42.2 entrypoint links to Ubuntu14.04.

Once we begin adding distros into this structure, we will get rid of linking.

---

> The naming convention for distributions is unclear: why is there a RHEL 7.3 ? Is 7.4 not supported? The Dockerfile says it is "FROM rhel7:7-released" so which one is it?

This seems like an issue that the RHEL 7.x release is not properly maintained by its owner rather than a problem with the project structure.

---

>The matrix of supported distro, distro versions, and Ceph releases makes any work or understanding of the project tremendously difficult (as exemplified by this pull request)
We keep questioning the value of supporting a matrix of multi-distros vs. a single one, as related to number of lines of code to maintain, number of builds to produce, and scenarios to test. For example Leap42.2 is not tested anywhere in our CI. Does that mean we should be responsible for building/releasing/supporting it? Adding just that one distro would make our test number explode.

As has been mentioned by @leseb, the goal is only to support building different releases. Testing is not guaranteed. I have spoken with @ErwanAliasr1 about how nice it would be to get each interested party to test the releases they want to support. For my part, I would like to find a way to contribute openSUSE CI to the project. As I am sure that Red Hat would like to CI CentOS and RHEL flavors. This currently does not exist, but we can work together to improve the project overall. The open source project, on its own, need only support one OS for each supported Ceph version. This seems that it would be Ubuntu.

BlaineEXE:
@ErwanAliasr1  I have now handled all exceptions that may occur with files.

stage.py
BlaineEXE: From PEP8:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

BlaineEXE:
Recent `push -f` fixes a bad assumption. There are now 4 parameters that affect OS and OS version.

Old:
`BASEOS_NAME, BASEOS_TAG, BASEOS_REG`
Does not support the case where I want to  build ceph-container source ubuntu-16.04 into alpine:3.6 container, as an example that might not work in reality.

New:
`OS_NAME, OS_VERSION, BASEOS_REG, BASEOS_TAG`
This will support specific ceph-container code into any arbitrary container base.

stagelib/replace.py
alfredodeza: @leseb yes, I've read that isssue before. What is a supported distro? Something where code exists to "create" a version of it? There aren't tests nor any sort of validation. What is the benefit of the "community" for something that is not tested? If it is essential/crucial, why aren't we expanding the matrix to really test every single permutation that this project attempts to support? 

What does it mean to "support multi-distro for builds only" ? Again, are you verifying these builds work? where are the tests for these? What is the use of a build for a project that we don't know if it runs?

This PR changes over six hundred files trying to improve the build situation. This is not a solution, it is a symptom of the larger problem.

The Ceph project for the Luminous release only builds (and tests!) Ubuntu Xenial, and CentOS7.

BlaineEXE:
I'm on US holiday today (19 Feb), but I thought I should take a couple minutes to post WIP commit of my current progress building the daemon-base/luminous-ubuntu image.

BlaineEXE:
As a side note, I spent some of my weekend getting a squid caching proxy set up on my laptop to speed up repeat builds a bit. Would it be worth adding the instructions for that process to this project?

ktdreyer:
I think we could drastically simplify our QE efforts and ceph-container code by reducing or eliminating the multiple-distro support.

stagelib/replace.py
liewegas: My $0.02: don't let perfect be the enemy of good (or even just better).

We need multi-distro support.  18.04 will be added shortly, and downstream vendors will be adding in their own distros as well.

We need multi-version support of some form.  We'll be building containers for luminous, soon mimic, and probably even jewel for some time.  That could be maintained in a parallel for each release but there appears to be value in this approach that automates some of that maintenance. 

Whether all distro combinations are tested seems like an orthogonal issue.  There is value in having an untested build (even if it is not as much value as a tested one), and having a tested distro target that is close to what you need (e.g. tested ubuntu vs untested debian, tested centos vs untested rhel).

ErwanAliasr1: thx @liewegas for sharing this vison. We do need all downstreams to be part of the effort. We are here, Suse does. That's a good start. I cannot see how downstreams could accept shipping containers without having their own stuff inside. This PR is about ease this process by having a clear segregation between upsteam & downstream code.

stagelib/git.py
```diff
@@ -17,3 +17,10 @@ def get_branch():
 
 def get_hash():
     return _run_cmd(['git', 'rev-parse', '--verify', 'HEAD'])
+
+
+def file_is_dirty(file_path):
+    file_status_msg = _run_cmd(['git', 'status', str(file_path)])
+    if 'nothing to commit' in file_status_msg:
```
ErwanAliasr1: Be careful, you are making an assumption about the locale. We are not all using en_US ;)
Maybe run_cmd() should enforce the locale to get a constant output.

ErwanAliasr1:
@BlaineEXE IIRC, that only needs a few variables to export to ask apt using the cache. That could be useful for some.

ErwanAliasr1:
@BlaineEXE I don't like the move of ceph-release in a .old then adding the files directly. That clearly kills the tracability of thoses files. A mv would be clearly better here.

ErwanAliasr1:
@BlaineEXE I also wonder what is the advantage of having src/daemon-base as we could have ceph-release/ALL/daemon-base/

So src could be useless ?

stagelib/git.py
BlaineEXE: Good catch. I'll see if there is a better way to get file dirty status or enforce a locale.

BlaineEXE:
@ErwanAliasr1 

RE: ceph-release.old - I'm not sure exactly what you are having issue with. Regardless the tmp commit says that it is temporary. It has been making it easier for me to work for now, and the git commits can be rebased/squashed/etc. later to make it appear as a `mv`.

RE: src/ vs ceph-release/ALL/ - you are correct. src/ is not strictly necessary. On one hand I don't like that it's unnecessary. But on another hand, having the core source base in src/ makes it clear that this is where the source of the project lives. Another alternative might be to have `src/ALL`, `src/ALL/ubuntu`, `src/luminous`, etc.

ErwanAliasr1:
@BlaineEXE in fact, I wanted to find kind of a stable tree we agree on to start my PR on importing the new OSes. I do understand that src/ represents the core of the project but having everything in ceph-releases/ would locate all the code in a single location and so maybe easier to track.

BlaineEXE:
I want to get all ceph versions of ubuntu building in this PR to set the framework for things to come. As well, we need documentation. That to me is our stable tree.

stagelib/git.py
BlaineEXE: Fixed in latest `push -f`

Makefile
```diff
@@ -0,0 +1,95 @@
+# Copyright 2016 The Rook Authors. All rights reserved.
```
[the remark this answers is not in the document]
ErwanAliasr1: @BlaineEXE this comment is still valid ;)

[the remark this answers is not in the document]
BlaineEXE: I'm not sure I can do this. I started with a makefile that ran integration tests and then added their content to that with changes.

BlaineEXE:
Latest update: now builds luminous-ubuntu-16.04 daemon-base and daemon

BlaineEXE:
## Topic for discussion - source layout
### Current
```
src/daemon
src/daemon-base
ceph-releases/ALL/ubuntu
ceph-releases/jewel
ceph-releases/kraken
```
This option is fine, but having `src` and `ceph-releases/ALL` isn't really necessary.

### `src` option
```
src/all-ceph-releases/daemon
src/all-ceph-releases/daemon-base
src/all-ceph-releases/ubuntu
src/jewel
src/kraken
```
I like this option the best. It is very clear where the main code for this project resides upon first glance at the repo because there is a clear `src` directory.

### `ceph-releases` option
```
ceph-releases/ALL/daemon
ceph-releases/ALL/daemon-base
ceph-releases/ALL/ubuntu
ceph-releases/jewel
ceph-releases/kraken
```
This is also a viable option. I think it is less clear than having a `src` directory, but it fits the existing codebase better.

dmick:
A few first-glance notes: 

1. The change to rgw_s3_client.py doesn't seem to belong in this changeset
2. the stage.py and friends need block and func doc comments badly
3. I assume before merging these commits will be rewritten (eliminating the merge turds from fbb5c49)

dmick:
printGlobal() really really confuses me...is it really searching globals() by *value*, and printing the first global var that matches the value?  i.e., if I had A=1 and B=1, printGlobal(A) would search for 1, find A, and print A : 1 ?

that doesn't at all seem like what you want from that function, and it means the implementation is really confusing; surely "printGlobal('A') -> print globals[key] is what you're really after?

psukys:
Is there a blueprint/roadmap for this PR, so that I could help, if possible?

Makefile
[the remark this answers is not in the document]
ErwanAliasr1: I'm really sorry. I don't get the issue you have. Why can't you put the original make in the inital commit to keep the original one and make a commit with the diff against the actual one to see your changes.
Today we have just a modified makefile where we can't compute the changes you made.

ErwanAliasr1:
@psukys I think we are very close from the merge, I doubt we can really receive contributions at this stage but appreciate your proposal. You still can review the actual code & comment it if you found something wrong in it. Testing will be avail soon too.

ErwanAliasr1:
@BlaineEXE can you please rebase against master to solve the conflicts ?

ErwanAliasr1:
@BlaineEXE can you please rebase against master to solve the conflicts ?

ErwanAliasr1:
@BlaineEXE I started the centos 7 support which is pretty easy to do.
I just had to add tweak the actual code to get a clean implementation.
Please have a look at the commits previous my Centos 7 support one on https://github.com/ceph/ceph-container/pull/908
If you find them good, please pick them, I'll rebase then after on your branch.

Makefile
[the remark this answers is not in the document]
dmick: There's no logical connection between the original rook makefile and the one adapted for ceph-container; I don't see why commit history would be relevant.

GregMeno:
@ErwanAliasr1 You claim that the point of this PR is to remove the symlinks, and I admit to being just as confused as @alfredodeza here. I'm happy that @BlaineEXE added a section to the README "Project structure and staging" AND I really don't think that is suitable in explaining how one adds support for a distro using the new structure or what the goals of the new structure are. The discussion in this PR certainly doesn't clarify that either. I'm -1 on merging this till you add clear docs in the README on what we are doing with this work.

stagelib/envglobals.py
dmick: I don't understand why envglobals.py exists either; also see prior comments about printGlobal just being wrong

stagelib/replace.py
dmick: on the template engine question: it seems odd to me that we're reinventing variable substitution here too.  The precedence definition/mechanism is outside the scope of the template engine.

BlaineEXE:
@dmick 
RE: #1 - Yes. This was unintentional. Needs to be fixed.
RE: #2 - Yes. Documentation is needed.
RE: #3 - I can rewrite commits as appropriate. Though I'm not sure what you are referring to by 'merge turds'.
RE: envglobals - 
The purpose of the envglobals module is firstly to make sure that the required environment variables are set and give a useful error message if they aren't. The printGlobal function exists to output the global var as a message to the user as well as a log message, though you are correct that print(globals()["global-var-name"]) would be more straight forward. This was perhaps a rabbit hole I got lost in. I should change that. To your confusion about what exists now, the `is` operator in Python does not check that the value is equal, it checks if the object "is" the exact same object (i.e., refers to the same memory location).

dmick:
Agh.  because the comments aren't replies, I don't even know what you're responding to.

dmick:
merge turds are >>>> and <<<< markers from merge conflicts

BlaineEXE:
@ErwanAliasr1 
RE > I'm really sorry. I don't get the issue you have. Why can't you put the original make in the inital commit to keep the original one and make a commit with the diff against the actual one to see your changes.
Today we have just a modified makefile where we can't compute the changes you made.

Moving the Makefile into the root in that commit would result in a non-working feature. I can move it, but I don't see the necessity when the commit introduces unused code.

BlaineEXE:
@dmick If there are changes that need to be made, please submit a change request. It's a lot easier for me to track what I have and haven't fixed that way.

stagelib/envglobals.py
```diff
@@ -0,0 +1,45 @@
...
+
+"""
+    sys.stderr.write(errtext.format(unset_var))
+    sys.exit(1)
+
+logger = logging.getLogger(__name__)
+
+
+def printGlobal(var):
+    varname = [name for name in globals() if globals()[name] is var][0]
```
BlaineEXE: This could be more straightforward `globals()['var_name']`

dmick: yeah, if anything it seems to me that the "is var" works from luck, as literals are implemented as global shared objects, so there is only one instance of "1", for example.  anyway...can't think of a reason to search by value.

BlaineEXE: I just did a quick test of this, and since strings are immutable in Python, you are right. And it's on my to-fix list.

stagelib/replace.py
BlaineEXE: If you have a template engine in mind, please suggest one. I am somewhat familiar with Jinja 2, but if we use it, templates must be filled by yaml files. Doing yaml file overriding of specific values without deleting different, existing variables seems non-trivial to me. Furthermore, my proposed implementation supports nested variables. I wasn't sure what Jinja's support of that would be. It seems easier to write a few python lines just to read a file for the replacement. And it was an extra 3 lines of code to support nested variables. It was the fastest path to success I found, and there are quite extensive integration tests for all of the features of this script to verify that it works as expected.

dmick: I don't think jinja2 even knows what data file input format you might use, much less yaml; you call render() with a set of kv pairs, and it's up to you where those come from and how overriding is done; I would assume most templating libs do the same.  And a 'merge' operation is pretty standard for multiple configuration sources; you just define order of precedence for duplicated keys.

I don't have any alternate design fleshed out in my head, it was just surprising to see basic variable replacement in text files *not* using an existing solution.  I don't yet really understand the nestedness

dmick:
It appears as though the Dockerfile fragments can't always be "empty"; they have to at least contain a valid command.  ":" or "/bin/true" will do.  (My current test is with the currently-default distro/release combo, luminous@amd64@ubuntu@16.04@_@ubuntu@16.04, and I must create a ceph-releases/ALL/ubuntu/daemon-base/__DOCKERFILE_CLEAN_DISTRO_SPECIFIC__ file containing something like "/bin/true" to allow it to build daemon-base)

BlaineEXE:
> It appears as though the Dockerfile fragments can't always be "empty"; they have to at least contain a valid command. ":" or "/bin/true" will do. (My current test is with the currently-default distro/release combo, luminous@amd64@ubuntu@16.04@_@ubuntu@16.04, and I must create a ceph-releases/ALL/ubuntu/daemon-base/DOCKERFILE_CLEAN_DISTRO_SPECIFIC file containing something like "/bin/true" to allow it to build daemon-base)

That's correct. If the `__VAR__` file is supposed to contain actions that need done, it needs to have `/bin/true` so that the `... & __VAR__ $ ...` will proceed correctly. I'll add this to my documentation todo.

BlaineEXE:
Ubuntu builds should now be working. Let me know if there are any issues. I am now working on documentation.
```
REPOSITORY          TAG                           IMAGE ID            CREATED             SIZE
ceph/daemon         jewel-ubuntu-14.04-amd64      9e8b01c1de72        10 minutes ago      919MB
ceph/daemon         luminous-ubuntu-16.04-amd64   cc7c8a8da2f7        15 minutes ago      612MB
ceph/daemon         kraken-ubuntu-16.04-amd64     d0108d6944f4        15 minutes ago      845MB
ceph/daemon         jewel-ubuntu-16.04-amd64      6ad4f998c094        15 minutes ago      907MB
ceph/daemon-base    kraken-ubuntu-16.04-amd64     2f56e1acb49a        29 minutes ago      686MB
ceph/daemon-base    luminous-ubuntu-16.04-amd64   0133081469f0        30 minutes ago      466MB
ceph/daemon-base    jewel-ubuntu-16.04-amd64      ac4226af691a        30 minutes ago      748MB
ceph/daemon-base    jewel-ubuntu-14.04-amd64      a8b9c0a87be0        33 minutes ago      817MB
```

BlaineEXE:
Readme updates are in progress. Committing my work so far for any early review. I've also rebased with master.

BlaineEXE:
@leseb - Thoughts on if and how the travis tests should be modified for this PR?

src/daemon/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+# Add bootstrap script, ceph defaults key/values for KV store
+ADD *.sh ceph.defaults check_zombie_mons.py ./osd_scenarios/* entrypoint.sh.in disabled_scenario /
+
+# Modify the entrypoint
+RUN bash "/generate_entrypoint.sh" && \
+  rm -f /generate_entrypoint.sh && \
+  bash -n /*.sh
+
+# Add volumes for Ceph config and data
+VOLUME ["/etc/ceph","/var/lib/ceph", "/etc/ganesha"]
```
ErwanAliasr1: @BlaineEXE , @leseb told me this wasn't required anymore.

ErwanAliasr1:
@BlaineEXE I cannot checkout the tree (I'm on PTO) but I have the feeling that some files are duplicated even they seems to be identical like docker_exec.sh.

Did you checked the similarity of files between src/ & jewel/kraken/luminous dirs to avoid duplication ?

ceph-releases/ALL/ubuntu/daemon/__DOCKERFILE_INSTALL__
```diff
@@ -0,0 +1,25 @@
...
+          | tee /etc/apt/sources.list.d/nfs-ganesha.list ; \
+      fi' && \
+    echo 'Install packages' && \
+      DEBIAN_FRONTEND=noninteractive apt-get update && \
+      DEBIAN_FRONTEND=noninteractive apt-get install -y \
+        wget unzip uuid-runtime python-setuptools udev dmsetup && \
+      apt-get update && apt-get install -y  --no-install-recommends --force-yes $PACKAGES && \
+      dpkg -s $PACKAGES && \
+      apt-get clean && \
+    echo 'Download & install confd' && \
```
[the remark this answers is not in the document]
BlaineEXE: All install and cleanup operations should be done in a single `RUN` command to minimize the size of the final docker image. Once `docker build --squash` is no longer experimental, we should revisit this. #913

ceph-releases/ALL/ubuntu/daemon/__DOCKERFILE_PREINSTALL__
```diff
@@ -0,0 +1,5 @@
+ARG CONFD_VERSION=0.10.0
```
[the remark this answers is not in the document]
BlaineEXE: The `CONFD_VERSION` and `KUBECTL_VERSION` args are used by the install.

src/daemon/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+
+#======================================================
+# Add ceph-container files
+#======================================================
+
+# Add s3cfg file
+ADD s3cfg /root/.s3cfg
+
+# Add templates for confd
+ADD ./confd/templates/* /etc/confd/templates/
```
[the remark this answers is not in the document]
BlaineEXE: Not with confd install in the install/cleanup `RUN` command.

src/daemon/Dockerfile
BlaineEXE: If that's true, that is a different PR.

BlaineEXE:
@ErwanAliasr1 
> I cannot checkout the tree (I'm on PTO) but I have the feeling that some files are duplicated even they seems to be identical like docker_exec.sh.
>
> Did you checked the similarity of files between src/ & jewel/kraken/luminous dirs to avoid duplication 

Yes. Thoroughly.

leseb:
@BlaineEXE we have to change the CI and how it builds images. I'll take care of that, I can put this at the top of your PR. I'll give it a shot tomorrow.

ceph-releases/ALL/ubuntu/daemon/__DOCKERFILE_INSTALL__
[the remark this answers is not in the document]
BlaineEXE: I think the core of this is that we want reuse. I've added __DAEMON_PACKAGES__ to be a record of what packages need installed. There are several __*_PACKAGE__ variables that allow OS-es to override these (for example to disable apt installs of confd).

Makefile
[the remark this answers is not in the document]
dmick: that would make sense if there were any logical connection between the original rook makefile and the one adapted here for ceph-container, but, there isn't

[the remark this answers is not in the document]
BlaineEXE: I think I agree with @dmick here, but my commits already are set up as requested by @ErwanAliasr1 . I do think that we should keep the Rook copyright; I used enough of the Rook-contributed code (or architecture) that I think it would plagiarism not to acknowledge them.

BlaineEXE:
I would consider this PR done save for any review comments I get. 

I did my best to take care of @ErwanAliasr1's latest comments and also keep image size to an absolute minimum.

Additionally, while editing the readme, I realized that a script similar in concept to the old `single-ceph-docker-workflow.sh` would be good for determining what ceph-container flavors are modified by a branch. This can be good for devs to understand the implications of their modifications, and it could be used for CI to only test container builds affected by updates in a PR. I added a `make flavors.modified` target for this script as well.

README.md
```diff
@@ -9,45 +9,174 @@ Ceph-related Docker files.
...
+ceph-releases/ALL/FILE
+src/{daemon-base,daemon}/FILE
+src/FILE
+# Least specific
+```
+
+#### Basic templating in staging
+##### Variable file replacements
+In any source file, a special variable in the form `__VAR_NAME__` (two leading and trailing
+underscores with captial letters and underscores between) can be placed. Once all files are staged,
```
leseb: s/captial/capital/

README.md
```diff
@@ -9,45 +9,174 @@ Ceph-related Docker files.
...
+`__VAR_NAME__`. Trailing whitespace is stripped from the variable file before insertion.
+`__VAR_NAME__` files are allowed to be empty, but they are not allowed to be nonexistent if a file
+declares them.
+
+If the `__DO_STUFF__` file is supposed to contain actions that need done it generally needs to return
+true. As an example, `echo 'first' && __DO_STUFF__ && echo 'last'` will print 'first' and 'last'
+correctly only if `__DO_SUFF__` returns true. A take-no-action override needs to have the content
+`/bin/true`, as an empty file will cause an error.
+
+##### Environment varable replacements
```
leseb: s/varable/variable/

flavors-modified-vs-master.py
```diff
@@ -0,0 +1,54 @@
...
+import sys
+
+
+def _run_cmd(cmd_array):
+    return subprocess.check_output(cmd_array).decode("utf-8")
+
+
+# Stage the flavor and return the staging dir
+def _stage_flavor(flavor):
+    stage_output = _run_cmd(['make', 'FLAVORS_TO_BUILD='+flavor, 'stage'])
```
leseb: missing whitespace around arithmetic operator

flavors-modified-vs-master.py
```diff
@@ -0,0 +1,54 @@
...
+    match = staging_dir_pattern.search(stage_output)
+    if not match or not match[1] or match[1] == '':
+        print("Could not find staging dir for:\n{}".format(stage_output))
+        sys.exit(1)
+    return match[1]
+
+
+def _filediff_intersects_sources(filediff, sources):
+    for f in filediff.splitlines():
+        file_pattern = re.compile(r'^.*\s+<-\s+'+f+'$', re.MULTILINE)
```
leseb: ditto

Makefile
```diff
@@ -0,0 +1,188 @@
...
+	@echo '                      USE AT YOUR OWN RISK! This may remove non-project images.'
+	@echo ''
+	@echo '  Testing:'
+	@echo '    lint              Lint the source code.'
+	@echo '    test.staging      Perform stageing integration test.'
+	@echo ''
+	@echo '  Help:'
+	@echo '    help              Print this help message.'
+	@echo '    show.flavors      Show all flavor options to FLAVORS_TO_BUILD.'
+	@echo "    flavors.modified  Show the flavors impacted by this branch's changes vs master."
```
leseb: It does not only show the flavors impacted, it also builds stages accordingly so we need to capture this in the help.

flavors-modified-vs-master.py
```diff
@@ -0,0 +1,54 @@
...
+        match = file_pattern.search(sources)
+        if match:
+            return True
+    return False
+
+
+# Get list of files different from master
+filediff = _run_cmd(['git', 'diff', '--name-only', 'master'])
```
leseb: I think using `origin/master` will be more accurate than the local master branch.

Makefile
leseb: Afterthoughts, I think we should re-build images after each PR merges so we can benefit from distro potential security updates. So I'll change the build system in that sense.

BlaineEXE:
@leseb - I'll work on your comments. As for ceph-releases-old, I had been keeping those around for reference. I can remove all those if you think that's best.

leseb:
@BlaineEXE yes, I think we should remove them.

BlaineEXE:
Latest update adds +x permission to all *.sh, *.sh.in, and *.py files.

leseb:
@BlaineEXE the job is ready and waiting for this to be merged: https://2.jenkins.ceph.com/job/ceph-container-build-push-imgs/4/console

It's currently failing because there is no Makefile, yet ;-).

BlaineEXE:
Added ganesha packages to the daemon-base.

README.md
```diff
@@ -9,67 +9,204 @@ Ceph-related Docker files.
...
+Ideally, adding a new Ceph release is fairly easy. In the best case, all that needs done is adding
+flavors for the new Ceph version to the Makefile. At minimum, `ALL_BUILDABLE_FLAVORS` must be
+updated in the Makefile.
+
+In the worst case, trying to make as few modifications as possible:
+1. Add flavors for new Ceph versions to the Makefile.
+   - At minimum: `ALL_BUILDABLE_FLAVORS`.
+2. Edit `src/` files to support the new version if necessary, making sure not to break previous
+   versions.
+3. Edit `src/ALL/<distro>` files to support the new version if necessary, making sure not to break
```
ErwanAliasr1: @I'm very confused why you put it in src/ALL and not in ceph-release/ALL.

I think we shall strictly define when a file is supposed to go in src/ or in ceph-release/

README.md
```diff
@@ -9,67 +9,204 @@ Ceph-related Docker files.
...
+   previous versions.
+4. Add `src/<new ceph version>` files to support the new version if necessary.
```
ErwanAliasr1: Diito. Why not in ceph_release/<new ceph version> ?

README.md
```diff
@@ -9,67 +9,204 @@ Ceph-related Docker files.
...
+1. Add flavors for the new distro to the Makefile.
+   - At minimum: `ALL_BUILDABLE_FLAVORS`.
+2. Add a `src/ALL/<new distro>` directory for the new distro
+   - Make sure to install all the required packages for `daemon-base` (see
+     `src/daemon-base/__CEPH_BASE_PACKAGES__)` and for `daemon` (see
+     `src/daemon/__DAEMON_PACKAGES__`).
+   - Make sure to specify all `__VAR__` files without sane defaults for the container builds.
+   - Make sure to override any `__VAR__` file sane defaults that do not apply to the new distro.
+   - Refer to other distros for inspiration.
+3. If necessary (try to avoid this as much as possible), add a `src/<ceph release>/<new distro>`
```
ErwanAliasr1: I think we have to be explicit why we shall avoid this.

README.md
```diff
@@ -9,67 +9,204 @@ Ceph-related Docker files.
...
+   previous versions.
+4. Add `src/<new ceph version>` files to support the new version if necessary.
+5. Build test versions of the images (`make FLAVORS_TO_BUILD=<new release flavors> build`).
+6. Test the images in your environment.
+7. Make a PR of your changes.
+
+### Adding a distro build
+1. Add flavors for the new distro to the Makefile.
+   - At minimum: `ALL_BUILDABLE_FLAVORS`.
+2. Add a `src/ALL/<new distro>` directory for the new distro
```
ErwanAliasr1: The actual code doesn't do that at all as everything which is distro specific is in ceph-release/.
I'm currently updating the doc in my centos branch, I can handle all those changes if you like.

README.md
BlaineEXE: Typo

README.md
BlaineEXE: There is no `src/All`. A typo.

README.md
BlaineEXE: Typo. Good catch. I'm working on patching the docs

README.md
BlaineEXE: Yes.

Makefile
```diff
@@ -0,0 +1,170 @@
...
+
+build.all:
+	@$(MAKE) FLAVORS_TO_BUILD="$(ALL_BUILDABLE_FLAVORS)" build.parallel
+
+
+# ==============================================================================
+# Clean targets
+.PHONY: clean clean.nones clean.all clean.nuke
+
+clean.image.%: do.image.%
```
ErwanAliasr1: That's pretty curious to have do.image as a dependency of clean.image. That's explain why a make clean does populate the staging dir.

Btw, why not having clean.all being the default clean target ?
That looks like more consistent with a user expectation.

Makefile
```diff
@@ -0,0 +1,170 @@
...
+
+do.image.%: daemon.% ;
+
+stage: $(foreach p, $(FLAVORS_TO_BUILD), stage.$(p)) ;
+build: $(foreach p, $(FLAVORS_TO_BUILD), do.image.$(p)) ;
+push:  $(foreach p, $(FLAVORS_TO_BUILD), do.image.$(p)) ;
+
+build.parallel:
+# Due to output-sync, will not output results until finished so there is no text interleaving
+	@$(MAKE) --jobs --output-sync build
```
ErwanAliasr1: my make 3.82 doesn't have --output-sync making this target failing.

ErwanAliasr1: The PR looks good to me.

ErwanAliasr1:
The resync required to keep the tracability for the moved filed required to rework the commit while @BlaineEXE is away.

Commit https://github.com/ceph/ceph-container/pull/927/commits/c16a4bf24fd389d4cc174fa1adc8ecbc8e2374bc is handling the resync task.

@BlaineEXE I'm closing in favor of #927 to handle the final integration before the merge.

PR #910, opened 2018-02-18
Exit CephNFS with error if ganesha daemon fails

tombarron:
while still logging the daemon both to STDOUT and the location
configured in the daemon's configuration file.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1546192


guits:

tombarron:
None of the above failing tests exercised start_nfs.sh, the only place touched by the candidate patch.

PR #931, opened 2018-03-08
staging: fix bug when checking for dirty branch

BlaineEXE:
Previously, `file_is_dirty('')` was used to determine if the branch had
any dirty files, but because of the regex match in the method, this
cannot be guaranteed to return as expected. Add a `branch_is_dirty`
method to fix this.


ErwanAliasr1:
Thx !

PR #932, opened 2018-03-08
Envglobals cleanup

BlaineEXE:
I noticed that I had forgotten to add DAEMON_BASE_IMAGE and DAEMON_IMAGE to the envglobals file. They are required variables and should be checked.

While going about the small addition, I realized that it was really unclear that the main purpose of envglobals was to verify that certain env vars were set, as they are known to be required for operation. I refactored envglobals to make this both more clear and more easily extensible.

This refactor also changed how I start to initialize logging, so that is taken care of at the same time so it doesn't end up broken.


BlaineEXE:
@ErwanAliasr1 / @leseb Merge thoughts?

ErwanAliasr1:
Looks good to me.

Do we really need keeping so many commits ? Can't we merge some of them ?

BlaineEXE:
I think I can squash some of them ...

leseb:
@BlaineEXE looks like you went from 8 to 4 commits, LGTM, @ErwanAliasr1 ?

maint-lib/stage.py
```diff
@@ -48,14 +47,16 @@
...
-    logging.info('Git repo:   {}'.format(git.get_repo()))
-    logging.info('Git branch: {}'.format(git.get_branch()))
-    logging.info('Git commit: {}'.format(git.get_hash()))
 
     print('')
     verifyRequiredEnvVars()
     print('')
+
     exportGitInfoEnvVars()
+    logging.info('GIT_REO:    {}'.format(getEnvVar('GIT_REPO')))
```
ErwanAliasr1: Typo : GIT_REO/GIT_REPO

BlaineEXE:
Typo fixed.

PR #942, opened 2018-03-09
[skip ci] Refactor arch

BlaineEXE:
Refactor ARCH to refer to Ceph's arch, not golang arch.

Convert ARCH to GO_ARCH where needed.
Ceph x86_64 -> golang amd64
Ceph aarch64 -> golang arm64

Built on changes from #932

PR #947, opened 2018-03-12
[skip ci] Remove blacklist

### LINKED ISSUE #938: Staging blacklist is unnecessary?
Do we want to remove the blacklist functionality? It might be unnecessary.

BlaineEXE:
This initially seemed like it might be necessary, but overrides and
    blank `__VAR__` files do the job very well. The blacklist functionality
    adds unnecessary complexity that might be confusing, so we should remove
    it rather than let it be vestigial.

This is based on #932 

Fixes #938


ErwanAliasr1:
Seems to be more on #942 than #932.

Why do you stack those 2 PRs ? They doesnt' seems so dependent.

BlaineEXE:
I did it so I don't have to rebase after refactor-arch is merged since they edit some of the same files.

ErwanAliasr1:
Please rebase against master

PR #955, opened 2018-03-14
[skip ci] ci: use jenkins variables

leseb:
Instead of discovering branche and commit we can simply re-use jenkins
env variables during the job execution.


leseb:
need rebase on top of https://github.com/ceph/ceph-container/pull/942

PR #970, opened 2018-03-16
make: improve flavor modified

leseb:
We must fetch origin master to make sure our understanding of
origin/master's content is up to date. If we don't do this we will find
irrelevant/wrong results.

### DIFF
--- a/maint-lib/flavors-modified-vs-master.py
+++ b/maint-lib/flavors-modified-vs-master.py
@@ -42,9 +42,10 @@ if 'VS_BRANCH' in os.environ and not os.environ['VS_BRANCH'] == '':
 
 # Get list of files different from VS_BRANCH
 try:
+    _run_cmd(['git', 'fetch', 'origin', 'master'])
     filediff = _run_cmd(['git', 'diff', '--name-only', VS_BRANCH])
 except subprocess.CalledProcessError as c:
-    _fatal('Could not get file diff. Is the VS_BRANCH specified a real branch?')
+    _fatal('Could not fetch origin master or get file diff. Is your remote named origin? Is the VS_BRANCH specified a real branch?')
 
 # Files that haven't been committed don't show up in git diff, so also list those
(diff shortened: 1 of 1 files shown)


leseb:

leseb:
jenkins test flake8

maint-lib/flavors-modified-vs-master.py
```diff
@@ -42,9 +42,10 @@ def _filediff_intersects_sources(filediff, sources):
 
 # Get list of files different from VS_BRANCH
 try:
+    _run_cmd(['git', 'fetch', 'origin', 'master'])
     filediff = _run_cmd(['git', 'diff', '--name-only', VS_BRANCH])
 except subprocess.CalledProcessError as c:
-    _fatal('Could not get file diff. Is the VS_BRANCH specified a real branch?')
+    _fatal('Could not fetch origin master or get file diff. Is your remote named origin? Is the VS_BRANCH specified a real branch?')
```
BlaineEXE: The line length here does not pass linting.

maint-lib/flavors-modified-vs-master.py
```diff
@@ -42,9 +42,10 @@ def _filediff_intersects_sources(filediff, sources):
 
 # Get list of files different from VS_BRANCH
 try:
+    _run_cmd(['git', 'fetch', 'origin', 'master'])
```
BlaineEXE: This command outputs the below text  before the flavor output.
```
From github.com:SUSE/ceph-container
 * branch            master     -> FETCH_HEAD
```
Is this intentional? It seems like a side-effect. I originally wrote this to only return flavors modified with no other text so it could be used as input for scripting.

leseb: You're right, this actually breaks one of my scripts. At first, I thought this was a good indication. Let me fix that.

leseb: Fix in https://github.com/ceph/ceph-container/pull/976

PR #971, opened 2018-03-16
README: add doc for finding image tags

leseb:


ktdreyer:
thanks for adding docs here

PR #976, opened 2018-03-21
maint-lib: make git fetch silent

leseb:
This fixes the script integration by silencing any output.


leseb:

leseb:
jenkins test ceph_ansible-luminous-xenial-bluestore_osds_container

BlaineEXE:
The best way in Python to make sure this is always silent is to assign the output to the dummy variable `_` (single underscore). The quiet flag may not always mean no output, and that may change with different versions of git.

PR #978, opened 2018-03-23
Makefile: Adding CEPH_POINT_RELEASE support

ErwanAliasr1:
This patch is about adding an option to enforce a particular point release of ceph like "12.2.0" instead of the latest.

This requires adding a field (named CEPH_POINT_RELEASE) on the flavor name, just after the CEPH_VERSION field so we have something like :
    luminous,-12.2.2,[....]

This change implies this CEPH_POINT_RELEASE is encoding the release but also the syntax for a specific package manager.
For centos, yum expect the package version to be prefixed by a dash, so CEPH_POINT_RELEASE should have a dash for centos.
This make possible having a generic patch for all distro.

Note that if the field is let empty, like in "luminous,,x86_64,[...]", the field will be automatically replaced by an empty string.
This is very useful for users not to have the need of writing "luminous,'',x86_64,[...]"
An empty CEPH_POINT_RELEASE, which is the current default, leads to installing the lastest available version of the ceph packages.

This patch only implements it for centos as it will depends on distributions & repositories to offer/enable this feature.


Makefile
```diff
@@ -17,15 +17,15 @@
 # Build tunables
 
 # When updating these defaults, be sure to check that ALL_BUILDABLE_FLAVORS is updated
-# CEPH_VERSION,ARCH,OS_NAME,OS_VERSION,BASEOS_REG,BASEOS_REPO,BASEOS_TAG
+# CEPH_VERSION,CEPH_POINT_RELEASE, ARCH,OS_NAME,OS_VERSION,BASEOS_REG,BASEOS_REPO,BASEOS_TAG
```
leseb: remove the space after `CEPH_POINT_RELEASE,`

maint-lib/stagelib/envglobals.py
```diff
@@ -15,6 +15,7 @@
 # Allow lines to extend beyond 99 char limit by 6 chars to support output formatting to 100 cols
 REQUIRED_ENV_VARS = OrderedDict([
     ('CEPH_VERSION',      'Ceph named version being built (e.g., luminous, mimic)'),
+    ('CEPH_POINT_RELEASE', 'Ceph point release version (e.g., 12.2.0)'),
```
leseb: use a dash instead of the comma `e.g: -12.2.0`, and highlight the fact we need a dash depending on the distro.

src/__DOCKERFILE_TRACEABILITY_LABELS__
```diff
@@ -12,3 +12,6 @@ LABEL GIT_COMMIT="STAGE_REPLACE_WITH_GIT_COMMIT"
 
 # Was the repository clean when building ?
 LABEL GIT_CLEAN="STAGE_REPLACE_WITH_GIT_CLEAN"
+
+# What CEPH_POINT_RELEASE have been used ?
```
leseb: `s/have/has/`

BlaineEXE:
I'm not convinced we should add this feature without a use-case. If no one is using it, do we really need it, or are we over-engineering something?

Also, I have added a PR to change the flavor specification. If we do move forward with this, I would suggest that instead of adding another comma-delimited field to this, we allow the ceph-version to be optionally augmented with a version via a hyphen like:
 - Normal: `luminous,centos,7`
 - Augmented: `luminous-12.2.2,centos,7`

ErwanAliasr1:
@BlaineEXE That syntax would mean splitting the field into parts and would make it more difficult to substitute in the file, isn't it ?

ErwanAliasr1:
About the usage, that will be useful for debugging & testing and especially the upgrade scenario. Looks like this feature have some interest for @liewegas

BlaineEXE:
> That syntax would mean splitting the field into parts and would make it more difficult to substitute in the file, isn't it ?

I think it's a matter of parsing. I'd rather spend more code to parse than to make the flavor spec more confusing by having a double-comma required for most flavors. I changed the flavor spec initially because @leseb was finding it hard to remember to do the underscore. I think the double comma would be just as confusing.

As for whether to do this at all, I'd like to see that we have someone who is definitely going to use it before it's implemented. If @liewegas needs it, then that can be our someone. I'm concerned that adding this is adding to what we must support in our test matrices.

JrCs:
I think it's a good feature, so we can rebuild a specific ceph-container/ceph version and to not be force to use the lastest ceph version. I already made my own version of ceph-container (with ceph provisionners binaries in it) and i must have an option to choose which version of ceph i want to integrate in it.

leseb:
+1 here, we should force a user to use the latest version of Ceph if they use the latest stable tag. Of course, this makes our matrix bigger but at least operators have a better control of their Ceph versions. We can work on various latest tags for stable releases.

BlaineEXE:
I was talking to erwan about this. This feature starts adding
conditionality to the templating. I think once we start getting into
conditional replacements, we should think about a fully-fledged templating
engine. I did a really quick test on my own to see if we could use jinja2,
and I think it'll be a relatively easy replacement. I suggest that we do
this to achieve the technical means.

Excuse my brevity. I'm replying via email at dinner. 🙂

leseb:
@BlaineEXE so we put this one on hold and wait agreement on https://github.com/ceph/ceph-container/pull/981?

BlaineEXE:
I think switching to jinja, at least as I was able to get it working, would be a little too much work for a little too little gain. As for tagging point releases, openSUSE doesn't really support it, nor do I need the feature. And for Ubuntu, it's also not supported. I think maybe this PR is fine, but we should document that it is not all-encompassing.

So if we use the flavor `luminous-12.2.2,centos,7`, the `__CEPH_BASE_PACKAGES__` file should show packages like
```
ca-certificates \
ceph-common-12.2.2 \
ceph-mon-12.2.2 \
... etc ...
```
And if the flavor is `luminous,centos,7` with no spec, the file should show packages like normal
```
ca-certificates \
ceph-common \
ceph-mon \
... etc ...
```
I think this was part of the @ErwanAliasr1's original plan. I just want to make sure nothing has changed.

I think the `__CEPH_BASE_PACKAGES__` file should look like below. Side note that I like the thought of changing `STAGE_REPLACE_WITH_POINT_RELEASE` to `__ENV_[CEPH_POINT_RELEASE]__`, as I think it's clearer to read the template that all variables have the same format (double underscores) and there is a special format to give environment variables. It does look weird to have 2 variables side-by-side b/c of the underscores, but it should still parse properly. I think we should add the point release spec to this file only and not to the RADOSGW or CEPH_MGR files so that different distros can safely override these files with only package names (e.g., `ceph-rgw` or whatever) instead of having to also override the point release text.
```
\
        ca-certificates \
        ceph-common__ENV_[CEPH_POINT_RELEASE]__ \
        ceph-mon__ENV_[CEPH_POINT_RELEASE]__ \
        ceph-osd__ENV_[CEPH_POINT_RELEASE]__ \
        ceph-mds__ENV_[CEPH_POINT_RELEASE]__ \
        __CEPH_MGR_PACKAGE____ENV_[CEPH_POINT_RELEASE]__ \
        kmod \
        __RADOSGW_PACKAGE____ENV_[CEPH_POINT_RELEASE]__ \
        rbd-mirror \
        __GANESHA_PACKAGES__
```

BlaineEXE:
I just wrote a big comment, and it got lost. The TL;DR is that I think this PR is actually fine. Pinning to point releases can only be done on CentOS for the current distros. I don't need it in openSUSE, and Ubuntu doesn't support it. Since CentOS doesn't need 2 different files (one with and one without the point release text) I think it might not be that important to nitpick over this.

For `__CEPH_BASE_PACKAGES__` file, I have a few thoughts:
1. I think (see #981) it makes more sense for `STAGE_REPLACE_WITH_X` to be renamed to something like `__ENV_[X]__` so that there is a stronger separation between an environment variable insertion and other text. `ca-certificatesSTAGE_REPLACE_WITH_CEPH_VERSION` doesn't read as easily as `ca-certificates__ENV_[CEPH_VERSION]__`.
2. I think we should include the point release spec in this file and not in the `__CEPH_MGR_PACKAGE__` or `__RADOSGW_PACKAGE__` files so that a user does not (or we do not) forget to override the package in the file. So both of these files can stay the same for this PR.
```
\
        ca-certificates__ENV_[CEPH_POINT_RELEASE]__  \
        ceph-common__ENV_[CEPH_POINT_RELEASE]__  \
        ceph-mon__ENV_[CEPH_POINT_RELEASE]__  \
        ceph-osd__ENV_[CEPH_POINT_RELEASE]__  \
        ceph-mds__ENV_[CEPH_POINT_RELEASE]__  \
        __CEPH_MGR_PACKAGE____ENV_[CEPH_POINT_RELEASE]__  \
        kmod \
        __RADOSGW_PACKAGE____ENV_[CEPH_POINT_RELEASE]__  \
        rbd-mirror \
        __GANESHA_PACKAGES__
```

ErwanAliasr1:
About your 2. item :  Having " __RADOSGW_PACKAGE____ENV_[CEPH_POINT_RELEASE]__" would fail if CEPH_POINT_RELEASE is defined but not RADOSGW_PACKAGE. That's why the versionning should be inside the  __RADOSGW_PACKAGE__

ErwanAliasr1:
As for zypper, dpkg supports versionning as in "apt-get install package=version".

So we have 2 syntaxes : package-version and package=version.

I don't feel very satisfied if we have a feature which is only supported/compatible with a single set of distro (centos / rhel).

BlaineEXE:
> About your 2. item : Having " RADOSGW_PACKAGE____ENV[CEPH_POINT_RELEASE]_" would fail if CEPH_POINT_RELEASE is defined but not RADOSGW_PACKAGE. That's why the versionning should be inside the RADOSGW_PACKAGE

Good catch. 
... I don't really love this option since it's easy for a flavor to forget to include it in a package name ... but as long as the OS packages are built with proper dependencies, the package manager solver should be able to figure out that `ceph-v12.2.2`'s only supported `radosgw` package is `radosgw-v12.2.2`. So maybe this isn't actually a problem. =)

>As for zypper, dpkg supports versionning as in "apt-get install package=version".
So we have 2 syntaxes : package-version and package=version.
I don't feel very satisfied if we have a feature which is only supported/compatible with a single set of distro (centos / rhel).

The syntax would be supported, but Ubuntu's repos, and openSUSE's repos don't have old packages, meaning versioning isn't really useful. It's not that ceph-container doesn't support it but that the OS repos don't.

Side notes: When issuing `zypper info ceph=12.2.2`, zypper outputs failure, but returns with a zero exit code, so pinning on opensuse won't break things, but it will prevent nice package output. And it potentially will fail to build if a package matching a given version can't be found in the repos. I'm notsure about Ubuntu. Also, I think zypper can specify packages by `>`, `<`, `>=`, and `<=` as well as `=`. 

Update:
Pinning on opensuse would work fine if the entire, obnoxiously long version name is given as `package-VERS`. `zypper in ceph-13.0.1.3204+g17e0216271-1.1` on Tumbleweed works perfectly, for example.

Final update:
I have no idea how any of this is affected on Ubuntu. I'm assuming it's relatively the same.

ErwanAliasr1:
So I've been updating my PR so that the release is encoded with the ceph version so it doesn't requires a new variable.
The __UNVERSIONNED_PACKAGES_ are automatically generated and STAGE_REPLACE_WITH_CEPH_POINT_RELEASE are removed from them. Note that I renamed some variable so it ends with __PACKAGES_ so the magic replace occurs.
This way it's possible to get versionned & unversionned packages.
That's clearly a workaround but pretty small and very few will use it. That said, both "-" and "=" separator are supported so if we do the reprepo being updated on ceph.com that feature would be easy to use with ubuntu or even suse if needed.
The documentation is updated accordingly.

BlaineEXE:
I really don't like the magic replace, even if it is in the documentation. This feels like feature creep, and if we go down this path, we will keep going down it until we have a fully-featured and cobbled together replacement engine.

I think it is sufficient just to support packages in the format `package-version` with the dash and limit it to that since we don't need to add magic replacement/generation of files, and it will still work for pinning a specific release.

ErwanAliasr1:
To be rebased since #982

maint-lib/stagelib/filetools.py
```diff
@@ -70,6 +71,12 @@ def copy_files(filenames, src_path, dst_path, files_copied={}):
         logging.info(COPY_LOGTEXT.format(dirty_marker + ' ' + file_path, dst_path))
         files_copied[os.path.join(dst_path, f)] = dirty_marker + ' ' + file_path
         _copy_file(file_path, dst_path)
+        if f.endswith("_PACKAGES__"):
```
BlaineEXE: All of this code is interleaved into existing code. It make it harder to read/modify all of the code later. We should do this in a separate step. I think we copy files into staging as normal, and then in staging, look for files matching `_PACKAGE` (to support `__RADOSGW_PACKAGE__`, for exaple) or `_PACKAGES`, and then make a copy of it there named `_UNVERSIONED_PACKAGE` or `_UNVERSIONED_PACKAGES` without the `X_POINT_RELEASE` env replacment. We also need to replace any reference to `_PACKAGES` or `_PACKAGE` with `_UNVERSIONED_PACKAGE` or `_UNVERSIONED_PACKAGES` respectively.

This way the normal file replacement doesn't have to change, and all of this code can be viewed and edited separately from the rest.

maint-lib/stagelib/envglobals.py
```diff
@@ -48,7 +48,18 @@ def _verifyRequiredEnvVar(varname):
         sys.stderr.write(_NOT_SET_TEXT.format(varname))
         sys.stderr.write(_REQUIRED_VAR_TEXT.format(varname))
         sys.exit(1)
+    # We need to extract a possible CEPH_POINT_RELEASE from CEPH_VERSION
+    # It could be under "package_name-package_version" or "package_name=package_version"
+    if varname == "CEPH_VERSION":
+        os.environ["CEPH_POINT_RELEASE"] = " "
```
BlaineEXE: This function is a verify function and should not modify variables. I am opposed to doing this in Python, but if you must, follow the formula I have followed before with `exportGitInfoEnvVars`.

Also, do not forget to add a test for this new functionality into `tests/stage-test`

maint-lib/stagelib/replace.py
```diff
@@ -22,7 +22,11 @@
 # Read text from the file
 def _get_file_text(file_path):
     with open(file_path, 'r') as txtfile:
-        return txtfile.read()
+        try:
+            return txtfile.read()
+        except Exception:
```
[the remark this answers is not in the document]
BlaineEXE: There is a pattern already in place for this:
```py
from stagelib.filetools import IOOSErrorGracefulFail
try:
    # thing that can throw IOError and OSError
except (OSError, IOError) as o:
    IOOSErrorGracefulFail(o, "fail message")
```

BlaineEXE:
@ErwanAliasr1 
I don't think creating a special `_UNVERSIONED_PACKAGES` is necessary for any of our current OSes.

On CentOS:
```
yum install ceph-12.2.2`
rpm -q ceph-12.2.2
```

On openSUSE:
```
zypper install ceph-12.2.2
zypper info ceph-12.2.2
```

On Ubuntu/Debian:
```
apt-get install ceph=12.2.2
apt-cache show ceph=12.2.2
```

Example Ubuntu output:
```
> apt-cache show gzip=1.6-4ubuntu1
Package: gzip
Essential: yes
Status: install ok installed
Priority: required
Section: utils
Installed-Size: 240
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: amd64
Version: 1.6-4ubuntu1
Depends: dpkg (>= 1.15.4) | install-info
Pre-Depends: libc6 (>= 2.17)
Suggests: less
Description: GNU compression utilities
 This package provides the standard GNU file compression utilities, which
 are also the default compression tools for Debian.  They typically operate
 on files with names ending in '.gz', but can also decompress files ending
 in '.Z' created with 'compress'.
Description-md5: 100720c9e2c6508f1a1f3731537b38e5
Original-Maintainer: Bdale Garbee <bdale@gag.com>
```

As long as Ubuntu/Debian use `apt-cache show pkg=version`, there is no need to keep any package names without versions.

ErwanAliasr1:
@BlaineEXE  "the package manager solver should be able to figure out that ceph-v12.2.2's only supported radosgw package is radosgw-v12.2.2. So maybe this isn't actually a problem." ... That what I though too but by testing it, if we don't define it everywhere, it doesn't works fine :(

maint-lib/stagelib/replace.py
```diff
@@ -75,7 +79,11 @@ def _env_var_replace(template_text, file_path):
     for var in variable_matches:
         var_name = var[len('STAGE_REPLACE_WITH_'):]  # strip STAGE_REPLACE_WITH_
         try:
-            var_value = os.environ[var_name]
+            # If we have unversionned packages we shall not replace STAGE_REPLACE_WITH_CEPH_POINT_RELEASE
+            if "_UNVERSIONNED_PACKAGES__" in file_path and var_name == "CEPH_POINT_RELEASE":
+                var_value = ""
+            else:
```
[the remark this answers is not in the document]
ErwanAliasr1: I'm sorry but if the if match, the var_value = os.environ will be executed too, and var_value will be overwriten.

ErwanAliasr1:
@BlaineEXE so that's a headshot right ? Most of the code of this PR is unecessary. Will simplify it accordingly then.

maint-lib/stagelib/replace.py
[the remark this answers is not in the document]
ErwanAliasr1: Ok fixed this way. thx

maint-lib/stagelib/envglobals.py
```diff
@@ -48,7 +48,18 @@ def _verifyRequiredEnvVar(varname):
         sys.stderr.write(_NOT_SET_TEXT.format(varname))
         sys.stderr.write(_REQUIRED_VAR_TEXT.format(varname))
         sys.exit(1)
+    # We need to extract a possible CEPH_POINT_RELEASE from CEPH_VERSION
+    # It could be under "package_name-package_version" or "package_name=package_version"
+    if varname == "CEPH_VERSION":
```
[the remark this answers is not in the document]
ErwanAliasr1: It's much more easier to handle this specific case in python than in makefile

maint-lib/stagelib/envglobals.py
ErwanAliasr1: I will

maint-lib/stagelib/envglobals.py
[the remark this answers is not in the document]
BlaineEXE: If `CEPH_POINT_RELEASE` is going to magically appear in staging, please add it to the `/homesrc/STAGING_ENV_VARS.md` doc, and make sure to add the new computed var into the `tests/stage-test/src/daemon/computed-vars` file.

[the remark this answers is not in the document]
ErwanAliasr1: @BlaineEXE CEPH_POINT_RELEASE is a much magic as HOST_ARCH and ARCH ;o)
I updated the testing code.

maint-lib/stagelib/envglobals.py
ErwanAliasr1: Good catch.
Updated by creating populateCephPointRelease() which take care of it and so verifyRequiredEnvVars is untouched and the printing done in the proper order.

README.md
```diff
@@ -64,6 +64,28 @@ Flavors can be built in parallel easily with the `build.parallel` target:
...
 
+### Building with a specific version of Ceph
+Centos builds can select a specific version of Ceph to install. You just have to pad the required version to the ceph release code name.
+
+The version will be saved in an implicit ``CEPH_POINT_RELEASE`` variable (including the version separator) and the ``CEPH_VERSION`` is filtered out of the point release. The implicit ``CEPH_POINT_RELEASE`` variable could be used like any other environment variables. This variable is empty if no point release is given.
+
+```
+make FLAVORS=luminous-12.2.2,x86_64,centos,7,_,centos,7 build
+
+CEPH_POINT_RELEASE: -12.2.2
```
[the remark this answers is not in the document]
ErwanAliasr1: Fixed since populateCephPointRelease() does exist. Thanks.

ErwanAliasr1:
@BlaineEXE Do you ack this PR since the latest updates I pushed ?

README.md
```diff
@@ -64,6 +64,20 @@ Flavors can be built in parallel easily with the `build.parallel` target:
...
+CEPH_VERSION      : luminous
+ARCH              : x86_64
```
BlaineEXE: This will need to be HOST_ARCH after #979 is merged

maint-lib/stagelib/envglobals.py
```diff
@@ -15,6 +15,7 @@
 # Allow lines to extend beyond 99 char limit by 6 chars to support output formatting to 100 cols
 REQUIRED_ENV_VARS = OrderedDict([
     ('CEPH_VERSION',      'Ceph named version being built (e.g., luminous, mimic)'),
+    ('CEPH_POINT_RELEASE', 'Magically extract from CEPH_VERSION and point to a particular release of Ceph'),
```
BlaineEXE: The required vars are for inputs only and not generated vars. `CEPH_POINT_RELEASE` doesn't belong here.

src/STAGING_ENV_VARS.md
```diff
@@ -3,6 +3,7 @@ See `ceph-container/maint-lib/stagelib/envglobals.py` for most updated list and
 
 ## Required as input to staging
  - CEPH_VERSION
+ - CEPH_POINT_RELEASE
```
BlaineEXE: `CEPH_P_R` is not required as input (as much as I can tell). It is generated and so belongs in a below section. I would suggest 

```
## Generated during staging
### Ceph version
 - CEPH_POINT_RELEASE

File: maint-lib/stagelib/envglobals.py [hunk 13]
```diff
@@ -49,6 +50,7 @@ def _verifyRequiredEnvVar(varname):
         sys.stderr.write(_REQUIRED_VAR_TEXT.format(varname))
         sys.exit(1)
     varval = os.environ[varname]
+
```
BlaineEXE: Why add a space here? Accidental?

tests/stage-test/src/daemon/computed-vars
```diff
@@ -3,3 +3,5 @@ Git branch: __ENV_[GIT_BRANCH]__
 Git commit: __ENV_[GIT_COMMIT]__
 Git clean: __ENV_[GIT_CLEAN]__
 Go arch: __ENV_[GO_ARCH]__
+RELEASE: __ENV_[RELEASE]__
```
BlaineEXE: No need to put `RELEASE` here. This test is for computed vars only.

tests/stage-test/stage-key/staging_output.txt
```diff
@@ -1,5 +1,6 @@
 
   CEPH_VERSION      : luminous
+  CEPH_POINT_RELEASE:
```
BlaineEXE: Best to specify the CEPH_VERSION as `luminous-1.2.3-1` or something so we test that point releases are working properly in staging.

maint-lib/stage.py
```diff
@@ -45,10 +45,26 @@
...
+                version, point_release = os.environ["CEPH_VERSION"].split(separator)
+                os.environ["CEPH_VERSION"] = version
+                os.environ["CEPH_POINT_RELEASE"] = "{}{}" . format(separator, point_release)
+
+
 def main(CORE_FILES_DIR, CEPH_RELEASES_DIR):
     logging.info('\n\n\n')  # Make it easier to determine where new runs start
     logging.info('Start time: {}'.format(time.ctime()))
 
+    populateCephPointRelease()
```
BlaineEXE: This should go after `verifyRequiredEnvVars()`

maint-lib/stagelib/envglobals.py
```diff
@@ -15,6 +15,7 @@
 # Allow lines to extend beyond 99 char limit by 6 chars to support output formatting to 100 cols
 REQUIRED_ENV_VARS = OrderedDict([
     ('CEPH_VERSION',      'Ceph named version being built (e.g., luminous, mimic)'),
```
BlaineEXE: The doc for `CEPH_VERSION` needs to change to make note that it is allowed to be `luminous` or `luminous-1.2.3-1` point release. I think we **definitely** want to avoid changing what an input variable is set to during staging, so we should use `CEPH_VERSION` to compute 2 new vars: `CEPH_VERSION_NAME` (which replaces the old `CEPH_VERSION`) and `CEPH_POINT_RELEASE`.

maint-lib/stage.py
```diff
@@ -45,10 +45,26 @@
     sys.exit(1)
 
 
+def populateCephPointRelease():
```
BlaineEXE: Move to `envglobals.py` with the other computed variables. rename to `exportCephVersionSpecifics()`.

maint-lib/stage.py
```diff
@@ -45,10 +45,26 @@
...
+    # We need to extract a possible CEPH_POINT_RELEASE from CEPH_VERSION
```
BlaineEXE: Turn into proper pydoc using triple-quotes
```py
"""
Extract Ceph version specifics from `CEPH_VERSION="version-point.release.1.2.3-1"
Sets 'CEPH_VERSION_NAME=version` and `CEPH_POINT_RELEASE='-point.release.1.2.3-1'`
"""

Comment (BlaineEXE) [reply maint-lib/stage.py]: This also reminds me that we should make sure we handle the case where the point release text has dashes in it.

File: maint-lib/stage.py [hunk 20]
```diff
@@ -45,10 +45,26 @@
...
+    # It could be under the two following syntax :
+    # - package_name-package_version
+    # - package_name=package_version"
+    os.environ["CEPH_POINT_RELEASE"] = ""
+    if os.environ["CEPH_VERSION"]:
+        for separator in ["-", "="]:
+            if separator in os.environ["CEPH_VERSION"]:
+                version, point_release = os.environ["CEPH_VERSION"].split(separator)
```
BlaineEXE: This may split into more than 2 pieces if the point release spec has dashes (e.g., `ceph-12.2.2-5` which would indicate ceph v12.2.2 build 5). This should handle that and assume that the version name is before the first dash and that everything after it belongs to `point_release`.

```py
>>> vers
'luminous-12.2.2-5'
>>> vers.split('-')
['luminous', '12.2.2', '5']
>>> v, p = vers.split('-')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: too many values to unpack (expected 2)
```

BlaineEXE:
```py
>>> splitvers = vers.split('-')
>>> v = splitvers[0]
>>> v
>>> p = '-'.join(splitvers[1:])
>>> p
'12.2.2-5'
```

maint-lib/stage.py
```diff
@@ -45,10 +45,26 @@
     sys.exit(1)
 
 
+def populateCephPointRelease():
+    # We need to extract a possible CEPH_POINT_RELEASE from CEPH_VERSION
+    # It could be under the two following syntax :
+    # - package_name-package_version
+    # - package_name=package_version"
+    os.environ["CEPH_POINT_RELEASE"] = ""
+    if os.environ["CEPH_VERSION"]:
```
BlaineEXE: Use the preexisting method `getEnvVar` to both get the env var and throw error if it doesn't exist. Suggest looking at and modeling this method after `exportGoArchEnvVar()`

maint-lib/stagelib/envglobals.py
ErwanAliasr1: Yes & no. As CEPH_POINT_RELEASE is used in the base code, it have to exist. It's kind a way to double-check this variable have been created properly as all important ones.

src/STAGING_ENV_VARS.md
ErwanAliasr1: Right.

tests/stage-test/src/daemon/computed-vars
ErwanAliasr1: Ok

tests/stage-test/stage-key/staging_output.txt
ErwanAliasr1: Good idea.

maint-lib/stage.py
ErwanAliasr1: No as it does populate the C_P_R

maint-lib/stage.py
ErwanAliasr1: Ok

maint-lib/stage.py
ErwanAliasr1: Fixed.

maint-lib/stage.py
ErwanAliasr1: Ok. Fixed this way.

maint-lib/stagelib/envglobals.py
BlaineEXE: This section is for verifying inputs only. If it needs verified, then it should be verified in the `exportCephVersionSpecifics()`. Alternately, note that the `getEnvVar()` function verifies the existence itself.

maint-lib/stagelib/envglobals.py
```diff
@@ -105,3 +106,19 @@ def exportGoArchEnvVar():
...
+     - package_name-package_version
+     - package_name=package_version"
+    """
+    os.environ["CEPH_POINT_RELEASE"] = ""
+    original_ceph_version = getEnvVar("CEPH_VERSION")
+    if original_ceph_version:
+        for separator in ["-", "="]:
+            if separator in original_ceph_version:
+                splitted = original_ceph_version.split(separator)
+                os.environ["CEPH_VERSION"] = splitted[0]
```
BlaineEXE: The code should not be reassigning user-specified environment variables. That would be very confusing to users to have a variable they personally set change values during runtime. We need to extract this into 2 parts.

maint-lib/stage.py
```diff
@@ -49,6 +50,8 @@ def main(CORE_FILES_DIR, CEPH_RELEASES_DIR):
     logging.info('\n\n\n')  # Make it easier to determine where new runs start
     logging.info('Start time: {}'.format(time.ctime()))
 
+    exportCephVersionSpecifics()
```
BlaineEXE: Should go after `verifyRequiredEnvVars()`.

Also, since still reworking, following the pattern set by GoArch and GitInfo, this should be renamed to something like `exportCephVersionSpecificEnvVars()`

README.md
```diff
@@ -64,6 +64,20 @@ Flavors can be built in parallel easily with the `build.parallel` target:
 make FLAVORS="<flavor> <other flavor> <...>" build.parallel
 ```
 
+### Building with a specific version of Ceph
+Centos builds can select a specific version of Ceph to install. You just have to pad the required version to the ceph release code name.
+
+The version will be saved in an implicit ``CEPH_POINT_RELEASE`` variable (including the version separator) and the ``CEPH_VERSION`` is filtered out of the point release. The implicit ``CEPH_POINT_RELEASE`` variable could be used like any other environment variables. This variable is empty if no point release is given.
+
+```
+make FLAVORS=luminous-12.2.2,x86_64,centos,7,_,centos,7 build
```
BlaineEXE: Version spec has changed.

README.md
```diff
@@ -64,6 +64,20 @@ Flavors can be built in parallel easily with the `build.parallel` target:
 make FLAVORS="<flavor> <other flavor> <...>" build.parallel
 ```
 
+### Building with a specific version of Ceph
```
BlaineEXE: Please follow exising readme style, split lines to 100 chars or less, and use single grave-accents.

maint-lib/stagelib/envglobals.py
```diff
@@ -105,3 +106,19 @@ def exportGoArchEnvVar():
...
+                os.environ["CEPH_POINT_RELEASE"] = "{}" . format(separator.join(splitted[1:]))
```
BlaineEXE: Can add `_verifyRequiredEnvVar('CEPH_POINT_RELEASE')` and `_verifyRequiredEnvVar('CEPH_VERSION_NAME')` here to print and log the values.

README.md
```diff
@@ -64,6 +64,20 @@ Flavors can be built in parallel easily with the `build.parallel` target:
...
+Centos builds can select a specific version of Ceph to install. You just have to pad the required version to the ceph release code name.
```
BlaineEXE: Doc should probably be rephrased to not mention centos specifically.  Maybe something like.
```
Builds can specifiy a specific point release version of Ceph to install by appending the package-manager-appropriate version to the ceph version. Only `-` and `=` are supported as specifiers.

E.g., `luminous-12.2.2` (for yum/zypper) or `luminous=12.2.2` (for apt)

The `CEPH_VERSION` will be interpreted during staging into 2 new environment variables:
- `CEPH_VERSION_NAME` will get the ceph version without point release text
- `CEPH_POINT_RELEASE` will get the point release text including the `-` or `=`
```

maint-lib/stagelib/envglobals.py
ErwanAliasr1: I forgot that one. I updated the documentation.
CEPH_VERSION will always be luminous and CEPH_POINT_RELEASE will be empty or set.
So for a user point of view, the semantic is always the same.
I do agree about the confusion that it could create since we rewrite CEPH_VERSION but that reduce the number of variables to handle/support.

Makefile
```diff
@@ -158,13 +158,14 @@ help:
...
 	@echo '  FLAVORS - ceph-container images to operate on in the form'
-	@echo '    <CEPH_VERSION>,<BASEOS_REPO>,<BASEOS_TAG>'
+	@echo '    <CEPH_VERSION>[CEPH_POINT_RELEASE],<BASEOS_REPO>,<BASEOS_TAG>'
 	@echo '    and multiple forms may be separated by spaces.'
 	@echo '      CEPH_VERSION - named ceph version (e.g., luminous, mimic)'
-	@echo '      BASEOS_REPO - The base image to use for the daemon-base container. This is also'
-	@echo '                    the distro path sourced from ceph-container (e.g., ubuntu, centos)'
-	@echo '      BASEOS_TAG - Tagged version of the base repo to use. Also the distro version'
-	@echo '                   sourced from ceph-container (e.g., ubuntu:"16.04", centos:"7")'
+	@echo '      CEPH_POINT_RELEASE - Optional field to select a particular version if Ceph (e.g -12.2.2)'
```
BlaineEXE: I would add more clarification that dash '-' and equal '=' are the supported point release separators and which one to use should be based on the package manager used in the image. Dash for yum/zypper. Equal for apt.

Makefile
```diff
@@ -158,13 +158,14 @@ help:
...
+	@echo '      BASEOS_REPO  - The base image to use for the daemon-base container. This is also'
+	@echo '                      the distro path sourced from ceph-container (e.g., ubuntu, centos)'
+	@echo '      BASEOS_TAG   - Tagged version of the base repo to use. Also the distro version'
+	@echo '                      sourced from ceph-container (e.g., ubuntu:"16.04", centos:"7")'
 	@echo '    e.g., FLAVORS="luminous,ubuntu,16.04 jewel,ubuntu,14.04"'
```
BlaineEXE: It looks like I forgot to modify this in my short-flavors PR.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,39 @@
...
+WHAT_TO_EXTRACT=$2
+
+# Search for the two possible separator between CEPH_VERSION and CEPH_POINT_RELEASE
+# Let's consider CEPH_VERSION_SPEC=luminous-12.2.0-1
+for separator in "=" "-"; do
+  # If the line doesn't have a separator, let's try the next one
+  echo "${CEPH_VERSION_SPEC}" | grep -q "${separator}" || continue
+  case "${WHAT_TO_EXTRACT}" in
+    "CEPH_POINT_RELEASE")
+      # i.e Extracting -12.2.0-1 from CEPH_VERSION_SPEC
```
BlaineEXE: "e.g.", not "i.e."

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
```
BlaineEXE: Add `set -euo pipefail` here so that script fails on errors, unset variables, and failures within piped commands.

Makefile
ErwanAliasr1: Added.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,39 @@
...
+CEPH_VERSION_SPEC=$1
+WHAT_TO_EXTRACT=$2
+
+# Search for the two possible separator between CEPH_VERSION and CEPH_POINT_RELEASE
+# Let's consider CEPH_VERSION_SPEC=luminous-12.2.0-1
+for separator in "=" "-"; do
+  # If the line doesn't have a separator, let's try the next one
+  echo "${CEPH_VERSION_SPEC}" | grep -q "${separator}" || continue
```
BlaineEXE: When possible, bash constructs should be used over outside tools and pipes. In this case, it is easier and clearer to write:
```sh
if [[ ! "${CEPH_VERSION_SPEC}" =~ "${separator}" ]]; then continue; fi
```

Makefile
ErwanAliasr1: Yup, I took it no worries ;)

maint-lib/ceph_version.sh
ErwanAliasr1: fixed.

CONTRIBUTING.md
```diff
@@ -176,12 +176,12 @@ flavors for the new Ceph version to the Makefile. At minimum, `ALL_BUILDABLE_FLA
 updated in the Makefile. If distro source is properly configured to support multiple Ceph releases
 and there are no special updates required, they are likely to work with just this minimal change.
 
-Note the `$CEPH_VERSION` variable usually used in `__DOCKERFILE_INSTALL__` is substituted from the
-first field of the flavor name by environment variable substitution.
+Note the `$CEPH_VERSION` and `$CEPH_POINT_RELEASE` variables usually used in `__DOCKERFILE_INSTALL__`
+are extracted from the first field of the flavor name.
 
-In this example, `luminous,x86_64,centos,7,_,centos,7`, `$CEPH_VERSION` will be set to **luminous**.
+In this example, `luminous,centos,7`, `$CEPH_VERSION` will be set to **luminous**.
```
BlaineEXE: Add an example with point release var for readers?

Makefile
```diff
@@ -158,13 +158,17 @@ help:
...
-	@echo '    <CEPH_VERSION>,<BASEOS_REPO>,<BASEOS_TAG>'
+	@echo '    <CEPH_VERSION>[CEPH_POINT_RELEASE],<BASEOS_REPO>,<BASEOS_TAG>'
 	@echo '    and multiple forms may be separated by spaces.'
 	@echo '      CEPH_VERSION - named ceph version (e.g., luminous, mimic)'
-	@echo '      BASEOS_REPO - The base image to use for the daemon-base container. This is also'
-	@echo '                    the distro path sourced from ceph-container (e.g., ubuntu, centos)'
-	@echo '      BASEOS_TAG - Tagged version of the base repo to use. Also the distro version'
-	@echo '                   sourced from ceph-container (e.g., ubuntu:"16.04", centos:"7")'
+	@echo '      CEPH_POINT_RELEASE - Optional field to select a particular version of Ceph'
+	@echo '                           Regarding the package manager the version separator may vary :'
```
BlaineEXE: For a user who doesn't know that the `CEPH_POINT_RELEASE` field is directly appended to package names, this could still be confusing. It may be best just to say explicitly that the CPR value (including the separator '-' or '=') will be appended to each ceph package name to set a specific version. And that the package specification separator for yum/dnf/zypper is a dash, and apt is equal. It's verbose, but it provides necessary clarity.

README.md
```diff
@@ -64,6 +64,25 @@ Flavors can be built in parallel easily with the `build.parallel` target:
 make FLAVORS="<flavor> <other flavor> <...>" build.parallel
 ```
 
+### Building with a specific version of Ceph
+Some distributions can select a specific version of Ceph to install.
+You just have to pad the required version to the ceph release code name.
```
BlaineEXE: "pad" is not an accurate word here. Suggest: "append"

BlaineEXE: Not just the version but also the package-manager-appropriate version specifier/separator '-' or '=' must also be part of what is appended.

README.md
```diff
@@ -64,6 +64,25 @@ Flavors can be built in parallel easily with the `build.parallel` target:
...
+
+The required version will be saved in ``CEPH_POINT_RELEASE`` variable (including the version separator).
```
BlaineEXE: No double-grave's please.

README.md
```diff
@@ -64,6 +64,25 @@ Flavors can be built in parallel easily with the `build.parallel` target:
...
+The version separator is usually a dash ('-') or an equal sign ('=').
+``CEPH_POINT_RELEASE`` remains empty if no point release is given.
+
+Note that ``CEPH_VERSION`` variable still feature the ceph code name, **luminous** in this example.
```
BlaineEXE: There are many short and choppy paragraphs in this section that are hard to read. Combine them into a single paragraph so it is clear the information is part of the same concept.

README.md
```diff
@@ -64,6 +64,25 @@ Flavors can be built in parallel easily with the `build.parallel` target:
...
+
+```
```
BlaineEXE: Having output from make in the doc is a bad idea. It's not strictly necessary, and the make output may change over time, and then this easily becomes outdated.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,39 @@
...
+  case "${WHAT_TO_EXTRACT}" in
+    "CEPH_POINT_RELEASE")
+      # e.g Extracting -12.2.0-1 from CEPH_VERSION_SPEC
+      echo -n "${separator}"
+      echo "${CEPH_VERSION_SPEC}" | cut -d "${separator}" -f 2-
+      exit 0
+      ;;
+    *)
```
BlaineEXE: What to extract can only be 2 values in reality 'CEPH_POINT_RELEASE' or 'CEPH_VERSION'. But in code, the values can be 'CEPH_POINT_RELEASE' or <anything> which could be problematic. These case statements should specify only the 2 possible values, and the default option `*)` should exit with error.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,39 @@
...
+      # e.g Extracting luminous from CEPH_VERSION_SPEC
+      echo "${CEPH_VERSION_SPEC}" | cut -d "${separator}" -f 1
+      exit 0
+      ;;
+  esac
+done
+
+# If no separator got found, it means it doesn't have any
+# Let's consider CEPH_VERSION_SPEC=luminous
+case "${WHAT_TO_EXTRACT}" in
```
BlaineEXE: There are 2 case statements which have the same test. This suggests that these 2 cases can be combined together. 

Another code-flow option would be to take the `CEPH_VERSION_SPEC` input and pull it apart into it's 2 parts initially with CEPH_POINT_RELEASE getting empty string if it can't be pulled apart. Then trivially output (echo) the desired part in a single case statement that comes after. Code will be DRYer.

maint-lib/makelib.mk
```diff
@@ -5,21 +5,24 @@ HOST_ARCH ?= $(shell uname --machine)
...
+# Note that CEPH_VERSION_SPEC is split into CEPH_VERSION and CEPH_POINT_RELEASE
 # Note that this format is the same as the FLAVORS spec format with HOST_ARCH prepended; it's
 #   necessary to pass HOST_ARCH in so we can do parallel cross-builds of different arches later.
 comma := ,
 define set_env_vars
 $(shell bash -c 'set -eu ; \
 	function set_var () { export $$1="$$2" ; echo -n $$1="\"$$2\" " ; } ; \
 	set_var HOST_ARCH         "$(word 1, $(subst $(comma), ,$(1)))" ; \
-	set_var CEPH_VERSION      "$(word 2, $(subst $(comma), , $(1)))" ; \
+	CEPH_VERSION_SPEC="$(word 2, $(subst $(comma), ,$(1)))" ; \
```
BlaineEXE: Following the pattern established below, lower case `ceph_version_spec` here.

maint-lib/makelib.mk
```diff
@@ -5,21 +5,24 @@ HOST_ARCH ?= $(shell uname --machine)
...
+	set_var CEPH_VERSION   "$$(bash maint-lib/ceph_version.sh "$$CEPH_VERSION_SPEC" CEPH_VERSION)" ; \
```
BlaineEXE: This line and the line below break the right-side column alignment. Please take the time to correct the alignment, which I believe will require moving the column right a few chars since `CEPH_POINT_RELEASE` is such a long name.

maint-lib/stagelib/envglobals.py
```diff
@@ -15,6 +15,7 @@
 # Allow lines to extend beyond 99 char limit by 6 chars to support output formatting to 100 cols
 REQUIRED_ENV_VARS = OrderedDict([
     ('CEPH_VERSION',      'Ceph named version being built (e.g., luminous, mimic)'),
+    ('CEPH_POINT_RELEASE', 'Points to specific version of ceph (e.g -12.2.0) or empty'),
```
BlaineEXE: Please keep to the preexisting right-column alignment.

tests/stage-test/test_staging.sh
```diff
@@ -3,12 +3,16 @@ set -euo pipefail
 
 # Integration test to make sure staging is working properly
 
-export CEPH_VERSION=luminous
+export CEPH_VERSION_SPEC=luminous-12.2.1-0
+# shellcheck disable=SC2155
+export CEPH_VERSION=$(maint-lib/ceph_version.sh "${CEPH_VERSION_SPEC}")
```
BlaineEXE: I don't feel comfortable about ignoring SC2155 here. If you are already adding the shellcheck disable line, it is no more verbose to set CEPH_VERSION on one line and then export it after.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
BlaineEXE: Best shebang is `#!/usr/bin/env bash`

maint-lib/ceph_version.sh
ErwanAliasr1: ok

maint-lib/ceph_version.sh
ErwanAliasr1: We already have some code using grep. Is that much important ?

CONTRIBUTING.md
ErwanAliasr1: There is already one in "### Building with a specific version of Ceph" at README.md
I'd like not duplicating information too much and this part of the documentation doesn't speak about this specific case.

Makefile
ErwanAliasr1: This documentation is not about how it is implemented but what is the goal of that field. For the user, this is enough information to understand that CPR is used to pin to a particular release.

README.md
ErwanAliasr1: I disagree. First part is the why we need that, 2nd is how it works, 3rd is just a notice.
Their are all under the same topic which only address this issue. It is clear this about the same concept.

README.md
ErwanAliasr1: ok

maint-lib/ceph_version.sh
ErwanAliasr1: That was done on purpose. We'll have CEPH_VERSION unless we explicit wants 'CEPH_POINT_RELEASE'

Makefile
BlaineEXE: The user needs enough information to use the field effectively. It would be unclear when to pick `-` vs `=` for a package manager that we don't already have in the codebase unless the user understands that all ceph packages will have the field appended including what we've called the separator.

maint-lib/ceph_version.sh
BlaineEXE: This is part of shell scripting best practices. More pipes means more opportunity for errors to occur. Like if a particular version of grep is bugged for some reason. If the dependency on grep is removed, that is no longer a failure mode.

maint-lib/ceph_version.sh
BlaineEXE: https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang/10383546#10383546

maint-lib/makelib.mk
BlaineEXE: Readability and maintainability: If upper-case is used here, it may imply to the reader/maintainer that this value is supposed to become an environment variable rather than just being a local value.

maint-lib/makelib.mk
BlaineEXE: Right-side column alignment already exists to improve code readability for maintainers. Breaking readability will make maintenance harder in the future.

maint-lib/stagelib/envglobals.py
BlaineEXE: Right-side column alignment already exists to improve code readability for maintainers. Breaking readability will make maintenance harder in the future.

tests/stage-test/test_staging.sh
BlaineEXE: It is for code safety that SC2155 throws an error. Ignoring it risks errors on corner cases.

maint-lib/ceph_version.sh
ErwanAliasr1: I simplified the bash script, hope you'll like it

maint-lib/makelib.mk
ErwanAliasr1: ok

maint-lib/makelib.mk
ErwanAliasr1: oops. Fixed.

maint-lib/stagelib/envglobals.py
ErwanAliasr1: I'll move all the code to right then to avoid error 231

tests/stage-test/test_staging.sh
ErwanAliasr1: why not. doesn't make a big change.

maint-lib/ceph_version.sh
ErwanAliasr1: ok.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,29 @@
...
+# By default, we consider that CEPH_VERSION_SPEC in this style: luminous
+# So no point release, just a version
+CEPH_POINT_RELEASE=""
+CEPH_VERSION="${CEPH_VERSION_SPEC}"
+
+# Search for the two possible separators between CEPH_VERSION and CEPH_POINT_RELEASE
+# Let's consider CEPH_VERSION_SPEC=luminous-12.2.0-1
+for separator in "=" "-"; do
+  # If the line doesn't have a separator, let's try the next one
+  echo "${CEPH_VERSION_SPEC}" | grep -q "${separator}" || continue
```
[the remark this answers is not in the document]
ErwanAliasr1: Ok.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,29 @@
...
+
+  # If we found it, let's save both parts in the respective variables
+  # shellcheck disable=SC2034
+  CEPH_VERSION=$(echo "${CEPH_VERSION_SPEC}" | cut -d "${separator}" -f 1)
```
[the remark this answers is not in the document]
ErwanAliasr1: Ok for the builtin but I still need the disable as shellcheck doesn't foresee the eval as a potential usage of those variables.

maint-lib/ceph_version.sh
```diff
@@ -0,0 +1,29 @@
...
+  # shellcheck disable=SC2034
+  CEPH_POINT_RELEASE=$(echo -n "${separator}"; echo "${CEPH_VERSION_SPEC}" | cut -d "${separator}" -f 2-)
+
+  # Don't continue if we found something, it's time to print it
+  break
+done
+
+# Let's print the requested variable
+eval "echo \$$WHAT_TO_EXTRACT"
```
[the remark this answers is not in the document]
ErwanAliasr1: Didn't knew that one.

PR #979, opened 2018-03-23
make: shorten flavors

BlaineEXE:
Shorten the flavor spec to `CEPH_VERSION,BASEOS_REPO,BASEOS_TAG`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,20 +79,10 @@ build: $(foreach p, $(FLAVORS), do.image.$(p)) ;
 push:  $(foreach p, $(FLAVORS), do.image.$(p)) ;
 
 push.parallel:
-# Due to output-sync, will not output results until finished so there is no text interleaving
-ifeq (4.00,$(firstword $(sort $(MAKE_VERSION) 4.00)))
-	@$(MAKE) --jobs $(nproc) --output-sync push
-else
-	@$(MAKE) --jobs $(nproc) push
-endif
+	@$(MAKE) $(PARALLEL) push
 
 build.parallel:
-# Due to output-sync, will not output results until finished so there is no text interleaving
-ifeq (4.00,$(firstword $(sort $(MAKE_VERSION) 4.00)))
-	@$(MAKE) --jobs $(nproc) --output-sync build
-else
-	@$(MAKE) --jobs $(nproc) build
-endif
+	@$(MAKE) $(PARALLEL) build
 
 build.all:
 	@$(MAKE) FLAVORS="$(ALL_BUILDABLE_FLAVORS)" build.parallel
--- a/Makefile
+++ b/Makefile
@@ -17,15 +17,15 @@
 # Build tunables
 
 # When updating these defaults, be sure to check that ALL_BUILDABLE_FLAVORS is updated
-# CEPH_VERSION,ARCH,OS_NAME,OS_VERSION,BASEOS_REG,BASEOS_REPO,BASEOS_TAG
+# CEPH_VERSION,BASEOS_REPO,BASEOS_TAG
 FLAVORS ?= \
-	luminous,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	jewel,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	jewel,x86_64,ubuntu,14.04,_,ubuntu,14.04 \
-	kraken,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	luminous,x86_64,centos,7,_,centos,7 \
-	jewel,x86_64,centos,7,_,centos,7 \
-	kraken,x86_64,centos,7,_,centos,7 \
+	luminous,ubuntu,16.04 \
+	jewel,ubuntu,16.04 \
+	jewel,ubuntu,14.04 \
+	kraken,ubuntu,16.04 \
+	luminous,centos,7 \
+	jewel,centos,7 \
+	kraken,centos,7 \
 
 REGISTRY ?= ceph
 
@@ -39,44 +39,37 @@ RELEASE ?= $(shell git rev-parse --abbrev-ref HEAD)
 include maint-lib/makelib.mk
 
 # All flavor options that can be passed to FLAVORS
-# CEPH_VERSION,ARCH,OS_NAME,OS_VERSION,BASEOS_REG,BASEOS_REPO,BASEOS_TAG
+# CEPH_VERSION,BASEOS_REPO,BASEOS_TAG
 ALL_BUILDABLE_FLAVORS := \
-	luminous,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	jewel,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	jewel,x86_64,ubuntu,14.04,_,ubuntu,14.04 \
-	kraken,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	luminous,x86_64,centos,7,_,centos,7 \
-	jewel,x86_64,centos,7,_,centos,7 \
-	kraken,x86_64,centos,7,_,centos,7 \
-	luminous,x86_64,opensuse,42.3,_,opensuse,42.3 \
+	luminous,ubuntu,16.04 \
+	jewel,ubuntu,16.04 \
+	jewel,ubuntu,14.04 \
+	kraken,ubuntu,16.04 \
+	luminous,centos,7 \
+	jewel,centos,7 \
+	kraken,centos,7 \
+	luminous,opensuse,42.3 \
 
 # ==============================================================================
 # Build targets
 .PHONY: all stage build build.parallel build.all push
 
 stage.%:
-	@$(call set_env_var,CEPH_VERSION,$*) $(call set_env_var,ARCH,$*) \
-	$(call set_env_var,OS_NAME,$*) $(call set_env_var,OS_VERSION,$*) \
-	$(call set_env_var,BASEOS_REG,$*) $(call set_env_var,BASEOS_REPO,$*) \
-	$(call set_env_var,BASEOS_TAG,$*) $(call set_env_var,IMAGES_TO_BUILD,$*) \
-	$(call set_env_var,STAGING_DIR,$*) $(call set_env_var,BASE_IMAGE,$*) \
-	$(call set_env_var,DAEMON_BASE_IMAGE,$*) $(call set_env_var,DAEMON_IMAGE,$*) \
-	$(call set_env_var,RELEASE,$*) \
-	sh -c maint-lib/stage.py
+	@$(call set_env_vars,$*) sh -c maint-lib/stage.py
 
 daemon-base.%: stage.%
-	@$(call set_env_var,STAGING_DIR,$*) ; $(MAKE) -C $$STAGING_DIR/daemon-base $(MAKECMDGOALS) \
-	$(call set_env_var,DAEMON_BASE_IMAGE,$*)
+	@$(call set_env_vars,$*); $(MAKE) -C $$STAGING_DIR/daemon-base \
+	  $(call set_env_vars,$*) $(MAKECMDGOALS)
 
 daemon.%: daemon-base.%
-	@$(call set_env_var,STAGING_DIR,$*) ; $(MAKE) -C $$STAGING_DIR/daemon $(MAKECMDGOALS) \
-	$(call set_env_var,DAEMON_IMAGE,$*)
+	@$(call set_env_vars,$*); $(MAKE) $(call set_env_vars,$*) -C $$STAGING_DIR/daemon \
+	  $(call set_env_vars,$*) $(MAKECMDGOALS)
 
 do.image.%: daemon.% ;
 
-stage: $(foreach p, $(FLAVORS), stage.$(p)) ;
-build: $(foreach p, $(FLAVORS), do.image.$(p)) ;
-push:  $(foreach p, $(FLAVORS), do.image.$(p)) ;
+stage: $(foreach p, $(FLAVORS), stage.$(HOST_ARCH),$(p)) ;
+build: $(foreach p, $(FLAVORS), do.image.$(HOST_ARCH),$(p)) ;
+push:  $(foreach p, $(FLAVORS), do.image.$(HOST_ARCH),$(p)) ;
 
 push.parallel:
 	@$(MAKE) $(PARALLEL) push
@@ -93,9 +86,9 @@ build.all:
 .PHONY: clean clean.nones clean.all clean.nuke
 
 clean.image.%: do.image.%
-	@$(call set_env_var,STAGING_DIR,$*); rm -rf $$STAGING_DIR
+	@$(call set_env_vars); rm -rf $$STAGING_DIR
 
-clean: $(foreach p, $(FLAVORS), clean.image.$(p))
+clean: $(foreach p, $(FLAVORS), clean.image.$(HOST_ARCH),$(p))
 
 clean.nones:
 	@docker rmi -f $(shell docker images | egrep "^<none> " | awk '{print $$3}' | uniq) || true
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -112,9 +112,10 @@ clean.nones:
 
 clean.all: clean.nones
 	@rm -rf staging/
-	# Don't mess with other registries for some semblance of a safe nuke.
+	# Don't mess with other registries for some semblance of a safe clean.
 	@docker rmi -f \
-		$(shell docker images | egrep "^$(REGISTRY)/daemon(-base)? " | awk '{print $$3}' | uniq) || true
+		$(shell docker images | egrep "^$(REGISTRY)/daemon(-base)? " | \
```
ErwanAliasr1: Can you explain in what | uniq is "broken" ?
Please make the commit message explicit on that point.

Makefile
```diff
@@ -17,15 +17,15 @@
...
+# CEPH_VERSION,BASEOS_REPO,BASEOS_TAG
 FLAVORS ?= \
-	luminous,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	jewel,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	jewel,x86_64,ubuntu,14.04,_,ubuntu,14.04 \
-	kraken,x86_64,ubuntu,16.04,_,ubuntu,16.04 \
-	luminous,x86_64,centos,7,_,centos,7 \
-	jewel,x86_64,centos,7,_,centos,7 \
-	kraken,x86_64,centos,7,_,centos,7 \
+	luminous,ubuntu,16.04 \
```
ErwanAliasr1: Why does ARCH has vanished ? We need it for arm64 support.

BlaineEXE: ARCH hasn't vanished. It is computed automatically from the host arch.

Makefile
BlaineEXE: The single quotes on the `clean.nones` were misplaced. It seems the `uniq` functionality was added but not fully tested.

ErwanAliasr1: We saw it, I though it was fixed.

Makefile
ErwanAliasr1: OK

BlaineEXE: I'll have to add support for crossbuilds later. I'm not sure crossbuilds would actually have worked in the previous schema anyway.

Makefile
BlaineEXE: I see it was fixed at some point.

Makefile
ErwanAliasr1: Tried it but failed. Looks like we'll need dedicated build node and so the implicit ARCH would be ok.

BlaineEXE: I talked with Bassam a bit yesterday about this. I think it's possible to do crossbuilds via make, but it'll require work to support and make sure it works.

BlaineEXE:
I also had to edit the travis tests. I'm not sure what needs to be done to support this new format to get tests passing to github. @leseb ?

tests/tox.sh
```diff
@@ -89,14 +89,14 @@ CURRENT_CEPH_STABLE_RELEASE="$(echo $FLAVOR|awk -F ',' '{ print $1}')"
...
-  FLAVOR="$CEPH_STABLE_RELEASE,x86_64,ubuntu,16.04,_,ubuntu,16.04"
+  FLAVOR="$CEPH_STABLE_RELEASE,ubuntu,16.04"
 fi
 
 echo "Building flavor $FLAVOR"
 sudo make FLAVORS="$FLAVOR" build.parallel
 
 current_branch="$(git rev-parse --abbrev-ref HEAD)"
-# build tag from luminous,x86_64,centos,7,_,centos,7 to luminous-centos-7-x86_64
+# build tag from luminous,centos,7 to luminous-centos-7-x86_64
```
leseb: Since ` x86_64` is implicit you need to force it by detecting it.
So something like this should work:

```
TAG=$(echo "$FLAVOR"|awk -F ',' '{ print $1"-"$2"-"$3 }')-$(arch)
```

BlaineEXE:
I'm not sure why Travis is failing on `make FLAVORS=luminous,centos,7 build`. It works locally.

ErwanAliasr1:
jenkins test this please

ErwanAliasr1:
@BlaineEXE this is where using github as origin could be useful to ease reviewers jumping in the branch & test it.

tests/tox.sh
```diff
@@ -93,16 +93,15 @@ if [[ "$CEPH_STABLE_RELEASE" != "$CURRENT_CEPH_STABLE_RELEASE" ]]; then
...
+make_output=$(sudo make FLAVORS="$FLAVOR" build.parallel)
+daemon_image=$(echo "${make_output}" | grep "DAEMON_IMAGE ") # Find DAEMON_IMAGE line
+daemon_image="${daemon_image#*DAEMON_IMAGE*: }" # Remove DAEMON_IMAGE from beginning
+daemon_image="$(echo "${daemon_image}" | tr -s ' ')" # Remove whitespace
 
 # start a local docker registry
 sudo docker run -d -p 5000:5000 --restart=always --name registry registry:2
 # add the image we just built to the registry
-sudo docker tag ceph/daemon:"$TAG" localhost:5000/ceph/daemon:"$CEPH_STABLE_RELEASE"-latest
+sudo docker tag ${daemon_image} localhost:5000/ceph/daemon:"$CEPH_STABLE_RELEASE"-latest
```
ErwanAliasr1: Sloppy reports an error here :+1: 
In /tox.sh line 104:
sudo docker tag ${daemon_image} localhost:5000/ceph/daemon:"$CEPH_STABLE_RELEASE"-latest
                ^-- SC2086: Double quote to prevent globbing and word splitting.

BlaineEXE: Fixed.

BlaineEXE:
> this is where using github as origin could be useful to ease reviewers jumping in the branch & test it.

I am making it a point to only work out of the SUSE fork since that is the process we have intentionally set in place here so we can collaborate internally and so we can keep a stable branch for ourselves once we get to our release.

If you want to pull from a PR directly, GitHub makes this easier:
```
git fetch origin pull/<ID>/head:BRANCHNAME
git checkout BRANCHNAME
```
https://help.github.com/articles/checking-out-pull-requests-locally/

BlaineEXE:
There was a section in the centos/daemon-base `__DOCKERFILE_INSTALL__`: `STAGE_REPLACE_WITH_OS_NAME/STAGE_REPLACE_WITH_OS_VERSION`, which I replaced with
`STAGE_REPLACE_WITH_BASEOS_REPO/STAGE_REPLACE_WITH_BASEOS_TAG`

BlaineEXE:
It looks like when travis is doing `make build`, the `$(call set_env_vars,$*)` line isn't returning anything, so no env vars are being set. I'm not sure why. I fixed the parenthese issue by shebanging bash.

BlaineEXE:
I think I figured out that in whatever environment make is running in travis, bash is not the default shell, so it's not playing nice. I change my `set_env_vars` to explicitly call bash to prevent this, and it seems to at least start the travis build ok.

BlaineEXE:
I think this is complete. I checked `Testing: ceph_ansible-luminous-centos7-cluster`, and it looks like it may be a vagrant issue, unless there is something I missed in the log.

tests/tox.sh
```diff
@@ -93,16 +93,15 @@ if [[ "$CEPH_STABLE_RELEASE" != "$CURRENT_CEPH_STABLE_RELEASE" ]]; then
 fi
 
 echo "Building flavor $FLAVOR"
-sudo make FLAVORS="$FLAVOR" build.parallel
-
-current_branch="$(git rev-parse --abbrev-ref HEAD)"
-# build tag from luminous,centos,7 to luminous-centos-7-x86_64
-TAG="${current_branch}-$(echo "$FLAVOR"|awk -F ',' '{ print $1"-"$3"-"$4"-"$2}')"
+make_output=$(sudo make FLAVORS="$FLAVOR" build.parallel)
+daemon_image=$(echo "${make_output}" | grep "DAEMON_IMAGE ") # Find DAEMON_IMAGE line
```
ErwanAliasr1: @leseb Do you agree on this change ?

leseb: Nevermind, got it.

tests/tox.sh
```diff
@@ -89,20 +89,19 @@ CURRENT_CEPH_STABLE_RELEASE="$(echo $FLAVOR|awk -F ',' '{ print $1}')"
...
 fi
 
 echo "Building flavor $FLAVOR"
-sudo make FLAVORS="$FLAVOR" build.parallel
-
-current_branch="$(git rev-parse --abbrev-ref HEAD)"
-# build tag from luminous,x86_64,centos,7,_,centos,7 to luminous-centos-7-x86_64
-TAG="${current_branch}-$(echo "$FLAVOR"|awk -F ',' '{ print $1"-"$3"-"$4"-"$2}')"
+make_output=$(sudo make FLAVORS="$FLAVOR" build.parallel)
+daemon_image=$(echo "${make_output}" | grep "DAEMON_IMAGE ") # Find DAEMON_IMAGE line
```
ErwanAliasr1: grep "^DAEMON_IMAGE" would be safer

ErwanAliasr1: This seems to be the last change before merge.

tests/tox.sh
```diff
@@ -89,20 +89,19 @@ CURRENT_CEPH_STABLE_RELEASE="$(echo $FLAVOR|awk -F ',' '{ print $1}')"
...
+  FLAVOR="$CEPH_STABLE_RELEASE,ubuntu,16.04"
 fi
 
 echo "Building flavor $FLAVOR"
-sudo make FLAVORS="$FLAVOR" build.parallel
-
-current_branch="$(git rev-parse --abbrev-ref HEAD)"
-# build tag from luminous,x86_64,centos,7,_,centos,7 to luminous-centos-7-x86_64
-TAG="${current_branch}-$(echo "$FLAVOR"|awk -F ',' '{ print $1"-"$3"-"$4"-"$2}')"
+make_output=$(sudo make FLAVORS="$FLAVOR" build.parallel)
```
leseb: The problem I've seen with this approach is that when running the output is hidden in the variable, once done, the `-x` takes effects and show the content of the variable `make_output`. What's a bit troublesome is that if the command fails, you will get the error right away and then you will get the full output. This makes debugging a bit harder as we have to scroll all the way up to find the root cause.
Without the output inside a variable, you will get the error right at the end.

So now, I'm a bit hesitant with the new approach, what do you guys think?

ErwanAliasr1: Good catch. We should not hide the build process just to catch a variable.
Two options: we can catch that variable out of the make build or we keep the actual code.

leseb: To illustrate what I just said: https://paste.fedoraproject.org/paste/hxkygIjRXZmWKhWvjbITrQ

BlaineEXE: We should parse the output so that if the format for tags changes in the future, we don't have to update this code again.

ErwanAliasr1: Agreed but not on the build.parallel

ErwanAliasr1:
To be updated since #982 got merged

tests/tox.sh
BlaineEXE: I will update this to fix.

BlaineEXE:
> The problem I've seen with this approach is that when running the output is hidden in the variable, once done, the -x takes effects and show the content of the variable make_output. What's a bit troublesome is that if the command fails, you will get the error right away and then you will get the full output. This makes debugging a bit harder as we have to scroll all the way up to find the root cause.
Without the output inside a variable, you will get the error right at the end.

I keep the `sudo make ... build.parallel` as it was now, and to get the `DAEMON_IMAGE` var, I use `make stage`, which has much less output than a parallel build. What do you think of this compromise?

BlaineEXE:
Looks like travis is failing because `libtcmu` isn't installed. Why does `src/daemon-base/__ISCSI_PACKAGES__` not have a default value?

tests/tox.sh
ErwanAliasr1: stage makes much more sense, perfect.

ErwanAliasr1:
@BlaineEXE As of today, only centos/rhel provides those packages. In a near future, I'm sure other distro will support it. So as a default, __ISCSI_PACKAGES__ is part of CEPH_BASE_PACKAGES (as it should be almost everywhere nearly soon), but stay empty for all but centos. Once more distro will support it, we'll populate the one from src and empty those who still don't have it.

BlaineEXE:
@leseb @ErwanAliasr1 I think I took care of all the issues flagged here. Thoughts?

leseb:
jenkins test ceph_ansible-luminous-xenial-filestore_osds_container

leseb:
jenkins test ceph_ansible-jewel-centos7-cluster

leseb:
jenkins test ceph_ansible-jewel-centos7-filestore_osds_container

leseb:

leseb:
jenkins test ceph_ansible-luminous-centos7-filestore_osds_container

Makefile
```diff
@@ -41,42 +41,35 @@ include maint-lib/makelib.mk
 # All flavor options that can be passed to FLAVORS
 # CEPH_VERSION,ARCH,OS_NAME,OS_VERSION,BASEOS_REG,BASEOS_REPO,BASEOS_TAG
```
ErwanAliasr1: Please update the comments accordingly.

Makefile
```diff
@@ -41,42 +41,35 @@ include maint-lib/makelib.mk
...
-	$(call set_env_var,DAEMON_BASE_IMAGE,$*) $(call set_env_var,DAEMON_IMAGE,$*) \
-	$(call set_env_var,RELEASE,$*) \
-	sh -c maint-lib/stage.py
+	@$(call set_env_vars,$*) sh -c maint-lib/stage.py
 
 daemon-base.%: stage.%
-	@$(call set_env_var,STAGING_DIR,$*) ; $(MAKE) -C $$STAGING_DIR/daemon-base $(MAKECMDGOALS) \
-	$(call set_env_var,DAEMON_BASE_IMAGE,$*)
+	@$(call set_env_vars,$*); $(MAKE) -C $$STAGING_DIR/daemon-base \
+	  $(call set_env_vars,$*) $(MAKECMDGOALS)
```
ErwanAliasr1: Why do you add extra spaces here ?

Makefile
```diff
@@ -41,42 +41,35 @@ include maint-lib/makelib.mk
...
 
 daemon.%: daemon-base.%
-	@$(call set_env_var,STAGING_DIR,$*) ; $(MAKE) -C $$STAGING_DIR/daemon $(MAKECMDGOALS) \
-	$(call set_env_var,DAEMON_IMAGE,$*)
+	@$(call set_env_vars,$*); $(MAKE) $(call set_env_vars,$*) -C $$STAGING_DIR/daemon \
+	  $(call set_env_vars,$*) $(MAKECMDGOALS)
```
ErwanAliasr1: Why do you add extra spaces here ?

Makefile
```diff
@@ -41,42 +41,35 @@ include maint-lib/makelib.mk
...
 
 do.image.%: daemon.% ;
 
-stage: $(foreach p, $(FLAVORS), stage.$(p)) ;
-build: $(foreach p, $(FLAVORS), do.image.$(p)) ;
-push:  $(foreach p, $(FLAVORS), do.image.$(p)) ;
+stage: $(foreach p, $(FLAVORS), stage.$(HOST_ARCH),$(p)) ;
```
ErwanAliasr1: So this implicit HOST_ARCH becomes ARCH ? Couldn't we have the same name across the code to avoid confusions ?

maint-lib/makelib.mk
```diff
@@ -1,28 +1,32 @@
 # Container images built for each flavor
 IMAGES_TO_BUILD := daemon-base daemon
 
+HOST_ARCH ?= $(shell uname --machine)
```
ErwanAliasr1: Why not naming it ARCH or rename ARCH to HOST_ARCH ? This double name for the same thing is confusing.

Do you envision that some override HOST_ARCH ?

tests/tox.sh
ErwanAliasr1: @BlaineEXE @leseb does this item have been handled ?

maint-lib/makelib.mk
BlaineEXE: `HOST_ARCH` here can be overridden easily in a future where we are doing crossbuilds. E.g.,

```
make HOST_ARCH=x86_64 build
make HOST_ARCH=aarch64 build
```

As for the double naming, I would rather rename our current usage of `ARCH` to `HOST_ARCH`, as just `ARCH` could be vague.

ErwanAliasr1: I'm fine with changing everything to HOST_ARCH but having two name of the same thing is disturbing.

Makefile
ErwanAliasr1: Ok didn't caught that.

Makefile
ErwanAliasr1: Ok

BlaineEXE:
Okay. Made changes and squashed a few of the commits.

PR #981, opened 2018-03-26
[WIP] Template using jinja2

BlaineEXE:
Please ignore the fact that this is all in a single commit. If this is something we want to continue with, I'll clean it up. Basically, I wanted to prototype what templating would look like if we used yaml for input values and jinja for templating instead of `__VARIABLE__` files. For yaml files, it's imperative that override yaml files don't overwrite the previous file but instead get merged into them. And we need to have one definitive yaml file, otherwise if the same key but with different values exists in 2 different files in the flat staging hierarchy, we don't know what the override priority should be.

After a simple prototype with jinja, I thought this would be straightforward, but it turns out that it's not. Our current schema uses top-down template replacements, basically, which is very simple. Jinja can do that easily too. But in order to do multi-layer templating and make it so the same list of packages can be used for both with-point-release packages and without-point-release packages, I had to implement bottom-up templating (recursively). I did this by adding a `finalize` hook onto the data that will render strings that are also templates before returning them.

Continuing on with recursive templating, we know our yaml input data can only be strings, lists, or dicts once they reach pyyaml. So it's easy to also detect if there is a list or dict and recursively render each item in the data structure. And because rendering always returns strings, we can try to deserialize string-representations of lists and dicts.

I wrote a jinja macro to render the packages cleanly with a selector for showing/not-showing the point release information, and I have to explicitly call render (via a jinja filter) on the input macros don't go through the normal rendering process.

Implementing this, I realized that one of the non-image-build source files has several `{{ xxx }}` sections in it for `kubectl`. Really, I don't think we want to do templating on the main source files, only on the files used for image builds, so instead of parsing all files, I hard-coded the `'Dockerfile'` since that's our only image build spec file. Eventually, we might have different files than docker, but that's an easy add later.

---

This all seems a little overwhelming maybe. But now that the hard part of converting to jinja is figured out, it is just a matter of deciding if the switch is worth it and updating the docs. 

Positives:
 - Jinja gives us more templating power including conditionals and loops for keeping the code DRY.
 - Single template config yaml file in each staging dir
 - yaml template config files allow for more comments to tell users how to configure fields

Neutral:
 - The template writing itself is about the same as before, just the syntax is different (other than the added conditionals/loops/macros)

Negatives:
 - The macro for expanding packages is complex and might seem overwhelming
   - Solution could be to write a custom jinja filter just for rendering packages since it's a common operation
   - a slight positive of the complexity is that it's a good example for others to write their own macros
 - Reading the template config files for parameters is a little more difficult due to yaml tabbing and generally longer files


BlaineEXE:
Follow-up:

I have only implemented the jinja templating for opensuse's daemon-base image, so you can see how the templating is working by:
```
make FLAVORS=luminous,x86_64,opensuse,42.3,_,opensuse,42.3 stage CEPH_POINT_RELEASE="=12.2.2"
cat staging/luminous-opensuse-42.3-x86_64/daemon-base/Dockerfile
```

ErwanAliasr1:
@BlaineEXE that is serious work. Congrats for putting it alive so its easier to estimate the changes & benefits. Your message is well detailed & elaborated.

The principles we used in that project were about keeping everything file based & easy to understand.
The matriochka approach we used induced a merge of data (files in the current code, yml in this poc).
In the current code, having a quick look at the directory tree, we can estimate how the files & variables will be stacked.
In this PR, understanding how the yml files will be merged requires a good mental skill. I don't have the feeling that we are winning on that part.

Aside the issue we have in #978, which is still unsolved, I don't really see a huge benefits of using jinja.

BlaineEXE:
@ErwanAliasr1 
> In this PR, understanding how the yml files will be merged requires a good mental skill. I don't have the feeling that we are winning on that part.

I think you're right. I forgot to mention that in my write-up I think, but it is my least favorite part of the jinja. I think this has been a reaffirmation of why we didn't want to use jinja initially.

ErwanAliasr1:
@BlaineEXE I'm glad we are on the same page regarding the goal & the spirit of that project.

Your proposal about renaming STAGE_REPLACE_WITH makes so much sense and I have to admit I also found that name too long and obfuscating.

So let's close this PR and create another to rename STAGE_REPLACE.

BlaineEXE:
@ErwanAliasr1 Already on it. =)

BlaineEXE:
I think @ErwanAliasr1 and I both agree that this is unnecessary. If we do need jinja later for some reason, we can always come back to this work, but I think it's a little too much for now.

Closing.

PR #982, opened 2018-03-28
staging: rename `STAGE_REPLACE_WITH_X` to `__ENV_[X]__`

BlaineEXE:
`STAGING_REPLACE_WITH_X` doesn't really make it clear the context from
which `X` will be retrieved unless you read the docs. Changing this to
`__ENV_[X]__` makes it more clear from reading the file. Additionally,
now all of our file variables are surrounded by double-underscores,
making them look more uniform.


ErwanAliasr1:
I like the principle of that PR. I'm not a big fan of the underscore between ENV and [.

Couldn't we have __ENV[x]__ which is closer to a usual bash or python code. The additional _ is kind of verbose and error prone.

I do agree that having all variables look like the same is nice. But isn't that strange for new comers than __ENV[x]__ isn't a file while all others __X__ are ?

BlaineEXE:
> I do agree that having all variables look like the same is nice. But isn't that strange for new comers than `__ENV[x]__` isn't a file while all others `__X__` are ?

I think there could be some confusion, but I think the combination of the text `ENV` and the brackets suggest that maybe something different is happening here.

Options I considered:
- `__ENV[VARIABLE]__` Clear, but ENV and VARIABLE visually run together a bit
- `__ENV['VARIABLE']__` Quotes are probably too much
- `__ENV_[VARIABLE]__` What I chose, which adds more dileniation between ENV and VARIABLE
- `ENV[VARIABLE]` In the context of a file, it's not immediately clear that this is a variable or is something within the file itself
- `__ENV__[VARIABLE]` I don't think this is a bad alternative to `__ENV[V]__` or `__ENV_[V]__`, but I thought maybe not having a double underscore at the end could be confusing. It also might get accidentally parsed as `<contents of file __ENV__>[VARIABLE]` where `[VARIABLE]` is left as written in the file.
- `__ENV__[VARIABLE]__` Probably too verbose. Possible parse issues as above.
- I tried to think of a different separator than double-underscores, but I couldn't think of anything that made more sense. 
  - I considered using double braces `{{VARIABLE}}` or `{{ENV[VARIABLE]}}` like jinja, but I don't want to use the same syntax as a well-known templating engine so as not to confuse anyone that way.
  - I considered double brackets `[[VARIABLE]]` or `[[ENV[VARIABLE]]]` as well, but there may be a corner case somewhere with someone using a python list within a list like `[[ENV[PYTHON_STUFF]]]`
  - Double parens `((ENV[VARIABLE]))` might have a similar corner case to above.

I think maybe the issue I am having where `__ENV[V]__` looks challenging to me is that `[` could be mistaken at a glance for `l` or `I` or `|` when it is scrunched together with other things.

I guess there are hybrid approach options like `__ENV[[VARIABLE]]__` More verbose than `__ENV[V]__` and still has possibly weak delineation between `ENV` and what comes after.

I guess the TL;DR is that I thought of all that and decided that I thought `__ENV_[VARIABLE]__` was the clearest representation

ErwanAliasr1:
@leseb any opinion on that topic ?

BlaineEXE:
@leseb ?

leseb:
I like @ErwanAliasr1's proposal to stick with `__ENV[VARIABLE]__` but I tend to agree that pattern `__ENV_[VARIABLE]__` brings a bit more clarity. So I'd say let's stick with @BlaineEXE's proposal.

ErwanAliasr1:
Merged. Thanks for the good work.

PR #983, opened 2018-03-29
[skip ci] iscsi: clarify scenario names

leseb:
Let's call the scenario with the name of daemon, it's clearer.


leseb:
There are no tests yet in ceph-ansible, this relies on https://github.com/ceph/ceph-ansible/pull/2466, merging as is for now.

PR #985, opened 2018-03-29
[skip ci] contrib: use centos as latest instead of ubuntu

leseb:
CentOS is the distro that currently has:

* the best CI coverage in terms of functional tests
* all the packages for all the various ceph projects

So this is natural to us to change our latest image to poin to CentOS
instead of Ubuntu.

Closes: https://github.com/ceph/ceph-container/issues/984


leseb:
Needs rebase on github.com/ceph/ceph-container/pull/979

PR #994, opened 2018-04-06
Updating confd/etcd/kubectl

ErwanAliasr1:
Confd/etcd/kubectl didn't got updated since a very long while and mostly because of the complex tree/file architecture.

Since we have the new tree & build tools, its much more easier to update versions for all flavors.

This PR offer to bump :

- etcd from 2.2.4 to 3.2.18
- confd from 0.10.0 to 0.15.0
- kubectl from 1.6.0 to 1.8.11

### DIFF
--- a/ceph-releases/ALL/daemon/__WEB_INSTALL_CONFD__
+++ b/ceph-releases/ALL/daemon/__WEB_INSTALL_CONFD__
@@ -1,5 +1,5 @@
 echo 'Web install confd' && \
-      CONFD_VERSION=0.10.0 && \
+      CONFD_VERSION=0.15.0 && \
       # Assume linux
       CONFD_ARCH=linux-__ENV_[GO_ARCH]__ && \
       wget -q -O /usr/local/bin/confd \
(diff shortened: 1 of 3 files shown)


Raboo:
is there a downside to upgrading etcdctl to v3? Currently all etcd keys are set using etcd v2 api. If etcdctl will try to read v3 API, it will not find any keys in the current etcd cluster.
etcd has different keys for v2 and v3 and they are not shared in the same etcd cluster.

Raboo:
I just checked, https://github.com/coreos/etcd/blob/master/Documentation/dev-guide/interacting_v3.md
> By default, etcdctl talks to the etcd server with the v2 API for backward compatibility.

So etcdctl will continue to work with the v2 API without modification.

timn:
@ErwanAliasr1 I have opened #996, but I currently do not have the time to prepare a PR. I also can't apply labels, otherwise I'd add the "help wanted" label there.

xcompass:
👍

ErwanAliasr1:
So anyone against that PR ?

Issue #937, opened 2018-03-09
CEPH_VERSION_NUMBER container env var


BlaineEXE:
Idea to introduce an environment variable like CEPH_VERSION_NUMBER in case we want to execute code paths conditionally for Ceph versions lt/gt rather than just by name. This will help us avoid the situation we have with jewel/kraken where we have unique, slightly modified source files in the ceph-releases tree. I see this as something we want to do for luminous+ but needn't refactor jewel/kraken.

Simple example will not execute a code path for earlier Ceph versions that don't support some ganesha feature:
```sh
if [ $CEPH_VERSION -ge 12 ]; then
    # do stuff with ganesha if ceph version is luminous or above
fi
```

ErwanAliasr1:
I'd also love having something like

if [ "$CEPH_NAME" = "luminous" ]; 

That would be more generic than the release itself.

BlaineEXE:
The `CEPH_VERSION` variable is the named ceph release. We could rename `CEPH_VERSION` to `CEPH_VERSION_NAME` for more clarity as part of this.

ErwanAliasr1:
CEPH_CODENAME & CEPH_VERSION ?

BlaineEXE:
I think we should be very specific. Already as-is, `CEPH_VERSION` is unspecific about whether it means number or name. I would use `CEPH_VERSION_NAME` and `CEPH_VERSION_NUMBER` for maximum clarity.

PR #997, opened 2018-04-06
variables_entrypoint: Exporting CEPH_* variables

ErwanAliasr1:
As per issue #937, it would be useful to let user scripts being able to
know what version of ceph is installed to have some conditional code
like :

    if [ "$CEPH_NAME" = "luminous" ];

CEPH_VERSION & CEPH_POINT_RELEASE variables are now published inside the container.


src/daemon/variables_entrypoint.sh
```diff
@@ -98,6 +98,12 @@ if [[ "$KV_TYPE" == "etcd" ]]; then
   ETCDCTL_OPTS=(--peers ${ETCD_SCHEMA}${KV_IP}:${KV_PORT})
 fi
 
+# CEPH VARIABLES
+CEPH_VERSION=__ENV_[CEPH_VERSION]__
+CEPH_POINT_RELEASE=__ENV_[CEPH_POINT_RELEASE]__
+export CEPH_VERSION
```
leseb: can't we combine this into 2 lines instead?

```
export CEPH_VERSION=__ENV_[CEPH_VERSION]__
export CEPH_POINT_RELEASE=__ENV_[CEPH_POINT_RELEASE]__
```

ErwanAliasr1: That is triggering SC issues ;)

leseb: Mouarf ok.

BlaineEXE: I don't feel great about having these template variables in non-build (i.e., Dockerfile) files.  `CEPH_VERSION` is already set in the dockerfile by this line `ENV CEPH_VERSION __ENV_[CEPH_VERSION]__`. I think we should just do the same for `CEPH_POINT_RELEASE`.

BlaineEXE: Aside:
In retrospect, allowing the templating system to do replacements in all source files was, IMO, a bad idea. I think it safer to only template the build files (we only have the Dockerfile at present, but we may have others if we support other container technologies in the future).

BlaineEXE: Update: I wanted to make sure that `ENV` variables were available at container runtime, and it seems that this is correct. TL;DR: `ENV` is available during build and during container runtime, and `ARG` is available only at build.

https://vsupalov.com/docker-arg-env-variable-guide/#arg-and-env-availability

src/daemon-base/Dockerfile
```diff
@@ -6,6 +6,7 @@ MAINTAINER __DOCKERFILE_MAINTAINER__
 __DOCKERFILE_TRACEABILITY_LABELS__
 
 ENV CEPH_VERSION __ENV_[CEPH_VERSION]__
```
BlaineEXE: As a suggestion (but I'm not hard-set), it could be useful to comment these so the user might have a better idea of what to expect. Like:

```
# Named Ceph version. e.g., luminous, mimic, ...
ENV CEPH_VERSION ...
# Specific Ceph version with a package version specifier symbol prepended (may be unset w/ empty string)
# e.g. -12.3.0<I'm not sure what the rest of the centos text is>, =12.3.0

ErwanAliasr1: That duplicates the information in many places ... we have a good documentation. So let's use it ;)

PR #999, opened 2018-04-09
[skip ci] contrib: build latest-bis image

leseb:
latest-bis is needed by ceph-ansible so it can test the restart
handlers on an image ID change.
So we rebuild latest again to get a different image ID.

We extracted the test for a tagged head so that we can separate each
function's call (push_ceph_imgs_latests and build_and_push_latest_bis).

latest-bis needs to exist if latest tags were built, so basically when
we are not building a release.

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -64,17 +64,18 @@ function build_ceph_imgs {
 function push_ceph_imgs {
   echo "Push Ceph container image(s) to the Docker Hub registry"
   make RELEASE="$RELEASE" push.parallel
+}
 
+function build_and_push_latest_bis {
+  # latest-bis is needed by ceph-ansible so it can test the restart handlers on an image ID change
+  # rebuild latest again to get a different image ID
+  make RELEASE="$BRANCH"-bis FLAVORS="${CEPH_RELEASES[-1]}",centos,7 build
+  docker tag ceph/daemon:"$BRANCH"-bis-"${CEPH_RELEASES[-1]}"-centos-7-x86_64 ceph/daemon:latest-bis
+  docker push ceph/daemon:latest-bis
 }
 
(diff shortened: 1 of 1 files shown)


contrib/build-push-ceph-container-imgs.sh
```diff
@@ -92,6 +99,7 @@ function push_ceph_imgs_latests {
       docker push ceph/$i:"$latest_name"
     done
   done
+  build_and_push_latest_bis
```
ErwanAliasr1: I don't really like having a build + push inside a push() function.
Please move it elsewhere to keep a better semantic.

If you need to be here for any particular reason, it have to be explicit.

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -64,17 +64,18 @@ function build_ceph_imgs {
 function push_ceph_imgs {
   echo "Push Ceph container image(s) to the Docker Hub registry"
   make RELEASE="$RELEASE" push.parallel
+}
 
+function build_and_push_latest_bis {
+  # latest-bis is needed by ceph-ansible so it can test the restart handlers on an image ID change
+  # rebuild latest again to get a different image ID
+  make RELEASE=master-bis FLAVORS="${CEPH_RELEASES[-1]}",centos,7 build
```
ErwanAliasr1: It would be better if you could reuse BRANCH variable here to avoid hardcoded values.

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -64,17 +64,18 @@ function build_ceph_imgs {
...
+  docker tag ceph/daemon:master-bis-"${CEPH_RELEASES[-1]}"-centos-7-x86_64 ceph/daemon:latest-bis
```
ErwanAliasr1: ditto

PR #1001, opened 2018-04-09
images: Use the tini entrypoint for both images

BlaineEXE:
All builds tested working on my machine.

----------------

Install tini in daemon-base image, and set it as part of the entrypoint
in both images. tini should be a nice safeguard to protect
ceph-container from zombie processes (resource leak), and it ensures
signal handlers work properly.

tini source: https://github.com/krallin/tini

------------------------------------------------------------------------
Q&A

Q: Docker 1.13 has tini built in, so what gives adding it here?
A: In order for Docker to use tini, the `--init` flag must be passed,
   which Kubernetes does not do automatically, nor could I find an
   option to pass it. Outside of Kubernetes, adding it in the container
   directly also gets rid of user error forgetting to add the `--init`
   flag.

Q: In the daemon image, why specify a custom entrypoint with tini
   instead of running `entrypoint.sh` from a `CMD` instruction?
A: Specifying a custom entrypoint allows the user to issue a custom
   command and still have the container initialize as normal using the
   entrypoint. Otherwise, the user might have to execute
   `/entrypoint.sh` manually before going about what they set out to do
   in the first place. The entrypoint can be disabled/overridden
   manually if the user feels the need.


BlaineEXE:
@bassam may also be interested in this.

ErwanAliasr1:
I"m not so sure about the real benefits of using tini.

In the actual code base, thanks to the docker_exec.sh, we have a couple of features that I don't see in tini:

- stayalive : it is possible to ask a container not to die and offer a shell if a process crash. That is useful for debugging/support/testing
- pre/post : we can define what to execute before and after a sigterm is received to cleanup the env properly

Calling the actual entrypoint with tini sounds like doing almost twice the same stuff as entrypoint.sh uses docker_exec.sh

BlaineEXE:
I don't think tini duplicates any functionality. 

Tini runs as a proper PID1 process, and does the necessary sigterm handling, which means that it "passes" all signals to the process without just eating them up and throwing them away. `docker_exec.sh` implements a signal "handler," but it is not in charge of passing the signals.

Zombie process reaping is another linux-expected role of the PID1 process, and Tini does this. `docker_exec.sh` does not. If we have any daemons that create zombie processes without reaping them we may experience resource leak, which I believe would lead to unexpectedly crashing containers. If this were something on a mon and considering the example where all mons to leak the same resources at about the same rate, this could lead to all mon containers crashing at about the same time, and it could be very painful to recover afterwards.

https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

ErwanAliasr1:
@BlaineEXE docker_exec pass the SIGTERM to the exec'd child but yes, not the others. What kind of signals do we expect from docker ?

I have to dig around the zombie case.

BlaineEXE:
> @BlaineEXE docker_exec pass the SIGTERM to the exec'd child but yes, not the others. What kind of signals do we expect from docker ?

> I have to dig around the zombie case.

I don't know all of the signals, but there are plenty of articles/blogs/etc. on the internet that suggest Docker's innate handling of PID1 is insufficient, and tini solves that issue nicely. I really haven't concerned myself as much with how it works as with the body of evidence that suggests that it is important and that not doing this small thing opens us up to unnecessary risks. Perhaps @bassam knows more about how/why it is necessary, but I think I have provided quite enough logic and rationality to why this should be done.

leseb:
I think we have to evaluate pros and cons of using Tini instead of what we currently have. As Erwan mentioned already if we go with Tini we lose a bunch of critical features from our reworked `exec` function (`docker_exec`):

* losing the ability to let the container alive if it crashes is crucial while debugging
* the pre/post exec functions are also nice to have, they might go away in the future but there are still required

To me, this is sufficient to not go with Tini, even though Tini seems appealing and well recognized our current implementation of signal handling hasn't shown its limit (yet?). I know going with Tini might prevent things in the future but in the end, it's all about the balance. Currently, I see more loss than gain.

One thing would be implementing our features in Tini, I'm just not sure how much time/effort it will take.
For now, this is -1 for me.

BlaineEXE:
since there are uncertainties, I can test the container keepalive to make sure it still works, but I don't understand why you think tini is mutually exclusive to docker_exec. It shouldn't be. In https://github.com/krallin/tini/issues/8, the tini author explains:

> Tini fixes by "forwarding signals": if you send a signal to Tini, then it sends that same signal to your child process (Jenkins in your case).

and

> Tini does install explicit signal handlers (to forward them, incidentally), so those signals no longer get dropped. 

He later explains:
> Tini is transparent, so if you're unsure, adding it shouldn't hurt.

All tini will do here is pass the signals to `entrypoint.sh` for it to handle as it will. If `entrypoint.sh` with `docker_exec` handles the signal it has had forwarded to it by dropping the signal, then the signal should be effectively dropped.

leseb:
@BlaineEXE oh sorry I might have misunderstood some of the Tini's goal. I guess the best thing to do is to test it out.

leseb:
I'm going to try it out too and will share my results here.

BlaineEXE:
@leseb, Thanks. Testing is definitely a good idea, but I don't have an environment ready right now. I'm fine letting this PR hang around for a while if we want to be rigorous about the testing.

BlaineEXE:
Are the 3 tests that are failing right now (xenial-bluestore, xenial..cluster_collocation, centos7-filestore) known to be not working now? As much as I can tell, the CI is still passing unless there is something I missed in any of those 3 tests.

BlaineEXE:
After messing with Rook's build, it seems to me that it isn't necessary to have tini in ceph-container, as it is added by Rook's build.

This doesn't mean that tini wouldn't be useful in a docker-only context, but perhaps that is something end users can add if they need it.

BlaineEXE:
After messing with Rook's build, it seems to me that it isn't necessary to have tini in ceph-container, as it is added by Rook's build.

This doesn't mean that tini wouldn't be useful in a docker-only context, but perhaps that is something end users can add if they need it.

leseb:
Yes, I think we can revisit this later if needed. Thanks!

PR #1007, opened 2018-04-12
[skip ci] introduce post install tweaks

leseb:
This commit introduces 'tweaks', tweak is a new capability that allows
you to configure the container OS. Typically this makes a particular
service more container aware or change its default configuration so it
can work better in containers.

Its first implementation is for configuring LVM.
lvm when creating devices tries to talk to udev. In a container this is
not possible and lvm can not reach udev on the host. Disabling the sync
and the rules help us getting through the lv creation.
Later to scan the lv on the host you can use lvs --readonly

### DIFF
--- a/src/__DOCKERFILE_POSTINSTALL_TWEAKS__
+++ b/src/__DOCKERFILE_POSTINSTALL_TWEAKS__
@@ -1,0 +1,5 @@
+# disable sync with udev since the container can not contact udev
+sed -i -e 's/udev_rules = 1/udev_rules = 0/' -e 's/udev_sync = 1/udev_sync = 0/' /etc/lvm/lvm.conf && \
+# validate the sed command worked as expected
+grep -sqo "udev_sync = 0" /etc/lvm/lvm.conf && \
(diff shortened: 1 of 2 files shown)


leseb:

src/__DOCKERFILE_POSTINSTALL_TWEAKS__
```diff
@@ -0,0 +1,2 @@
+# disable sync with udev since the container can not contact udev
+sed -i -e 's/udev_rules = 1/udev_rules = 0/' -e 's/udev_sync = 1/udev_sync = 0/' /etc/lvm/lvm.conf
```
ErwanAliasr1: I think we should be able to detect if the change got successful. Looks like it's not trivial doing it in sed, maybe greeping after ?

leseb: done.

leseb:
Travis successfully built an image so I guess there is no need to run the whole CI for this change.

PR #1014, opened 2018-04-16
Check installed packages _after_ cleanup 

ErwanAliasr1:
Installing packages & checking if they are properly installed just after
have a major issue : if the code than clean the container deletes too
much packages, this will not be caught and leads to an unusable
container.

This patch aims at moving this check inside the POSTINSTALL sections.


ErwanAliasr1:
@BlaineEXE I do agree on the dependencies but not on the project's packages. I don't see any valid reason why an upstream distro would remove any of the CEPH_BASE_PACKAGES.

BlaineEXE:
I think this patch constrains builds too much. It forces containers to always keep the installed packages installed as-is, but there may be reasons to force-remove parts of packages to reduce image size. The minute we start cleaning anything in the container, we risk breaking it. The only way to prevent cleaning from breaking things is to remove cleaning altogether. This is obviously too restrictive, and I think this patch leans too far in the direction of restrictions. At some point, we must trust that the user/developer has taken care to not delete what is needed in the container while also allowing them the freedom and flexibility to delete what they feel is necessary.

ceph-releases/ALL/ubuntu/__DOCKERFILE_POSTINSTALL_CLEANUP__
```diff
@@ -1,6 +1,5 @@
 echo 'Postinstall cleanup' && \
      ( echo "apt clean" && DEBIAN_FRONTEND=noninteractive apt-get clean && \
-      echo "apt autoremove" && DEBIAN_FRONTEND=noninteractive apt-get autoremove && \
```
BlaineEXE: Why remove the `autoremove` command. All this does is remove dependencies that are no longer needed. It shouldn't break things that were explicitly installed. But, for example, it will remove the unneeded dependencies pulled in by `wget` if `wget` is removed.

ErwanAliasr1: in fact it did removed everything, including ceph ...

BlaineEXE: Huh ... that's not good. I wouldn't think that should happen, but obviously it has to be removed.

ceph-releases/ALL/ubuntu/daemon-base/__EXTRA_POSTINSTALL_CLEANUP__
```diff
@@ -0,0 +1 @@
+dpkg -s __CEPH_BASE_PACKAGES__
```
BlaineEXE: @ErwanAliasr1 : As an aside, we forgot to change these to `apt-cache show` (I think `show` is the right command) when we made the point release update. Maybe you can make a new PR for this?

BlaineEXE: `apt-cache show` is correct


```
root@2d051cc01bd6:/# apt-cache show sed=4.2.2-7
Package: sed
Essential: yes
Status: install ok installed
Priority: required
Section: utils
Installed-Size: 304
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Architecture: amd64
Multi-Arch: foreign
Version: 4.2.2-7
Depends: dpkg (>= 1.15.4) | install-info
Pre-Depends: libc6 (>= 2.14), libselinux1 (>= 1.32)
Description: The GNU sed stream editor
 sed reads the specified files or the standard input if no
 files are specified, makes editing changes according to a
 list of commands, and writes the results to the standard
 output.
Description-md5: 67b5a614216e15a54b09cad62d5d5afc
Original-Maintainer: Clint Adams <clint@debian.org>
Homepage: http://www.gnu.org/software/sed/

root@2d051cc01bd6:/# apt-cache show sed=4.2.3-7
E: Version '4.2.3-7' for 'sed' was not found
E: No packages found
```

BlaineEXE: And when version is not found, the return code is `100`, so it does error. =)

BlaineEXE:
LGTM. Merge when you like.

ceph-releases/ALL/ubuntu/daemon-base/__EXTRA_POSTINSTALL_CLEANUP__
ErwanAliasr1: I added a very simple commit to handle that and put you as co-author.

PR #1019, opened 2018-04-17
[skip ci] ceph-releases: Adding rhel7

ErwanAliasr1:
This commit is about adding the necessary files for building a RHEL7 based image.
The main differences with the Centos image:
- no cleaning to keep packages untouched
- /var/log is cleaned up
- only use RHEL repositories (to be added by the buildsystem)
- specific LABELs
- disabling subscription-manager during the build

### DIFF
--- a/ceph-releases/ALL/rhel7/__DOCKERFILE_MAINTAINER__
+++ b/ceph-releases/ALL/rhel7/__DOCKERFILE_MAINTAINER__
@@ -1,0 +1,1 @@
+Erwan Velu <evelu@redhat.com>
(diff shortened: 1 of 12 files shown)


leseb:
@ktdreyer, please merge if this looks good to you.

ErwanAliasr1:
Items that I have issue with 

- does install.sh is really needed ?
- why do we have a k8s LABEL while kubectl isn't installed ?

ceph-releases/ALL/rhel7/__DOCKERFILE_MAINTAINER__
```diff
@@ -0,0 +1 @@
+Ken Dreyer "kdreyer@redhat.com"
```
ktdreyer: Would you please choose a different name? I am not the project maintainer.

ktdreyer: I do not

PR #1020, opened 2018-04-17
makefile: Let user override BASEOS_REGISTRY

ErwanAliasr1:
By default, the BASEOS_REGISTRY is set to "_" but some could need
overriding this variable to point to a particular registry to get
images.

This patch let users defining BASEOS_REGISTRY while keeping the same
default value.

This patch does change the behvior introduce by commit 6c90f47a44c93664f50acd67ebfe7e7786b2067d.
The registry is no more part of the BASE_IMAGE override. If one needs a
different name and a different registry, then it have to use both
BASEOS_REGISTRY & BASE_IMAGE overrides.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ RELEASE ?= $(shell git rev-parse --abbrev-ref HEAD)
 DAEMON_BASE_TAG ?= ""
 DAEMON_TAG ?= ""
 
+BASEOS_REGISTRY ?= "_"
 BASE_IMAGE ?= ""
 
 
@@ -173,6 +174,9 @@ help:
 	@echo '                      sourced from ceph-container (e.g., ubuntu:"16.04", centos:"7")'
 	@echo '    e.g., FLAVORS="luminous,ubuntu,16.04 jewel,ubuntu,14.04"'
 	@echo ''
+	@echo '  REGISTRY - The url of the registry to pull images from.'
+	@echo '             Default to "_" which means the default docker registry'
+	@echo ''
 	@echo '  TAG_REGISTRY - The name of the registry to tag images with and to push images to.'
 	@echo '             Defaults to "ceph".'
 	@echo '             If specified as empty string, no registry will be prepended to the tag.'
@@ -188,7 +192,7 @@ help:
 	@echo '  BASE_IMAGE - Do not compute the base image to be used as container base from BASEOS_REPO'
 	@echo '               and BASEOS_TAG. Instead, use the base image specified. The BASEOS_ vars will'
 	@echo '               still be used to determine the ceph-container source files to use.'
-	@echo '               e.g., BASE_IMAGE="myrepo/mycustomubuntu:mytag"'
+	@echo '               e.g., BASE_IMAGE="mycustomubuntu:mytag"'
 	@echo ''
 
 show.flavors:
--- a/maint-lib/makelib.mk
+++ b/maint-lib/makelib.mk
@@ -19,11 +19,11 @@ $(shell bash -c 'set -eu ; \
 	set_var CEPH_POINT_RELEASE "$$(bash maint-lib/ceph_version.sh "$$ceph_version_spec" CEPH_POINT_RELEASE)" ; \
 	set_var BASEOS_REPO        "$(word 3, $(subst $(comma), , $(1)))" ; \
 	set_var BASEOS_TAG         "$(word 4, $(subst $(comma), , $(1)))" ; \
-	set_var BASEOS_REGISTRY    "_" ; \
+	set_var BASEOS_REGISTRY    "$(BASEOS_REGISTRY)" ; \
 	set_var IMAGES_TO_BUILD    "$(IMAGES_TO_BUILD)" ; \
 
 	set_var STAGING_DIR       "staging/$$CEPH_VERSION$$CEPH_POINT_RELEASE-$$BASEOS_REPO-$$BASEOS_TAG-$$HOST_ARCH" ; \
-	base_img="$$BASEOS_REGISTRY/$$BASEOS_REPO:$$BASEOS_TAG" ; \
+	base_img="$$BASEOS_REPO:$$BASEOS_TAG" ; \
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -173,6 +174,9 @@ help:
 	@echo '                      sourced from ceph-container (e.g., ubuntu:"16.04", centos:"7")'
 	@echo '    e.g., FLAVORS="luminous,ubuntu,16.04 jewel,ubuntu,14.04"'
 	@echo ''
+	@echo '  REGISTRY - The url of the registry to pull images from.'
```
BlaineEXE: s/REGISTRY/BASEOS_REGISTRY/

maint-lib/makelib.mk
```diff
@@ -19,11 +19,11 @@ $(shell bash -c 'set -eu ; \
 	set_var CEPH_POINT_RELEASE "$$(bash maint-lib/ceph_version.sh "$$ceph_version_spec" CEPH_POINT_RELEASE)" ; \
 	set_var BASEOS_REPO        "$(word 3, $(subst $(comma), , $(1)))" ; \
 	set_var BASEOS_TAG         "$(word 4, $(subst $(comma), , $(1)))" ; \
-	set_var BASEOS_REGISTRY    "_" ; \
+	set_var BASEOS_REGISTRY    "$(BASEOS_REGISTRY)" ; \
 	set_var IMAGES_TO_BUILD    "$(IMAGES_TO_BUILD)" ; \
 
 	set_var STAGING_DIR       "staging/$$CEPH_VERSION$$CEPH_POINT_RELEASE-$$BASEOS_REPO-$$BASEOS_TAG-$$HOST_ARCH" ; \
-	base_img="$$BASEOS_REGISTRY/$$BASEOS_REPO:$$BASEOS_TAG" ; \
+	base_img="$$BASEOS_REPO:$$BASEOS_TAG" ; \
```
BlaineEXE: I see a couple problems with this change. I think this code will break builds. 

-----

Something that I missed is that in `src/daemon-base/Dockerfile`, we have 
```
FROM __ENV_[BASE_IMAGE]__
```
so we need the `BASE_IMAGE` variable to have the full <registry (unless '_')>/<repo>:<tag>.

I think the best thing to do is change the name of the `BASE_IMAGE` variable in the source to `BASE_IMAGE_FULLPATH` or something similar.

-----

For the second, this change doesn't ever propagate a user-specified `BASEOS_REGISTRY` forward. As a variable, it is never used here to set what is currently the `BASE_IMAGE` variable (and what I've proposed will become the `BASE_IMAGE_FULLPATH` variable). The line part `"$${base_img#_/}"` below is what strips off the leading `_/` if the registry is left as default (and leaves the registry intact if it is any other string).

maint-lib/makelib.mk
```diff
@@ -19,11 +19,11 @@ $(shell bash -c 'set -eu ; \
...
 	if [ -n "$(BASE_IMAGE)" ] ; then base_img="$(BASE_IMAGE)" ; fi ; \
```
BlaineEXE: This will still overwrite `BASE_IMAGE` with the user-specified value without adding the registry to it.

I think that the whole code flow, starting from what I see as line 26 to line 28 just below is:

```sh
base_img="$$BASEOS_REPO:$$BASEOS_TAG" ; \  # Set to the default computed value for <repo>:<tag>
if [ -n "$(BASE_IMAGE)" ] ; then base_img="$(BASE_IMAGE)" ; fi ; \  # override with user-specified <repo>:<ta> if BASE_IMAGE was specified by user
base_img="$$BASEOS_REGISTRY/$$base_img" ; \  # attach the registry to the front of <repo>:<tag>
set_var BASE_IMAGE_FULL       "$${base_img#_/}"  # Strip the registry (and the '/') if the registry is default '_', and then set the value to the env var BASE_IMAGE_FULL (which used to be BASE_IMAGE)
```

Makefile
BlaineEXE: I'm having a thought now that I'm seeing this in context: to keep this related variable consistent with `BASE_IMAGE` below, we should probably either name this `BASE_REGISTRY` OR rename `BASE_IMAGE` to `BASEOS_IMAGE`. That way there isn't a discrepancy between `BASE_` and `BASEOS_` that doesn't exist in the makefile. I think I'm more inclined to rename `BASEOS_REGISTRY` to `BASE_REGISTRY`, but I'm fine either way.

Makefile
```diff
@@ -173,6 +174,9 @@ help:
...
+	@echo '             Default to "_" which means the default docker registry'
```
BlaineEXE: I think "docker" here could be problematic. Docker (the company) has a copyright on the word "docker" in this context. I would suggest just leaving it out to remove any possibility of Docker (the company) taking legal action.

Makefile
```diff
@@ -188,7 +192,7 @@ help:
 	@echo '  BASE_IMAGE - Do not compute the base image to be used as container base from BASEOS_REPO'
 	@echo '               and BASEOS_TAG. Instead, use the base image specified. The BASEOS_ vars will'
 	@echo '               still be used to determine the ceph-container source files to use.'
-	@echo '               e.g., BASE_IMAGE="myrepo/mycustomubuntu:mytag"'
+	@echo '               e.g., BASE_IMAGE="mycustomubuntu:mytag"'
 	@echo ''
```
BlaineEXE: I think it would be useful to add an example here showing how the `BASEOS_REGISTRY` and `BASE_IMAGE` work to define a full base image path similar to what exists on line `e.g., TAG_REGISTRY="myreg" DAEMON_TAG="mydaemontag" will tag daemon "myreg/mydaemontag"`
`

maint-lib/makelib.mk
BlaineEXE: It occurs to me also that `BASE_IMAGE` should be a required env var in staging, but it isn't set as required. I'll work on a patch to fix this, and I'll plan to rebase it after we merge this.

leseb:
As @BlaineEXE mentioned, this PR was superseded by https://github.com/ceph/ceph-container/pull/1029 so I suspect this one can be closed. @ErwanAliasr1, please feel free to re-open if you have other concerns.
Thanks!

PR #1026, opened 2018-04-17
make: Revise clean targets

BlaineEXE:
Since it's now possible to change the image tags to arbitrary strings,
our clean operations might just do nothing. Revise the way the
`clean.all` target works. Inspect each docker image's metadata and look
for the string "CEPH_POINT_RELEASE" anywhere in the metadata.
"CEPH_POINT_RELEASE" is a string all of our build containers should
have, and it is unlikely that another container build will use the same
string. "CEPH_VERSION", is also probably unlikely, but it is certain to
be more likely than "CEPH_POINT_RELEASE".

Drop `clean.nuke` target, as its functionality is now encompassed by
`clean.all`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ build.all:
 
 # ==============================================================================
 # Clean targets
-.PHONY: clean clean.nones clean.all clean.nuke
+.PHONY: clean clean.nones clean.all
 
 clean.image.%: do.image.%
 	@$(call set_env_vars); rm -rf $$STAGING_DIR
@@ -96,18 +96,19 @@ clean.image.%: do.image.%
 clean: $(foreach p, $(FLAVORS), clean.image.$(HOST_ARCH),$(p))
 
 clean.nones:
-	@docker rmi -f $(shell docker images | egrep "^<none> " | awk '{print $$3}' | uniq) || true
+	@if [ -n "$(shell docker images --quiet --filter "dangling=true")" ] ; then \
+		docker rmi -f $(shell docker images --quiet --filter "dangling=true") || true ; \
+	fi
 
 clean.all: clean.nones
 	@rm -rf staging/
-	# Don't mess with other registries for some semblance of a safe clean.
-	@docker rmi -f \
-		$(shell docker images | egrep "^$(TAG_REGISTRY)/daemon(-base)? " | \
-		  awk '{print $$3}' | uniq) || true
-
-clean.nuke: clean.all
-	@docker rmi -f \
-		$(shell docker images | egrep "^.*/daemon(-base)? " | awk '{print $$3}' | uniq) || true
+	@# Inspect each image, and if we find 'CEPH_POINT_RELEASE' we can be pretty sure it's a
+	@# ceph-container image and safe to delete
+	@for image in $(shell docker images --quiet); do \
+		if docker inspect "$$image" | grep -q 'CEPH_POINT_RELEASE'; then \
+			docker rmi -f "$$image" || true ; \
+		fi ; \
+	done
 
 
 # ==============================================================================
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -88,26 +88,27 @@ build.all:
...
+.PHONY: clean clean.nones clean.all
 
 clean.image.%: do.image.%
 	@$(call set_env_vars); rm -rf $$STAGING_DIR
 
 clean: $(foreach p, $(FLAVORS), clean.image.$(HOST_ARCH),$(p))
 
 clean.nones:
-	@docker rmi -f $(shell docker images | egrep "^<none> " | awk '{print $$3}' | uniq) || true
+	@if docker images | egrep -q "^<none> "; then \
```
leseb: `egrep` is deprecated; please use `grep -E`.

Makefile
```diff
@@ -88,26 +88,27 @@ build.all:
...
-		$(shell docker images | egrep "^$(TAG_REGISTRY)/daemon(-base)? " | \
-		  awk '{print $$3}' | uniq) || true
-
-clean.nuke: clean.all
-	@docker rmi -f \
-		$(shell docker images | egrep "^.*/daemon(-base)? " | awk '{print $$3}' | uniq) || true
+	@# Inspect each image, and if we find 'CEPH_POINT_RELEASE' we can be pretty sure it's a
+	@# ceph-container image and safe to delete
+	@for image in $(shell docker images | grep ago | awk '{ print $$3 }'); do \
+		if $$(docker inspect "$$image" | grep -q 'CEPH_POINT_RELEASE'); then \
```
leseb: if you want to test whether the command returns true or false you don't need a subshell for this and simply do:

`if docker inspect "$$image" | grep -q 'CEPH_POINT_RELEASE'; then`

Unless this has something to do with make...

Makefile
```diff
@@ -88,26 +88,27 @@ build.all:
...
-	@docker rmi -f \
-		$(shell docker images | egrep "^$(TAG_REGISTRY)/daemon(-base)? " | \
-		  awk '{print $$3}' | uniq) || true
-
-clean.nuke: clean.all
-	@docker rmi -f \
-		$(shell docker images | egrep "^.*/daemon(-base)? " | awk '{print $$3}' | uniq) || true
+	@# Inspect each image, and if we find 'CEPH_POINT_RELEASE' we can be pretty sure it's a
+	@# ceph-container image and safe to delete
+	@for image in $(shell docker images | grep ago | awk '{ print $$3 }'); do \
```
leseb: Can you explain why are you grepping for "ago"? If you're looking at removing the header you can simply use `docker images -q` to only get the image's ID.

ErwanAliasr1: Grepping ago is a bad idea since its tied to a particular locale (en). If french that would fail. Getting the 4th column is surely safer.

ErwanAliasr1: Maybe docker images -q is much more direct regarding your goal

ErwanAliasr1: and -a also to get all images

BlaineEXE: > can simply use docker images -q to only get the image's ID
Waaaat!? My mind is blown. I'll fix that.

Makefile
BlaineEXE: It might have something to do with make ... or I might've just overlooked that. I'll check.

Makefile
```diff
@@ -88,26 +88,27 @@ build.all:
...
+.PHONY: clean clean.nones clean.all
 
 clean.image.%: do.image.%
 	@$(call set_env_vars); rm -rf $$STAGING_DIR
 
 clean: $(foreach p, $(FLAVORS), clean.image.$(HOST_ARCH),$(p))
 
 clean.nones:
-	@docker rmi -f $(shell docker images | egrep "^<none> " | awk '{print $$3}' | uniq) || true
+	@if docker images | grep -E -q "^<none> "; then \
```
ErwanAliasr1: Isn't that docker images --filter "dangling=true" ?

Makefile
```diff
@@ -88,26 +88,27 @@ build.all:
...
+		docker rmi -f $(shell docker images | grep -E "^<none> " | awk '{print $$3}' | uniq) || true ; \
```
ErwanAliasr1: docker images --filter "dangling=true" -q

PR #1027, opened 2018-04-18
[skip ci] ceph-releases: Adding centos-arm64

ErwanAliasr1:
This commit is about adding support for arm64 for centos.

In a perfect world, supporting a new arch is supposed to get anything
changed.

We are in a situation where the build of centos on ARM64 trigger the
following issues :
- nfs-ganesha & tcmu-runner packages are not available
- confd is not available from github

This commit is a little bit special as it try to workaround this
situation. It does create a centos-arm64 OS which links most of the
files from centos.

In this project, we are not supposed to have symlinks but as we are
using the same OS with a very few constrains specific to this arch,
Blaine & I considered that was the best approach.

The differences with centos are :

- __GANESHA_PACKAGES__ is empty to avoid it to be installed
- __WEB_INSTALL_CONFD__ is set to /bin/true to avoid the installation
- __ISCSI_PACKAGES__ is not defined to get it to empty to avoid the installation

This flavor have to be built in a particular fashion as centos-arm64 doesn't exist on the docker registry.
It have to be built with the following command :

    make FLAVORS=luminous,centos-arm64,7 BASE_IMAGE=centos:7 build

The BASE_IMAGE override the centos-arm64 to let docker use the centos:7 image.

### DIFF
--- a/contrib/build-push-ceph-container-imgs-arm64.sh
+++ b/contrib/build-push-ceph-container-imgs-arm64.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -ex
+
+#############
+# VARIABLES #
+#############
+
+CEPH_RELEASES=(luminous)
+
+
+#############
+# FUNCTIONS #
+#############
+
+function build_ceph_imgs {
+  echo "Build Ceph container image(s)"
+  make RELEASE="$RELEASE" FLAVORS="${CEPH_RELEASES[-1]},centos-arm64,7" BASE_IMAGE=centos:7 build
+  docker images
+}
+
+function push_ceph_imgs {
+  echo "Push Ceph container image(s) to the Docker Hub registry"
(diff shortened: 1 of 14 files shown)


ErwanAliasr1:
@leseb as confd, iscsi & ganesha are not installed shall we do anything to disable those features ?

contrib/build-push-ceph-container-imgs-arm64.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function build_and_push_latest_bis {
+  # latest-bis is needed by ceph-ansible so it can test the restart handlers on an image ID change
+  # rebuild latest again to get a different image ID
+  make RELEASE="$RELEASE"-bis FLAVORS="${CEPH_RELEASES[-1]},centos-arm64,7" BASE_IMAGE=centos:7 build
+  docker tag ceph/daemon:"$BRANCH"-bis-"${CEPH_RELEASES[-1]}"-centos-7-x86_64 ceph/daemon:latest-bis
+  docker push ceph/daemon:latest-bis
+}
+
```
leseb: missing one empty line

PR #1032, opened 2018-04-24
Dockerfiles: Remove dependency on awk

BlaineEXE:
awk is used to strip the trailing '/' from the output of du used to
estimate size savings in the container build. If a container release
doesn't include awk, it will fail here, and awk is totally unnecessary,
as bash string manipulation can work in its place.

### DIFF
--- a/src/daemon-base/Dockerfile
+++ b/src/daemon-base/Dockerfile
@@ -21,8 +21,8 @@ RUN \
     #   download and install packages from web, cleaning any files as you go.
     # Installs should support install of ganesha for luminous without installing for jewel/kraken.
     __DOCKERFILE_INSTALL__ && \
-    # Clean container, starting with record of current size
-    INITIAL_SIZE=$(du -sm / 2>/dev/null | awk '{print $1}') && \
+    # Clean container, starting with record of current size (strip / from end)
+    INITIAL_SIZE="$(bash -c 'sz="$(du -sm --exclude=/proc /)" ; echo "${sz%*/}"')" && \
     #
     #
     # Perform any final cleanup actions like package manager cleaning, etc.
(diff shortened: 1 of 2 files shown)


BlaineEXE:
@leseb, I can't see what code your comment was on after the update, but I forgot to also apply my update to the daemon Dockerfile, which is what I imagine you were commenting on. That's fixed now, and I did a `make build.all` that has everything building successfully here.

src/daemon-base/Dockerfile
```diff
@@ -21,8 +21,8 @@ RUN \
     #   download and install packages from web, cleaning any files as you go.
     # Installs should support install of ganesha for luminous without installing for jewel/kraken.
     __DOCKERFILE_INSTALL__ && \
-    # Clean container, starting with record of current size
-    INITIAL_SIZE=$(du -sm / 2>/dev/null | awk '{print $1}') && \
+    # Clean container, starting with record of current size (strip / from end)
+    INITIAL_SIZE="$(bash -c 'sz="$(du -sm / 2>/dev/null || true)" ; echo "${sz%*/}"')" && \
```
leseb: Instead of hiding stderr, I think we simply need to exclude `/proc` from `du`'s calculation.
I just tried in a container and I nicely get the size:

```
[root@ceph-nano-e-faa32aebf00b /]# du -sm /
du: cannot access '/proc/1558/task/1558/fd/4': No such file or directory
du: cannot access '/proc/1558/task/1558/fdinfo/4': No such file or directory
du: cannot access '/proc/1558/fd/3': No such file or directory
du: cannot access '/proc/1558/fdinfo/3': No such file or directory
1179    /
[root@ceph-nano-e-faa32aebf00b /]# du -sm --exclude=/proc /
1179    /
```

BlaineEXE:
OOPS! Didn't mean to close this.

PR #1041, opened 2018-05-02
[skip ci] common: fix detect_ceph_files function

leseb:
/etc/ceph wasn't beeing checked properly so now we check the length of
the output to see if there are any Ceph files.


leseb:
Closes: https://github.com/ceph/cn/issues/38

PR #1043, opened 2018-05-02
[skip ci] demo: do not bootstrap mgr on jewel

leseb:
We don't have official images before Jewel so it's safe to make an
exception only for this.


leseb:
Closes: https://github.com/ceph/cn/issues/38

PR #1058, opened 2018-05-14
[skip ci] demo: make sree bootstrap idempotent

leseb:
This now handles the case where the container restarts. We don't need to
go through the config again since it's already there. We just start the
API service.

### DIFF
--- a/src/daemon/demo.sh
+++ b/src/daemon/demo.sh
@@ -248,28 +248,32 @@ function bootstrap_mgr {
 # SREE #
 ########
 function bootstrap_sree {
-  if [ -z "$SREE_VERSION" ]; then
-    sree_latest=$(curl -s 'https://api.github.com/repos/leseb/Sree/releases/latest' | grep tarball_url | cut -d '"' -f 4)
-    curl -L "$sree_latest" -o sree.tar.gz
-  else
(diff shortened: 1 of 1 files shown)


src/daemon/demo.sh
```diff
@@ -248,25 +248,26 @@ function bootstrap_mgr {
...
-  cd sree
-  sed -i "s|ENDPOINT|http://${EXPOSED_IP}:${RGW_CIVETWEB_PORT}|" static/js/base.js
-  sed -i "s/ACCESS_KEY/$ACCESS_KEY/" static/js/base.js
-  sed -i "s/SECRET_KEY/$SECRET_KEY/" static/js/base.js
-  mv sree.cfg.sample sree.cfg
-  sed -i "s/RGW_CIVETWEB_PORT_VALUE/$RGW_CIVETWEB_PORT/" sree.cfg
-  sed -i "s/SREE_PORT_VALUE/$SREE_PORT/" sree.cfg
 
   # start Sree
   python app.py &
```
ErwanAliasr1: Prior this patch the app was reachable because of the cd at line 263.

I suggest the "cd sree" in the if() is managed by a pushd/popd
And then we get another one around the python app.py to insure we are in the proper dir.

PR #1067, opened 2018-05-15
Manifest 8

leseb:


ktdreyer:
trying to understand what "manifest-*" means?

leseb:
Manifest comes from the `docker manifest` command, it took me several attempts to validate this since we need to merge then the CI runs and we can check if it actually works. That's why I ended up enumerating my attempts.

PR #1072, opened 2018-05-23
DAEMON_PACKAGES: Adding rook package

ErwanAliasr1:
Rook is pretty popular, let's offer a variable to install it.
As very few distribution have it, let it undefined for all so one that
needs it can define it.

Rhel7 is defining rook but use it in daemon as the build process is a
little-bit special (contrib/compose.sh).


ErwanAliasr1:
Make sense. Thx.

PR #1073, opened 2018-05-23
maint-lib/makelib.mk: Removing double dash in STAGING_DIR

ErwanAliasr1:
Commit ed708988 introduced a double dash in the STAGING_DIR variable.

This generates staging dirs with two consecutive dashes which breaks the
previous naming.

This modification sounds like an unexpected change, this commit reverts
to the original naming.


ErwanAliasr1:
@BlaineEXE ack for merge ?

BlaineEXE:
I'm doing this on mobile. I did put the double dash in purposefully to
separate the Ceph version from the rest of the flavor text, but I don't
have a preference for double or single dash. Looks good to me if you want
to merge.

PR #1090, opened 2018-06-15
[skip ci] Stream logs

leseb:


ceph-releases/jewel/daemon/start_mon.sh
```diff
@@ -171,6 +171,11 @@ function start_mon {
     /usr/bin/ceph-mon "${DAEMON_OPTS[@]}" -i "${MON_NAME}" --mon-data "$MON_DATA_DIR" --public-addr "${MON_IP}:6789"
   else
     log "SUCCESS"
-    exec /usr/bin/ceph-mon "${DAEMON_OPTS[@]}" -i "${MON_NAME}" --mon-data "$MON_DATA_DIR" --public-addr "${MON_IP}:6789"
+    # enable cluster/audit/mon logs on the same stream
+    if [[ "$CEPH_VERSION" != "jewel" && "$CEPH_VERSION" != "kraken" ]]; then
```
ErwanAliasr1: As long as the command line is almost the same, I'd suggest here to just define a new option in the if and keep a single exec.

MON_OPTIONS=""
 if [[ "$CEPH_VERSION" != "jewel" && "$CEPH_VERSION" != "kraken" ]]; then 
MON_OPTIONS="--mon-cluster-log-to-stderr '--log-stderr-prefix=debug'"
fi
exec /usr/bin/ceph-mon "${DAEMON_OPTS[@]}" -i "${MON_NAME}" --mon-data "$MON_DATA_DIR" --public-addr "${MON_IP}:6789" $MON_OPTIONS

PR #1093, opened 2018-06-19
WIP: Initial commit of SSL support

### LINKED ISSUE #1074: Support RGW SSL
Hi, unless I missed it, it doesn't look like https://hub.docker.com/r/ceph/daemon/ supports SSL configuration for RGW. I think it should be fairly simple to add an option given a self-signed certificate if also setup by entrypoint.sh, e.g. http://www.ksingh.co.in/blog/2017/09/14/ceph-rgw-ssl/. 

Rationale: SSL (https) is necessary if testing AWS S3 Server-Side-Encryption with Customer-provided Keys (SSE-C) which Ceph supports. I'd like to be able to use ceph-container for testing an SSE-C app.

pixdrift:
Signed-off-by: pixdrift <support@pixeldrift.net>

Description of your changes:
Add environment variable RGW_SSLCERTIFICATE to pass certificate path to radosgw

Resolves #1074

Checklist:

### DIFF
--- a/src/daemon/start_rgw.sh
+++ b/src/daemon/start_rgw.sh
@@ -29,8 +29,14 @@ function start_rgw {
   log "SUCCESS"
 
   local rgw_frontends="civetweb port=$RGW_CIVETWEB_IP:$RGW_CIVETWEB_PORT"
-  if [ "$RGW_REMOTE_CGI" -eq 1 ]; then
-    rgw_frontends="fastcgi socket_port=$RGW_REMOTE_CGI_PORT socket_host=$RGW_REMOTE_CGI_HOST"
+
+  if [ ! -f "$RGW_SSL_CERTIFICATE" ]; then
+    # Suffix 's' on port required for SSL
+    rgw_frontends="${rgw_frontends}s ssl_certificate=${RGW_SSL_CERTIFICATE}"
+  else
+    if [ "$RGW_REMOTE_CGI" -eq 1 ]; then
+      rgw_frontends="fastcgi socket_port=$RGW_REMOTE_CGI_PORT socket_host=$RGW_REMOTE_CGI_HOST"
+    fi
   fi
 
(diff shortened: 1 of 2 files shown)


src/daemon/start_rgw.sh
```diff
@@ -29,8 +29,14 @@ function start_rgw {
   log "SUCCESS"
 
   local rgw_frontends="civetweb port=$RGW_CIVETWEB_IP:$RGW_CIVETWEB_PORT"
-  if [ "$RGW_REMOTE_CGI" -eq 1 ]; then
-    rgw_frontends="fastcgi socket_port=$RGW_REMOTE_CGI_PORT socket_host=$RGW_REMOTE_CGI_HOST"
+
+  if [ ! -f "$RGW_SSL_CERTIFICATE" ]; then
+    # Suffix 's' on port required for SSL
+    rgw_frontends="${rgw_frontends}s ssl_certificate=${RGW_SSL_CERTIFICATE}"
+  else
```
rawipfel: I don't know all of the RGW config options, but I guess this makes these two options mutually exclusive, is intentional?

leseb: You're right, although `RGW_REMOTE_CGI` is not encouraged and I wonder if we should not remove it.

pixdrift: It was intentional. They are mutually exclusive because rgw_frontends will be clobbered if RGW_REMOTE_CGI is true which overwrites the constructed SSL configuration.

The `else` can be removed, but that will just result in rgw_frontends being overwritten with CGI configuration if both are provided.

leseb:
@pixdrift sorry we should have gone through this one more rapidly, do you mind rebasing so we can do a new iteration? Thanks

leseb:
Closing this due to inactivity, feel free to re-open.

PR #1099, opened 2018-06-25
[skip ci] demo: import dir at running time

leseb:
If DATA_TO_SYNC is passed with a directory (also exposed) and DATA_TO_SYNC_BUCKET is
named then the content of DATA_TO_SYNC will be uploaded into the bucket
DATA_TO_SYNC_BUCKET.


leseb:
Closes: https://github.com/ceph/ceph-container/issues/1098

PR #1117, opened 2018-07-18
[skip ci] ceph-volume: show error before activate

leseb:
If the ceph-volume lvm list command fails we want to show the stack
trace from the log file.

### DIFF
--- a/src/daemon/osd_scenarios/osd_volume_activate.sh
+++ b/src/daemon/osd_scenarios/osd_volume_activate.sh
@@ -6,7 +6,10 @@ function osd_volume_activate {
 
   # Verify the device is a valid ceph-volume device
   # If not the following command will return 1 with "No valid Ceph devices found"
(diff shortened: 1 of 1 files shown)


src/daemon/osd_scenarios/osd_volume_activate.sh
```diff
@@ -6,7 +6,10 @@ function osd_volume_activate {
 
   # Verify the device is a valid ceph-volume device
   # If not the following command will return 1 with "No valid Ceph devices found"
-  ceph-volume lvm list "${OSD_DEVICE}"
+  if ! ceph-volume lvm list "${OSD_DEVICE}"; then
+    cat /var/log/ceph
```
alfredodeza: `/var/log/ceph` is a directory?

leseb: No this is where `ceph-volume` logs end up.

PR #1148, opened 2018-08-07
rgw: read config startup from ceph.conf

leseb:
It's easier to read rgw config options from the ceph.conf instead of
overriding it on the CLI.
This is a disruptive change for users that were relying on options like
RGW_CIVETWEB_IP and RGW_CIVETWEB_PORT. These options will have to
addressed in the ceph.conf so we don't break anything.

The options mentioned above are still active for the nano/demo scenario
which is really unique and not intended for production deployment.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1582411


src/daemon/demo.sh
```diff
@@ -26,6 +26,35 @@ MGR_IP=$MON_IP
...
+else
+  log "ERROR: unsupported rgw backend type $RGW_FRONTEND_TYPE"
+  exit 1
+fi
+
+: "${RGW_FRONTEND:="$RGW_FRONTEND_TYPE $RGW_FRONTED_OPTIONS"}"
+
+if [[ "$RGW_FRONTEND_TYPE" == "beast" ]]; then
+  if [[ "$CEPH_VERSION" == "jewel" || "$CEPH_VERSION" == "kraken" || "$CEPH_VERSION" == "luminous" ]]; then
+    RGW_FRONTEND_TYPE=beast
```
ktdreyer: We test if `[[ "$RGW_FRONTEND_TYPE" == "beast" ]]`, and then set `RGW_FRONTEND_TYPE=beast` again here?

leseb: Good catch, this line is not necessary since we fail eventually.

PR #1149, opened 2018-08-07
[DNM] [skip ci] Contrib build ceph versioned images for Rook

BlaineEXE:
During the container meeting this morning, we talked about Rook starting to use containers with different major releases of Ceph supported. Rook would like users to be able to decide to run Luminous or Mimic, for example, and upgrade from one to another when they wish.

Here I propose to run the script in `contrib/` I have created as a cron to ensure parity between the versions of Ceph available as RPMs and new images published to the Docker Hub. I have made some assumptions that merit discussion.

Decisions made based on assumptions:
 - ceph-container already builds the images it wants; since this is purely to build images that Rook wants, I build only the daemon-base image (with Ceph packages inside) that Rook uses
 - With the above, in order to distinguish these base images from the rest of ceph-container images (these images don't have a regular daemon component), I assume to create an image just called "ceph"
 - This new "ceph" image is a manifest link from `ceph/ceph:<version>` to `ceph/ceph-amd64:<version>` and `ceph/ceph-arm64:<version>` unless either of those architectures does not exist.

Things I have not done:
 - I have not (to my knowledge) put any changes in place to actually run this script from whatever publishing process ceph-container uses in its backend infrastructure
 - I have not uploaded or changed any images in the ceph library on Docker Hub

Description of your changes:

**Firstly, in preparation**
Make the `IMAGES_TO_BUILD` parameter configurable in the make
configuration. Add documentation to help text. This can speed up
builds/rebuilds during development if only one image is needed. It is
also intended to be used by the process that builds 'daemon-base' images
tagged with the particular ceph version which Rook will use as its
default Ceph image.

**And finally:**
The Rook project would like to have a standard way of stabilizing on a
specific Ceph package release. To accomplish this, add a script to
contrib that will build and push daemon-base images which are tagged
with the version of Ceph installed in them. Images are indended for Rook
and therefore CentOS-based.

The script enforces parity between Ceph versions available on the Ceph
package repository and the 'ceph' library on the Docker Hub. It looks
for all the Ceph package versions for a particular release or releases
(luminous and mimic to start), and it builds a Ceph image for each
package version which does not yet exist.

The first run of this script will likely be rather long, but subsequent
runs should build only 2 images at a time. The script is intended to be
able to be run as a regular cron job (daily? weekly?) that will pick up
the latest Ceph package version and publish it to the Ceph library on
Docker Hub without manual intervention.

The only manual intervention
needed should be when a new Ceph version is built and ready in
ceph-container; at that point, the `FLAVORS_TO_BUILD` variable in this
script should be updated.

Checklist:

### DIFF
--- a/contrib/build-push-cephonly-base-images.sh
+++ b/contrib/build-push-cephonly-base-images.sh
@@ -1,0 +1,309 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+#---------------------------------------------------------------------------------------------------
+# ABOUT
+#---------------------------------------------------------------------------------------------------
+
+# Build and push daemon-base images which are tagged with the version of Ceph installed in them.
+# Images are intended for Rook and are therefore CentOS-based.
+#
+# This script will look to see which Ceph package versions are available for given Ceph named
+# versions (e.g., luminous, mimic). If any of the available Ceph package versions don't have a
+# corresponding image and tag on the registry, the script will build the image and push it to the
+# registry. The script will do nothing if images are at parity with packages.
+#
+# This script can therefore be run on a cron regularly, and very little manual maintenance should be
+# necessary unless Ceph RPM releases or images published change. The only maintenance needed is to
+# add new Ceph releases to the `FLAVORS_TO_BUILD` variable below.
+
+# This script requires the hypriot/qemu-register container running as per the instructions for Rook
+# given in https://github.com/rook/rook/blob/master/INSTALL.md#building-for-other-platforms
+
+# Versioning:
+# Versions consist of 2 parts: the Ceph version and the build number following the form
+#     v<ceph version>-<build number> (e.g., v12.2.2-4)
+#   As a matter of semantics, this script calls the 'v<ceph version>' part the 'version_tag'
+
+
+#---------------------------------------------------------------------------------------------------
+# YOU CAN TUNE A SCRIPT, BUT YOU CAN'T TUNE A FISH
+#---------------------------------------------------------------------------------------------------
+
+# This is mostly an internal representation of `FLAVORS_TO_BUILD`. This script builds x86 and arm
+# images for centos, so the middle item doesn't really matter except to make this easier to grok at
+# first glance.
+#   Usage examples:
+# luminous,centos,7 = build luminous containers on centos 7
+# mimic,centos,7 = build mimic containers on centos 7
+# nautilus,centos,8 = build nautilus containers on centos 8
+FLAVORS_TO_BUILD="luminous,centos,7 mimic,centos,7"
+
+# Use the ceph library by default
+PUSH_LIBRARY="ceph"
+
+# Instead of 'daemon-base', call these images 'ceph' so users don't get these confused with the
+# daemon-base and daemon images. This is the registry where manifest images are pushed.
+PUSH_REPOSITORY="ceph"
+
+# Manifest images point to the below arch-specific images
+AMD64_PUSH_REPOSITORY="${PUSH_REPOSITORY}-amd64"
+ARM64_PUSH_REPOSITORY="${PUSH_REPOSITORY}-arm64"
+
+# Push to the docker hub by default
+#PUSH_REGISTRY_URL="https://registry.hub.docker.com/"
+PUSH_REGISTRY_URL=http://localhost:5000/
+
+# Get this many page entries at a time; this should be high enough that all tags for an image are
+# listed without pagination.
+ABSURD_PAGE_SIZE=100000
+
+
+
+#---------------------------------------------------------------------------------------------------
+# FUNCTIONS
+#---------------------------------------------------------------------------------------------------
+
+function info () {
+  echo "CEPHBUILD - INFO: ${*}"
+}
+
+# Return a list of the ceph version strings available on the server as:
+#   "12.0.0 12.0.1 12.0.2 12.1.0 ..."
+function ceph_versions_on_server () {
+  ceph_codename="${1}" ; distro_release="${2}" ; ceph_arch="${3}"
+  download_url="http://download.ceph.com/rpm-${ceph_codename}/el${distro_release}/${ceph_arch}/"
+  # Curl gives a listing of the URL like below:
+  # <a href="ceph-12.2.7-0.el7.x86_64.rpm">ceph-12.2.7-0.el7.x86_64.rpm</a>  17-Jul-2018 14:11  3024
+  # The ceph base package can be id'ed uniquely by the text ">ceph-" followed by a version string
+  pkg_regex=">ceph-[0-9.]+"
+  # pkg_list is returned in the form ">ceph-12.0.0 >ceph-12.0.1 >ceph-12.0.2 ..."
+  pkg_list="$(curl --silent "${download_url}" | grep --extended-regexp --only-matching "${pkg_regex}")"
+  version_list=""  # version strings only
+  for pkg in $pkg_list; do
+    version_list="${version_list} ${pkg#>ceph-}"  # strip '>ceph-' text from beginning of pkg names
+  done
+  info "${version_list}"
+}
+
+# Return 0 if the tag exists on the repository; nonzero otherwise
+function tag_exists_on_repository () {
+  tag="${1}"
+  repository="${2}"
+  tag_query_url="${PUSH_REGISTRY_URL}/v2/repositories/${PUSH_LIBRARY}/${repository}/tags/${tag}"
+  curl --silent --fail --list-only --show-error --location "${tag_query_url}" &> /dev/null
+}
+
+# Get all tags matching a given string on the repository
+function get_tags_matching () {
+  tag_matcher="${1}"
+  repository="${2}"
+  tag_list_url="${PUSH_REGISTRY_URL}/v2/repositories/${PUSH_LIBRARY}/${repository}/tags/?page_size=${ABSURD_PAGE_SIZE}"
+  curl --silent --fail --list-only --show-error --location "${tag_list_url}" &> /dev/null | \
+    jq -r ".results[] | select(.name | match(\"${tag_matcher}\")) | .name"
+    # jq: From the results of the curl, select all images with a name matching the matcher, and then
+    # output the name. Exits success w/ empty string if no matches found.
+}
+
+# Given a tag and a version, extract the build number for that version from the tag
+# Fail if the version doesn't match up with the tag
+function extract_build_number () {
+  tag="${1}"
+  version_tag="${2}"  # version allows us to verify that the build number version and tag match
+  local build_number
+  build_number="${tag#${version_tag}-}"  # remove '<version>-' from beginning of string
+  _=$(( build_number + 0 ))  # If build is not formatted properly, this will fail gloriously
+  echo "${build_number}"
+}
+
+# Get the tag matching a version with the highest build number
+# Return empty string if no matching version was found
+function get_tag_w_highest_build_number () {
+  version_tag="${1}"
+  repository="${2}"
+  tags_matching_version="$(get_tags_matching "${version_tag}" "${repository}")"
+  # Sort is safe here because we only get tags matching our specific version
+  echo "${tags_matching_version}" | sort | tail -1
+}
+
+# Get the next build number for a given Ceph version
+# Build number is defined as a dash followed by a build number after a version string
+#   e.g. - Ceph version 12.2.2 will be v12.2.2-0 for the first build, v12.2.2-1 for the next, etc.
+function get_next_build_number_for_tag () {
+  tag_w_highest_build_number="${1}"
+  version_tag="${2}"
+  # If there are no tags matching the version, report the next build number is zero
+  if [ -z "${tag_w_highest_build_number}" ]; then
+    echo '0'
+    return
+  fi
+  # Remove '<version tag>-' from the beginning of the string to get the build number
+  build_number="$(extract_build_number "${tag_w_highest_build_number}" "${version_tag}")"
+  echo "$((build_number + 1))"
+}
+
+MANIFEST_TOOL_VERSION="v0.7.0"
+MANIFEST_TOOL_LOCATION="/tmp/manifest-tool-${MANIFEST_TOOL_VERSION}"
+# Manifest tool: https://github.com/estesp/manifest-tool
+function download_manifest_tool () {
+  BUILD_SERVER_GOARCH="amd64"  # We assume below the build server is x86_64 (amd64)
+  if [[ ! -x "${TOOL_BINARY_LOCATION}" ]]; then
+    info "Manifest tool is not downloaded. Downloading it now."
+    curl --silent --location \
+      "https://github.com/estesp/manifest-tool/releases/download/${MANIFEST_TOOL_VERSION}/manifest-tool-linux-${BUILD_SERVER_GOARCH}" > "${MANIFEST_TOOL_LOCATION}"
+  fi
+}
+
+# For a given version of ceph, if there exists both an x86_64 image and an aarch64 image, push a
+# manifest image for multi arch support: https://blog.docker.com/2017/11/multi-arch-all-the-things/
+function push_manifest_image () {
+  ceph_version_tag="${1}"
+  amd64_tag="$(get_tag_w_highest_build_number "${ceph_version_tag}" "${AMD64_PUSH_REPOSITORY}")"
+  arm64_tag="$(get_tag_w_highest_build_number "${ceph_version_tag}" "${ARM64_PUSH_REPOSITORY}")"
+  if [ -z "${amd64_tag}" ]; then
+    info "Not pushing manifest image for tag ${ceph_version_tag} because there is no \
+x86_64 (amd64) image (${AMD64_PUSH_REPOSITORY}) tagged for this version."
+    return 0
+  elif [ -z "${arm64_tag}" ]; then
+    info "Not pushing manifest image for tag ${ceph_version_tag} because there is no \
+aarch64 (arm64) image (${ARM64_PUSH_REPOSITORY}) tagged with this version."
+    return 0
+  fi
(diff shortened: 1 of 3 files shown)


BlaineEXE:
I can't add @travisn as a reviewer, so I will just tag him here. You and @liewegas should look this over to make sure I have met Rook's needs and also to decide whether my first pass decision to create a new `ceph/ceph:<arch>-<version>` image is one you agree with.

I imagine @leseb and @ErwanAliasr1 will be along as they have time as well.

travisn:
@BlaineEXE Thanks for capturing the discussion. I believe the tag we want is as simple as `ceph/ceph:<version>`. Dockerhub has the architecture built-in to the repo name. For rook, we have the repos `rook/ceph`, `rook/ceph-amd64`, and `rook/ceph-arm64`. The arch-specific repos are private, while the arch-neutral repo is public. When you pull, docker will manage retrieving the architecture that is necessary for your local system.

BlaineEXE:
RE: @travisn 
> @ BlaineEXE Thanks for capturing the discussion. I believe the tag we want is as simple as ceph/ceph:<version>. Dockerhub has the architecture built-in to the repo name. For rook, we have the repos rook/ceph, rook/ceph-amd64, and rook/ceph-arm64. The arch-specific repos are private, while the arch-neutral repo is public. When you pull, docker will manage retrieving the architecture that is necessary for your local system.

Hmm. Is something like this set up for the ceph library? As in, do we (and by we, I mean someone with admin rights) need to make a `ceph-aarch64` and `ceph-x86_64` repo and make them private? And how do we associate `ceph/ceph` with `ceph/ceph-aarch64` and `ceph/ceph-x86_64` as arch variants? Or does docker just magically know?

travisn:
[This article](https://blog.docker.com/2017/11/multi-arch-all-the-things/) discusses how to create a manifest for working with multiple architectures. This happens in Rook in the [release makefile](https://github.com/rook/rook/blob/master/build/release/Makefile#L165). In that makefile you can also see the docker manifest tool that is called.

leseb:
@BlaineEXE @travisn I haven't looked at the PR yet, I just saw the comment about manifest and multi-arch, this is where/how we do it currently for the upstream images https://github.com/ceph/ceph-container/blob/master/contrib/build-push-ceph-container-imgs.sh#L187-L198

contrib/build-push-cephonly-base-images.sh
```diff
@@ -0,0 +1,157 @@
...
+#   "12.0.0 12.0.1 12.0.2 12.1.0 ..."
+function ceph_versions_on_server() {
+  ceph_codename="${1}" ; distro_release="${2}" ; ceph_arch="${3}"
+  download_url="http://download.ceph.com/rpm-${ceph_codename}/el${distro_release}/${ceph_arch}/"
+  # Curl gives a listing of the URL like below:
+  # <a href="ceph-12.2.7-0.el7.x86_64.rpm">ceph-12.2.7-0.el7.x86_64.rpm</a>  17-Jul-2018 14:11  3024
+  # The ceph base package can be id'ed uniquely by the text ">ceph-" followed by a version string
+  pkg_regex=">ceph-[0-9.]+"
+  # pkg_list is returned in the form ">ceph-12.0.0 >ceph-12.0.1 >ceph-12.0.2 ..."
+  pkg_list="$(curl --silent "${download_url}" | grep --extended-regexp --only-matching "${pkg_regex}")"
```
leseb: This feels similar to this wip code: https://github.com/ceph/ceph-container/pull/1108/files
Perhaps it can help

contrib/build-push-cephonly-base-images.sh
```diff
@@ -0,0 +1,157 @@
...
+    version_list="${version_list} ${pkg#>ceph-}"  # strip '>ceph-' text from beginning of pkg names
+  done
+  echo "${version_list}"
+}
+
+# Return 0 if the tag exists at the registry URL; nonzero otherwise
+function tag_exists_on_registry () {
+  tag="${1}"
+  curl --silent --fail --list-only --show-error --location \
+    "${PUSH_REGISTRY_URL}/tags/${tag}" &> /dev/null
```
leseb: You need to iterate through the pages of the docker hub, this will only report the last 10 tags, I don't think this is enough.

BlaineEXE: This checks for the existence of a single tag with curl. No iteration necessary.

leseb: Right, read too fast :).

BlaineEXE:
> I read your PR and I'm not sure I understood what is missing from the current status.
From https://hub.docker.com/r/ceph/daemon-base/tags/, what is preventing rook using them ?
Is it only the exact ceph release ?
Note we have the "latest" images which are always up to date.

Rook wants to be able to pull an image with Ceph binaries by the version of Ceph inside and not by the version of ceph-container. ceph-container-specific bits are only in the `daemon` image, so it doesn't really make sense to get base images by ceph-container versions. And Rook using a ceph latest image doesn't allow for stabilizing release branches of Rook to a specific ceph version.

contrib/build-push-cephonly-base-images.sh
BlaineEXE: @leseb @travisn @liewegas  I've been thinking more about this comment and my implementation in general. I think there is a potential issue with both my and with the #1108 implementation which is that there are not always versions of ceph for both x86_64 and aarch64. By looking, it appears that sometimes arm builds aren't done every time an x86_64 build is done. If we want to have a tag like `ceph/ceph:version` that will automatically get x86_64 or arm versions, we need to have both versions available.

For luminous, x86_64 versions  have 12.0.{0,1,2} which arm doesn't have builds for. aarch64 images start with 12.0.3. I'm assuming that this is a situation that could occur in the future. I think we can (and should) build images for both architectures, but we should not build a base package for it. Like we can have images `ceph/ceph-x86_64:12.0.2` without having a corresponding arm image, but we should not build a ceph image until we have both versions; so in my example below, the first ceph version would be `ceph/ceph:12.0.3` when we finally have an arm build. I think by doing this, we can still build the latest, but the Rook project will not need to do extra work to make sure both x86_64 and arm images exist since the existence of a `ceph` image implies that both exist.
```
ceph-12.0.0-0.el7.x86_64.rpm                       08-Feb-2017 16:40                3060
ceph-12.0.1-0.el7.x86_64.rpm                       27-Mar-2017 18:05                3060
ceph-12.0.2-0.el7.x86_64.rpm                       24-Apr-2017 14:06                3060
ceph-12.0.3-0.el7.x86_64.rpm                       17-May-2017 21:11                3060
ceph-12.1.0-0.el7.x86_64.rpm                       23-Jun-2017 13:42                3024
...
```
```
ceph-12.0.3-0.el7.aarch64.rpm                      23-Jun-2017 13:42                3032
ceph-12.1.0-0.el7.aarch64.rpm                      23-Jun-2017 13:46                2996
...
```

BlaineEXE:
@travisn, I updated my script to check if a manifest image can be built for an amd/arm image pair. Let me know what you think.

ErwanAliasr1:
@BlaineEXE I'm fine adding tags if that could help rook but it's a nack redevelopping a release process just for one project. We should be able having a single script that match all.

@leseb opened more than a month ago a PR https://github.com/ceph/ceph-container/pull/1108 to add an extra versionning. That seems to be on the idea so if we need this is the place where it should be developped.

Having a new script to handle this specific case would kind of duplicate the images increasing the support effort regarding the users (rook vs others).

contrib/build-push-cephonly-base-images.sh
```diff
@@ -0,0 +1,209 @@
...
+  tag="${1}"
+  registry="${2}"
+  tag_query_url="${PUSH_REGISTRY_URL}/v2/repositories/${PUSH_LIBRARY}/${registry}/tags/${tag}"
+  curl --silent --fail --list-only --show-error --location "${tag_query_url}" &> /dev/null
+}
+
+MANIFEST_TOOL_VERSION="v0.7.0"
+MANIFEST_TOOL_LOCATION="/tmp/manifest-tool-${MANIFEST_TOOL_VERSION}"
+# Manifest tool: https://github.com/estesp/manifest-tool
+function download_manifest_tool () {
```
leseb: Why don't you use the `docker manifest` command as suggested earlier? Is there anything you need that is not available?

BlaineEXE:
RE: @ErwanAliasr1 

> @BlaineEXE I'm fine adding tags if that could help rook but it's a nack redevelopping a release process just for one project. We should be able having a single script that match all.

> @leseb opened more than a month ago a PR #1108 to add an extra versionning. That seems to be on the idea so if we need this is the place where it should be developped.

> Having a new script to handle this specific case would kind of duplicate the images increasing the support effort regarding the users (rook vs others).

Having a script that tags releases with Ceph major and minor versions is good but doesn't allow for updating with ceph bugfix version numbers. 

This doesn't have to live in ceph-container. Another option is to have a specific project for building a container that has only Ceph binary packages in it. The problem I see is that versioning to ceph-container only matters if the ceph-container `daemon` image is being used. If the `daemon-base` is used, the ceph-container iteration doesn't matter. Rook is the only consumer at the moment, but there may be consumers in the future who wish to have officially released Ceph containers tied to the Ceph version without anything unnecessary. In my opinion, it will be vastly easier for users if they can specifiy in Rook configs that they want `ceph/ceph:12.2.2`, as this is a very clear container naming-versioning scheme. Introducing extra stuff into the tag only complicates the user experience when the ceph-container functionality is not desired.

ErwanAliasr1:
@BlaineEXE I'm maybe missing something. What prevent us in the actual release script to generate the "ceph/ceph:12.2.2" when we create/push the daemon-base to dockerhub ?

BlaineEXE:
Tagged Do Not Merge after starting a discussion about revisiting this another way.

BlaineEXE:
> @BlaineEXE  I'm maybe missing something. What prevent us in the actual release  script to generate the "ceph/ceph:12.2.2" when we create/push the  daemon-base to dockerhub ?

The thing that prevents us from doing this is that the scripts install whatever the latest version of Ceph on the download servers is without paying attention to the version. In order to know explicitly what version of ceph is installed within, the version must be tightly controlled by building with a flavor like `luminous-12.2.2,centos,7` to avoid a race condition where a script determines the latest version to be `X` and during the build, packages for `X+1` are built and uploaded to the server, and then the build will believe `X` packages are installed in the image when in reality it is `X+1`.

The script I propose here tightly controls the version to account for this race condition. Additionally, it does not add or modify code for the existing releases and will therefore be unlikely to break those accidentally. The independent script can be modified in the future while also avoiding fear of breaking the existing builds, and it can be enabled/disabled wholly without any modification of code other than to stop whatever mechanism calls this script.

As a note, the way I have written the script, it would be best to be either (1) triggered on uploads of new Ceph builds, or (2) run as a daily cron so that the versioned container images are released within 24 hours of new Ceph packages being published.

BlaineEXE:
From another conversation with @leseb:
> If we start building image less regularly we won't get any security updates from the distros.

This is a good point. I can look into modifying this script to rebuild images if the base image has been updated.

BlaineEXE:
> So if we take your approach that would mean 
> - having "latest" images for each major so the daemon image can use the latest in a easy fashion
It isn't necessary to have 'latest' images, but I could add this to this script.
> - that will have an impact on the CI for building the daemon image (who's handling that workload ?)
This doesn't change the CI at all, because these images aren't used by the daemon image.
> - why not renaming the daemon-base to ceph so that would be more obvious ?
This can be done, but we still want to release images tagged with only the ceph version, which is what this PR proposes.
> - we could remove the master-commit from the actual versionning as it doesn't have any ceph-container code in it ? Adding versioning in labels ?
That is what I have done in this PR.
> - what will be the preferred option to get up to date images ? cron ? checking if there is an update on the OS image ?
Cron was what I had in mind as I said above.

travisn:
A couple of thoughts related to versioning:
- If the OS image is updated and we rebuild the ceph image, would the currently proposal end up with the same image tag? Seems like we would need a new tag so we know something changed. 
- The `latest` tag should not be used by customers. Really it should only be used for testing. In Rook we've found that clusters end up with inconsistent versions if the `latest` or `master` tags are used. For example, if more OSDs are added or a mon fails over to a new node, it may pick up a new build instead of using the same one as other daemons.

BlaineEXE:
> A couple of thoughts related to versioning:
> - If the OS image is updated and we rebuild the ceph image, would the currently proposal end up with the same image tag? Seems like we would need a new tag so we know something changed.

That's also a good point. ... In our build service, we would just increment a build number. Something similar could work for cases where the base os is updated. `ceph/ceph:12.2.2-3` might be the third build of the `12.2.2` tag, for example.

> - The `latest` tag should not be used by customers. Really it should only be used for testing. In Rook we've found that clusters end up with inconsistent versions if the `latest` or `master` tags are used. For example, if more OSDs are added or a mon fails over to a new node, it may pick up a new build instead of using the same one as other daemons.

I forgot about using 'latest' in development. That is the only good use-case for 'latest' in my opinion for the reasons Travis stated above.

leseb:
Ok to summarize:

* `latest` is still needed for upstream CI so they can always test the latest Ceph code on each PR. This is a must-have.
* ceph will set up a CI daily cron to see if there is anything new to build. This can be:
  * a new ceph version
  * an OS update, this would mean updating all the images
* we can start building image like `ceph/ceph:12.2.x-y` where `x` is the point-release number, and `y` is the build number. This build number will be incremented each time there is something new from the base OS.
* `ceph/daemon` should start consumming these images as well, they should use `latest`.

@BlaineEXE I can see you are deeply committed to this matter, but @ErwanAliasr1 and I are trying to have a conversion and you keep on stating your Rook requirements. To be clear, the ceph-container project does not work for Rook. Its primary goal is to produce Ceph images that can be re-used by any projects. So far, Rook's requirements are excellent and we are in-lined with them, but do not assume this will always be the case. We never said "no" without any conversation, that's not going to change. And please stop bringing your fork argument each time something is not what you asked for, this is not helping.

So no hard feelings here, let's try to improve our communication.

To conclude, I'm fine having you @BlaineEXE pursuing your work on that script with the summary above.
Please add a thumb up to this comment if you agree, if not I'm happy to continue the discussion :).
Thanks @BlaineEXE @travisn @ErwanAliasr1.

BlaineEXE:
> Ok to summarize:

> * `latest` is still needed for upstream CI so they can always test the latest Ceph code on each PR. This is a must-have.
> * ceph will set up a CI daily cron to see if there is anything new to build. This can be:
>   * a new ceph version
>   * an OS update, this would mean updating all the images
> * we can start building image like `ceph/ceph:12.2.x-y` where `x` is the point-release number, and `y` is the build number. This build number will be incremented each time there is something new from the base OS.
> * `ceph/daemon` should start consumming these images as well, they should use `latest`.

I don't have this final requirement for the Rook work. The first step is to produce the images. The next (and separate step) will be for `ceph/daemon` to use them if that is the wish of the ceph-container project. To make this change in ceph-container will be a fair bit more work.

Some more specific requirements:
* Build both x86_64 and aarch64 packages into images, and make those available as `ceph/ceph-amd64` and `ceph/ceph-arm64`.
* Create a manifest image which points to arm and amd images if (and only if) both architecture images exist - this will prevent users from trying to use a `ceph/ceph` version that is compatible with one arch but not the other and becoming confused.
  * I can imagine that there may be other reasons to change the 'if and only if' part of this requirement if anyone has other considerations

> @BlaineEXE I can see you are deeply committed to this matter, but @ErwanAliasr1 and I are trying to have a conversion and you keep on stating your Rook requirements. To be clear, the ceph-container project does not work for Rook. Its primary goal is to produce Ceph images that can be re-used by any projects. So far, Rook's requirements are excellent and we are in-lined with them, but do not assume this will always be the case. We never said "no" without any conversation, that's not going to change. And please stop bringing your fork argument each time something is not what you asked for, this is not helping.

I still feel as though my motivations or position are misunderstood. I'd ask for a minute to consider things from my perspective: I want to clarify first that the word "fork" is not what I intend here. Semantics are important. I have distilled Rook's requirements into a fairly simple need but one with a complicated interaction with what currently is. The ceph-container project has its own requirements and desires, and if those two projects' needs/wants are no longer aligned, I don't think it is an inappropriate argument to question whether a new project (not fork) is a good solution for Rook's needs if it also benefits Ceph. And I have brought this up precisely because I don't want the ceph-container project to be working for Rook. I don't want the ceph-container project to be at a disadvantage because of the requirement I have. I have put a lot of work into ceph-container myself, and I want to see it be a success, even if that means the work I do there doesn't end up in Rook. Which would be sad, but sometimes that's how things need to evolve.

> So no hard feelings here, let's try to improve our communication.

I agree. All around.

> To conclude, I'm fine having you @BlaineEXE pursuing your work on that script with the summary above.

BlaineEXE:
All that said, I have pushed a new version of this script which adds support for build numbers.

I will still need to add code to support tagging `v12` = `v12.2` = `v12.2.7-0`, etc.

BlaineEXE:
I've spent a little more time thinking about this and looking at other images and at the ceph packages again.
1. I don't see any instances where Ceph RPMs have a different build number than 0, but if that happens, we may not know to build an update when ceph v12.2.2-1, for example, is released. I think we should consider the Ceph package's build number when deciding whether to build a version. 
2. Ubuntu solves the build number problem by having images like `bionic-20180821`. Using the build date as a build number would be deterministic, less prone to errors, *and* monotonically increasing.

After noting 1 and 2 above, I was going to to propose tagging Ceph images `v<ceph-version-with-build-number>-<build-date>` like `v12.2.2-0-20180827`. But when I try to `yum install ceph-12.2.2-0` on centos, it fails, but `yum install ceph-12.2.2` still works. Noting this, I would propose to build a new version tagged `v<ceph-version>-<build-date>` if the Ceph package's build number is updated. This would have to be tracked by adding a ceph version label to the image like `ceph-package-version: 12.2.2-1` so we could see if the package has been updated on the rpm server. Unless, of course, it is a matter of release policy that the RPMs don't get build version updates.

BlaineEXE:
Closed in favor of #1175

PR #1163, opened 2018-08-20
[DNM] daemon-base: Minimize image to only Ceph packages

BlaineEXE:
Establish a contract for the daemon-base image to be as minimal as
possible. It should have all of the Ceph binaries installed in it and
nothing extra.

Move Ganesha and iSCSI packages to the daemon image to meet this new
contract.

Checklist:


src/daemon/__DAEMON_PACKAGES__
```diff
@@ -1,7 +1,13 @@
...
+          e2fsprogs \
+          __ETCD_PACKAGE__ \
           __FOREGO_PACKAGE__ \
+          __GANESHA_PACKAGES__ \
+          __ISCSI_PACKAGES__ \
+          kmod \
           __KUBECTL_PACKAGE__ \
-          __ETCD_PACKAGE__ \
-          __S3CMD_PACKAGE__
+          lvm2 \
```
leseb: `lvm2` is needed by `ceph-volume`.

BlaineEXE: If it is needed, it is installed by `yum`, as it is a tracked dependency.

leseb: I don't think it is. @ktdreyer or is it?
If we have that here, I guess this is for a good reason :)

BlaineEXE:
> Aren't iSCSI packages somewhat Ceph binaries? You will need them if you want to deploy iSCSI gateways. For Ganesha, it's somewhat different but a similar too. Isn't Rook going to deploy these at some point too?

The discussion I had on Thursday with Travis was that it seems better to not force users to have iSCSI or Ganesha packages when all they may want is Ceph. The most minimal Ceph base image is Ceph and nothing extra. For Rook's part, it would desire the option to have Ganesha and iSCSI in separate, independent, and decoupled images.

src/daemon/__DAEMON_PACKAGES__
ErwanAliasr1: +1

src/daemon-base/__CEPH_BASE_PACKAGES__
```diff
@@ -1,14 +1,8 @@
 \
-        ca-certificates \
-        e2fsprogs \
         ceph-common__ENV_[CEPH_POINT_RELEASE]__  \
         ceph-mon__ENV_[CEPH_POINT_RELEASE]__  \
         ceph-osd__ENV_[CEPH_POINT_RELEASE]__ \
         ceph-mds__ENV_[CEPH_POINT_RELEASE]__ \
         rbd-mirror__ENV_[CEPH_POINT_RELEASE]__  \
-        __CEPH_MGR_PACKAGE__\
-        kmod \
```
ErwanAliasr1: modprobe & modinfo can be called by ceph

leseb:
@BlaineEXE but iSCSI and Ganesha are somehow part of Ceph, especially iSCSI. So I don't see why these should be leftover. They have dependencies on Ceph libraries so I don't see why they need to be part of another image. This does not seem to be necessary.

@travisn, any thoughts?

src/daemon/__DAEMON_PACKAGES__
ErwanAliasr1: ceph-osd ask for lvm2. So that's safe.

src/daemon-base/__CEPH_BASE_PACKAGES__
BlaineEXE: I'm pretty sure that when I checked, `kmod` wasn't a dependency for installing ceph binaries. This to me would seem to be a bug in Ceph that it doesn't have these as dependencies if they are needed rather than a bug in ceph-container.

BlaineEXE:
RE: @leseb 
> @BlaineEXE but iSCSI and Ganesha are somehow part of Ceph, especially iSCSI. So I don't see why these should be leftover. They have dependencies on Ceph libraries so I don't see why they need to be part of another image. This does not seem to be necessary.

I think that it is still worthwhile to have a minimalist base image that is only ceph w/o ganesha and iSCSI. iSCSI may be dependent on Ceph, but Ceph is still not dependent on iSCSI.

There is another possible option which is to make ceph-container build 3 images. The new, minimalist, `ceph` image; `daemon-base` which installs iSCSI and Ganesha on top of `ceph` (effectively the same as now), and the existing `daemon` which builds on top of `daemon-base`. And FWIW, I wrote the makefile for `ceph-container` such that it should be fairly trivial to add an additional image there.

src/daemon-base/__CEPH_BASE_PACKAGES__
ErwanAliasr1: @ktdreyer What's your though on ceph.spec which doesn't requires kmod while the source code does some modprobe or modinfo ?

ErwanAliasr1:
The definition of a "minimum" image really depends on projects. Some could even need a smaller image that what you request here. So depending on the 3rd party project, the concept of minimum will vary a lot.

So the choice made here was to have a base image which contains the Ceph components + iscsi/ganesha so any "regular" of Ceph is possible without having to derive that base image too much. These components are linked by the version & the distribution so it make sense having them glued together.

@travisn, @BlaineEXE : what is the real issue having those two components added in the base image of Ceph ? These packages aren't big, I don't get why its a show stopper for you ?

travisn:
From Rook's perspective for the [design](https://github.com/rook/rook/pull/2014) to decouple the version of ceph from the version of Rook, there are really just two basic requirements: 
- The rook operator container would only contain rook and the Ceph client tools
- The Ceph container would contain the daemons, which is this discussion.

For the daemon container, it would certainly be ideal to have a minimal container where nothing else is included. If you take this argument too far, you would have a different container for each daemon. At the other end of the spectrum, we could use one container for all the daemons, including nfs and iscsi. In the middle, we could have a daemon container, then have another container layered on top for nfs, iscsi, and other non-core functionality. However, the last option would complicate things in Rook because it would require either another image name to be specified in the CRD, or Rook would have to guess the name of the image. This also introduces a potential failure point if the wrong version of the image is used. The more I think about that complication, it would be nice to stay with a single daemon image consumed by rook and we avoid the versioning issue. This will soon be an issue with native [NFS support](https://github.com/rook/rook/pull/1800) coming.

@ErwanAliasr1 do you have an estimate on how much those components add to the image? @BlaineEXE any concerns about a single daemon image if it isn't much different in size from splitting it out?

ErwanAliasr1:
@travisn I did a build without ISCSI and we drop from 663 to 643 (20MB) and removing NFS removes an extra 6 MB.

So both means 26MB. I don't think it worth removing them and keep the actual split between daemon-base & daemon. Renaming the daemon-base would be interesting to get a "ceph" name in it.

BlaineEXE:
@travisn @ErwanAliasr1 

I think that there are 2 ways to do NFS. One of them is to use an NFS bundled in the Ceph container, as seems to be implied by https://github.com/rook/rook/pull/1800/. But the other option is to do as https://github.com/rook/rook/pull/1783/files does and create a persistent volume claim against Ceph and then expose NFS via its own NFS container. The second option allows Rook to deploy NFS for any of its backends, whereas the first would allow it to deploy NFS only for Ceph. I'm not an NFS expert; there may be some performance loss by making a PVC to turn it into NFS compared to within a Rook-Ceph deployment. But it seems to me a longer-term solution to use the PVC method, and it has the benefits that the NFS container can be independently versioned itself.

I suppose this discussion is separate from whether or not to minimize the Ceph image.

src/daemon-base/__CEPH_BASE_PACKAGES__
ktdreyer: Where does it do that?

PR #1167, opened 2018-08-21
Removing Ubuntu

ErwanAliasr1:
Since there is strictly no support of Ubuntu on this project and as Centos is now the default since a while, let's remove all this dead code.

This removes the tests, the documentation & the code base but also the CI runs we perform on this platform.


ErwanAliasr1:
@BlaineEXE We'll make a release for it and the Readme also report that information.

PR #1170, opened 2018-08-22
[skip ci] ceph-releases: Removing Jewel Support

ErwanAliasr1:
Since Jewel is EOL since July 2018 (http://docs.ceph.com/docs/master/releases/schedule/), let's remove Jewel support from master.
If users needs a jewel build, they will use the actual stable branches.


src/daemon/demo.sh
```diff
@@ -6,7 +6,7 @@ unset "DAEMON_OPTS[${#DAEMON_OPTS[@]}-1]" # remove the last element of the array
 PREPARE_OSD=$OSD_PATH
 ACTIVATE_OSD=$OSD_PATH
 # the following ceph version can start with a numerical value where the new ones need a proper name
-if [[ "$CEPH_VERSION" == "jewel" || "$CEPH_VERSION" == "kraken" || "$CEPH_VERSION" == "luminous" ]]; then
+if [[ "$CEPH_VERSION" == "kraken" || "$CEPH_VERSION" == "luminous" ]]; then
```
leseb: I think this should stay, just in case.

src/daemon/demo.sh
```diff
@@ -48,7 +48,7 @@ fi
 : "${RGW_FRONTEND:="$RGW_FRONTEND_TYPE $RGW_FRONTED_OPTIONS"}"
 
 if [[ "$RGW_FRONTEND_TYPE" == "beast" ]]; then
-  if [[ "$CEPH_VERSION" == "jewel" || "$CEPH_VERSION" == "kraken" || "$CEPH_VERSION" == "luminous" ]]; then
+  if [[ "$CEPH_VERSION" == "kraken" || "$CEPH_VERSION" == "luminous" ]]; then
```
leseb: ditto

src/daemon/demo.sh
```diff
@@ -110,7 +110,7 @@ function bootstrap_osd {
     if ! grep -qE "osd data = $OSD_PATH" /etc/ceph/"${CLUSTER}".conf; then
       echo "osd data = $OSD_PATH" >> /etc/ceph/"${CLUSTER}".conf
     fi
-    if [[ "$CEPH_VERSION" != "jewel" && "$CEPH_VERSION" != "kraken" ]]; then
+    if [[ "$CEPH_VERSION" != "kraken" ]]; then
```
leseb: ditto

src/daemon/demo.sh
```diff
@@ -131,7 +131,7 @@ function bootstrap_osd {
   # start OSD
   chown --verbose -R ceph. "$OSD_PATH"
   ceph-osd "${DAEMON_OPTS[@]}" -i 0
-  if [[ "$CEPH_VERSION" == "jewel" || "$CEPH_VERSION" == "kraken" ]]; then
+  if [[ "$CEPH_VERSION" == "kraken" ]]; then
```
leseb: ditto

src/daemon/demo.sh
```diff
@@ -337,7 +337,7 @@ function build_bootstrap {
   # this is will prevent someone writing daemons in the wrong order
   # once both mon and mgr are up we don't care about the order
   bootstrap_mon
-  if [[ "$CEPH_VERSION" != "jewel" && "$CEPH_VERSION" != "kraken" ]]; then
+  if [[ "$CEPH_VERSION" != "kraken" ]]; then
```
leseb: ditto

src/daemon/start_mon.sh
```diff
@@ -171,7 +171,7 @@ function start_mon {
     /usr/bin/ceph-mon "${DAEMON_OPTS[@]}" -i "${MON_NAME}" --mon-data "$MON_DATA_DIR" --public-addr "${MON_IP}:6789"
   else
     # enable cluster/audit/mon logs on the same stream
-    if [[ "$CEPH_VERSION" != "jewel" && "$CEPH_VERSION" != "kraken" ]]; then
+    if [[ "$CEPH_VERSION" != "kraken" ]]; then
```
leseb: ditto

tox.ini
```diff
@@ -1,6 +1,5 @@
 [tox]
-envlist = {ceph_ansible,ceph_ansible2.2}-{jewel}-{centos7}-{cluster,filestore_osds_container}
-	{ceph_ansible,ceph_ansible2.3}-{luminous}-{centos7}-{cluster,filestore_osds_container,bluestore_osds_container,docker_cluster_collocation}
+envlist = {ceph_ansible,ceph_ansible2.3}-{luminous}-{centos7}-{cluster,filestore_osds_container,bluestore_osds_container,docker_cluster_collocation}
```
leseb: Not sure about this, let's keep it around and wait for @guits input regarding the CI.

src/daemon/demo.sh
ErwanAliasr1: As long as jewel is removed from master, it is no more possible to run that code inside a jewel container. So keeping dead code isn't a good idea. Let's remove what is deprecated. If some wants some jewel, let's use the stable branches then.

src/daemon/demo.sh
ErwanAliasr1: ditto

src/daemon/demo.sh
ErwanAliasr1: ditto

src/daemon/start_mon.sh
ErwanAliasr1: ditto

tox.ini
ErwanAliasr1: Let's wait for @guits but if we don't build the jewel image on master, I don't see any valid reason trying to test it.

Maybe we should replace jewel by mimic instead of removing jewel ?

src/daemon/demo.sh
ErwanAliasr1: Ack to resolve this part ?

tox.ini
ErwanAliasr1: @guits Any concerns with this patch ?

guits: @ErwanAliasr1 +1 replacing jewel by mimic.
@leseb Does that mean we should drop jewel support from ceph-ansible as well?

ErwanAliasr1: @guits What should be the envlist for jewel then ?
Jewel our ceph-ansible master sounds like strange for me while it's perfect in the stable branches.

leseb: Agree with @ErwanAliasr1 Jewel will remain in stable branch (3.0 and 3.1) where current master (future 3.2) shouldn't need to support Jewel.

guits: @ErwanAliasr1 @leseb that's correct

ErwanAliasr1: @guits @leseb If you agree with the current state of the PR, can you approve it ?

ErwanAliasr1:
Just rebased it to fix a conflict

tox.ini
```diff
@@ -25,7 +24,6 @@ setenv=
   filestore_osds_container: REGISTRY_ADDRESS = 192.168.55.1:5000
   bluestore_osds_container: REGISTRY_ADDRESS = 192.168.35.1:5000
   centos7: IMAGE_DISTRO = centos/7
-  jewel: CEPH_STABLE_RELEASE = jewel
   luminous: CEPH_STABLE_RELEASE = luminous
   ceph_ansible: CEPH_ANSIBLE_BRANCH = master
   ceph_ansible2.2: CEPH_ANSIBLE_BRANCH = stable-2.2
```
guits: shouldn't we remove this since we removed it from `envlist` ?

guits: (same thing at L34)

ErwanAliasr1: Yes. Fixed. Thanks.

src/daemon/demo.sh
leseb: Yes!

PR #1172, opened 2018-08-22
[skip ci] demo: create rbd pool when bootstrapping osd

leseb:
When demo deploys an OSD (100%) of the time it'll also create a RBD
pool.
Default name is 'rbd', but you can use -e RBD_POOL=toto to change it.

Closes: https://github.com/ceph/ceph-container/issues/1164


frivoire:
Tested and matchs exactly the feature request (https://github.com/ceph/ceph-container/issues/1164)

PR #1173, opened 2018-08-22
tests: set owner of staging dir to jenkins user

### LINKED ISSUE #1048: staging dir owned by root:root
ceph-container nightly Jenkins jobs are failing on static slaves because the staging dir gets created and is owned (recursively) by root:root.  The jobs fail immediately when they try to wipe out the workspace.

I'm wondering if the staging dir needs to be owned by root.  If not, I will try to find a way to use sudo to clean it up.

E.g., https://2.jenkins.ceph.com/job/ceph-container-nightly-ceph_ansible-luminous-xenial-bluestore_osds_container/178/console

djgalloway:
Description of your changes: Sets owner of `$CEPH_ANSIBLE_SCENARIO_PATH/staging` to jenkins user so the slaves can wipe the workspace before each job.

Resolves #1048 

Checklist:

### DIFF
--- a/tests/teardown.sh
+++ b/tests/teardown.sh
@@ -4,4 +4,6 @@
 . "$WORKSPACE/.tox_vars"
 cd "$CEPH_ANSIBLE_SCENARIO_PATH" || exit 1
 vagrant destroy --force
+# see https://github.com/ceph/ceph-container/issues/1048
+sudo chown -R "$(whoami)" ./staging || true
 cd "$WORKSPACE" || exit


tests/teardown.sh
```diff
@@ -4,4 +4,6 @@
 . "$WORKSPACE/.tox_vars"
 cd "$CEPH_ANSIBLE_SCENARIO_PATH" || exit 1
 vagrant destroy --force
+# see https://github.com/ceph/ceph-container/issues/1048
+sudo chown -R "$(whoami)" ./staging || true
```
leseb: why `|| true`?

ErwanAliasr1: @GregMeno If you ping me on a bug I'd love having the time to wakeup and comment before you actually merge without any feedback.

The teardown is about cleaning and the makefile have some cleaning routines. So a sudo make clean should be enough.

PR #1193, opened 2018-10-02
[skip ci] contrib: fix build ceph base

leseb:
Add a meaningful error message if ARCH is not provided when executing
the script.

### DIFF
--- a/contrib/ceph-build-config.sh
+++ b/contrib/ceph-build-config.sh
@@ -536,3 +536,10 @@ function intersect_lists () {
   # Use sorted_a and _b as inputs to comm, which effectively returns the intersection of the lists
   comm -12 <(echo "${sorted_a}") <(echo "${sorted_b}")
 }
+
+function install_docker {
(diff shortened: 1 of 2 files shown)


contrib/ceph-build-config.sh
```diff
@@ -11,7 +11,7 @@ AARCH64_FLAVORS_TO_BUILD="luminous,centos,7 mimic,centos,7"
 # Allow running this script with the env var ARCH='aarch64' to build arm images
 # ARCH='x86_64'
 # ARCH='aarch64'
-ARCH=${ARCH}  # This is here to prevent shellcheck errors about ARCH not being defined
+: "${ARCH:?must be declared with either x86_64 or aarch64!}"
```
ErwanAliasr1: I'm not really sure to get how it's supposed to work.

leseb: It's a bit redundant with `-u` but will raise a more meaningful message so if `ARCH` does not have any value/not declared the error message is raised and the script fails.

PR #1216, opened 2018-10-22
[skip ci] contrib: ability to override BUILD_SERVER_GOARCH

leseb:
So we can attempt to run the manifest script on arm64.


BlaineEXE:
I'm not sure the exact reason why you want to run the manifest script on
arm. I fear there may be some confusion. The manifest script only needs to
be run once after both the arm and and builds have finished. It can be run
from either arm or x86, but it doesn't need to be run on both. I assumed
x86 when I wrote the script, because that seemed easiest. The manifests
don't hold an intrinsic architecture of their own.

PR #1219, opened 2018-10-24
[skip ci] mon: create bootstrap keys in //

leseb:
This will speed up the bootstrap of the initial monitor since we are now
using a loop and pile up all the creations and then wait for them to be
complete.
Also, there is no need to check for the existence of the keys since this
code is only triggered when a new monitor is detected.

Description of your changes:

Resolves #

Checklist:


leseb:
Closed for https://github.com/ceph/ceph-container/pull/1238

PR #1222, opened 2018-10-25
contrib: do not build master on 'base' builds

leseb:
Since 4cab875f888cb7933423430bd7b80e4792c2115a we fail for return code
!= 0 in subshell (-E option), thus when trying to fetch the repo for the
'master' release this fails since master packages do not live in
download.ceph.com. Building master was introduced in
c8c1931c8dc575495607fea7794d5aaab3b6f457 and I'm partially reverting it.
In the first, the base build script should have not been modified.


leseb:
Because of that daily base jobs are failing, https://2.jenkins.ceph.com/job/ceph-container-build-ceph-base-push-imgs/32/consoleFull

@BlaineEXE I believe, it'll be nice to have the ability to push base images too for master. I just can't estimate how much effort this involves, I'm merging this as is to unlock the CI but would like your inputs as well, thanks!

PR #1223, opened 2018-10-25
contrib: declare DRY_RUN empty

leseb:
We must declare DRY_RUN but not assign anything, if we don't the script
will fail because of the -u option.

Fix /home/jenkins-build/build/workspace/ceph-container-build-ceph-base-push-imgs/ceph-container/contrib/ceph-build-config.sh: line 464: DRY_RUN: unbound variable


leseb:
Follow up on the addition of `-E`.

leseb:
@BlaineEXE ^

PR #1226, opened 2018-10-29
start_mon: fix NETWORK_AUTO_DETECT=1 IPv4 lookup

leseb:
`get_ip` incorrently returned string `[]` when IPv6 detection
failed, leading to skipped IPv4 auto detection and unknown
address to continue.


leseb:
https://github.com/ceph/ceph-container/pull/1109

PR #1227, opened 2018-10-30
Make it explicit which architectures are supported

tbreeds:
Exit early if trying to run the daemon container on untested
architectures.  This has been written such that it's clear what's
supported but also so it can be tested trivially without requiring a
rebuild.

Description of your changes:

Checklist:

### DIFF
--- a/src/daemon/entrypoint.sh.in
+++ b/src/daemon/entrypoint.sh.in
@@ -40,6 +40,11 @@ CEPH_DAEMON=$(to_lowercase "${CEPH_DAEMON}")
 
 create_mandatory_directories
 
+if [[ ! "x86_64 aarch64" =~ $CEPH_ARCH  ]] ; then
+    echo "$CEPH_DAEMON is not supported on $CEPH_ARCH" >&2
+    exit 1
+fi
+
 # If we are given a valid first argument, set the
 # CEPH_DAEMON variable from it
 case "$CEPH_DAEMON" in
--- a/src/daemon/variables_entrypoint.sh
+++ b/src/daemon/variables_entrypoint.sh
@@ -65,6 +65,7 @@ fi
 : "${GANESHA_OPTIONS:=""}"
 : "${GANESHA_EPOCH:=""}" # For restarting
 : "${MGR_IP:=0.0.0.0}"
+: "${CEPH_ARCH:=$(uname -m)}"
 
 # Make sure to change the value of one another if user changes some of the default values
 while read -r line; do


yselkowitz:
What else is needed here?

tbreeds:
It looks like the CI is prompting for a password which is why it's timing out.  Is there something wrong with: https://github.com/ceph/ceph-container/blob/master/.travis.yml#L19

yselkowitz:
This allows the container to be used for its client bits (ceph-common) on arches where Ceph server is not yet supported, such as on IBM Power LE in conjunction with OpenStack deployment (which uses this container via ceph-ansible).

tbreeds:
> I'm not sure to get what was the starting point of this. Can you detail the case you had that brings you in this situation ?

Currently Ceph-ansible uses this container unconditionally (when using site-docker.yaml) for the client only bits the container is created and the the entry point overridden to run specific ceph commands

- [get ceph version](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-container-common/tasks/main.yml#L32-L39)
- [create client keys](https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-client/tasks/create_users_keys.yml)
 We added a filter to ensure that the dummy container is only run on x86_64 but that has limitations that we're trying to fix with this

The idea behind this change is to create setup the container so it can be used for the client bits on ppc64le, and facilitate testing without stating that the daemon processes have been tested. 

> I mean if you start the container on a different arch, I wonder how it could start the container.

Just calling ceph-ansible with an inventory that contains a mix of CPU architectures

ErwanAliasr1:
@tbreeds I surely miss something here. How can an x86|arm build could even start on ppc64 ?

yselkowitz:
It won't, and that's not what is being guarded here.  The aim is to build this container for ppc64le but only allow use of the client bits thereon for the ceph-ansible case mentioned above.

tbreeds:
> @tbreeds I surely miss something here. How can an x86|arm build could even start on ppc64 ?

@ErwanAliasr1 As @yselkowitz states, an arm or x86_64 build itself can't be started on ppc64le but what we're tryign to setup is the ability to build the container on the existing platforms (arm, x86_64) and add ppc64le  *without* enabling the daemons on ppc64le.  Then we can build the container on the 3 platforms without altering the functionality on arm or x86_64 and giving us the ability to run the client bits on ppc64le.

ErwanAliasr1:
@tbreeds so to be sure I got your idea. You want to build ceph-container for the 3 arches(ppc/arm/x86) but make the ppc one returning immediately.
Then, what is the purpose of building the ppc if it doesn't do anything ?

yselkowitz:
This commit hobbles the entrypoint script so that customers cannot try to deploy the server through the usual means, since that is not supported on POWER.  ceph-ansible does not use the entrypoint script, it bypasses that and calls the necessary (client) binaries directly.

tbreeds:
> @tbreeds so to be sure I got your idea. You want to build ceph-container for the 3 arches(ppc/arm/x86) but make the ppc one returning immediately.
> Then, what is the purpose of building the ppc if it doesn't do anything ?

The main use case is ceph-ansible which will call into the container with an explicit ``--entrypoint`` to run ceph commands directly.  The links previously supplied show how/where this is done.

I agree this is somewhat strange but this is the only was we can see to create a client that fulfil that role *without* making an implied statement that ppc64le is tested in the community *and* also not make that testing impossible

ktdreyer:
To summarize: we want to be able to run the Ceph client pieces on ppc64le, but we don't want to give an impression that the server pieces are supported on ppc64le.

tbreeds:
> To summarize: we want to be able to run the Ceph client pieces on ppc64le, but we don't want to give an impression that the server pieces are supported on ppc64le.

Correct!

leseb:
@tbreeds @yselkowitz thanks, for your patience and clarification.

PR #1232, opened 2018-11-02
nano: add static sree

leseb:
We don't have a package yet, the code hasn't changed for months, the
tarball is under 500KB so let's add it to the daemon image.
This avoids downloading the tar each time we boot a new cn container and
also brings back the ability to work offline or behind a proxy.

Closes: https://github.com/ceph/cn/issues/72


src/daemon/demo.sh
```diff
@@ -302,30 +302,23 @@ function bootstrap_mgr {
...
 ########
 function bootstrap_sree {
-  if [ ! -f sree.tar.gz ]; then
-    if [ -z "$SREE_VERSION" ]; then
-      sree_latest=$(curl -s 'https://api.github.com/repos/leseb/Sree/releases/latest' | grep tarball_url | cut -d '"' -f 4)
-      curl -L "$sree_latest" -o sree.tar.gz
-    else
-      curl -L https://github.com/leseb/Sree/archive/"$SREE_VERSION".tar.gz -o sree.tar.gz
-    fi
-    mkdir sree && tar xzvf sree.tar.gz -C sree --strip-components 1
```
ErwanAliasr1: Why do you remove the tar xvzf ?
I don't get how you extract the tarball then.

maint-lib/stagelib/replace.py
```diff
@@ -118,6 +118,11 @@ def do_variable_replace(replace_root_dir):
     logging.info('    Replacing variables')
     for dirname, subdirs, files in os.walk(replace_root_dir, topdown=True):
         for f in files:
+            # do not process the tarball
+            # this prevents the following error:
+            # UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
+            if "Sree-0.1.tar.gz" in f:
```
ErwanAliasr1: If "Sree-" would be more generic and work across releases if you update it later

src/daemon/demo.sh
```diff
@@ -302,30 +302,23 @@ function bootstrap_mgr {
...
-
+  if [ ! -f /Sree-0.1/sree_done ]; then
     ACCESS_KEY=$(awk '/Access key/ {print $3}' /ceph-demo-user)
     SECRET_KEY=$(awk '/Secret key/ {print $3}' /ceph-demo-user)
 
-    pushd sree
+    pushd Sree-0.1
```
ErwanAliasr1: I'm not version found of the versionned tree. Why not extracting the tarball in a fixed name dir as you did before ? This way, we have fewer code to modify when we update Sree.

src/daemon/demo.sh
```diff
@@ -302,30 +302,23 @@ function bootstrap_mgr {
...
     mv sree.cfg.sample sree.cfg
     sed -i "s/RGW_CIVETWEB_PORT_VALUE/$RGW_CIVETWEB_PORT/" sree.cfg
     sed -i "s/SREE_PORT_VALUE/$SREE_PORT/" sree.cfg
+    touch sree_done
     popd
   fi
 
   # start Sree
-  pushd sree
+  pushd Sree-0.1
```
ErwanAliasr1: Ditto

src/daemon/demo.sh
```diff
@@ -302,30 +302,23 @@ function bootstrap_mgr {
...
-  if [ ! -f sree.tar.gz ]; then
-    if [ -z "$SREE_VERSION" ]; then
-      sree_latest=$(curl -s 'https://api.github.com/repos/leseb/Sree/releases/latest' | grep tarball_url | cut -d '"' -f 4)
-      curl -L "$sree_latest" -o sree.tar.gz
-    else
-      curl -L https://github.com/leseb/Sree/archive/"$SREE_VERSION".tar.gz -o sree.tar.gz
-    fi
-    mkdir sree && tar xzvf sree.tar.gz -C sree --strip-components 1
-
+  if [ ! -f /Sree-0.1/sree_done ]; then
```
ErwanAliasr1: Same here (about versionned dir)

src/daemon/demo.sh
leseb: The tarball gets extracted by Docker during build time .

leseb: New approach please have a look.

leseb: New approach please have a look.

PR #1233, opened 2018-11-02
[skip ci] Ceph image dockerhub descriptions

BlaineEXE:
Add an .md file containing description text for the `ceph` images now
published on DockerHub.

@leseb, could you update the Docker Hub description of the `ceph/ceph-{,amd64,arm64}` images based on the text here once the text is approved? This file needn't make it into the ceph-container source, as the description on Docker Hub is the more important item; however, I think it might be a good idea to have this versioned.

### DIFF
--- a/contrib/ceph-versioned-image-descriptions.md
+++ b/contrib/ceph-versioned-image-descriptions.md
@@ -1,0 +1,55 @@
+DockerHub descriptions for ceph-vX.Y.Z images
+==============================================
+
+ceph/ceph
+----------
+
+### Short Description
+Images containing all Ceph binaries.
+
+### Full Description
+Images contain all Ceph binaries as well as NFS-Ganesha and iSCSI binaries since these components
+are heavily tied to Ceph's version to be compatible. Unless otherwise noted in the tag, all images
+are CentOS based. New images are built within 24 hours of a new Ceph version being published to
+the official Ceph package repository.
+
+#### Image tag breakdown
+
+**Build date** </br>
+Some images are suffixed with an 8-digit build date in the form `YYYMMDD`. This indicates the date
+the image was built and is used similarly to a build number for typical packages. For an image which
+already exists (e.g., `v12.2.7-20181023`), a new version with a different build date will be built
+when there is a base image update. For example, if the CentOS base image gets a security fix on
+10 February 2080, the example image above will get a new image built with tag `v12.2.7-20800210`.
+
+**Versions** </br>
+There are a few ways to choose the Ceph version you desire:
+ - Full semantic version with build date, e.g., `v12.2.9-20181026`
+   - These tags are intended for use when precise control over image upgrade is desired and are
+     recommended for production use.
+ - Major version, e.g. `v12` (a.k.a. _Ceph Luminous_)
+   - These tags are always the most recent build of the newest Ceph major release matching the tag.
+ - Minor version (e.g., `v12.1`)
+   - These tags are always the most recent build of the newest Ceph minor release matching the tag
+     for environments where more precise control is needed than a major version but where bug fixes
+     both in Ceph and in the base image are desired.
+
+#### Image architecture
+These images are manifests which will pull amd64 or arm64 images automatically based on the
+architecture of the host system.
+
+#### Image source
+Images are built from the [ceph/ceph-container](https://github.com/ceph/ceph-container) project on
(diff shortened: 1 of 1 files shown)


BlaineEXE:
@tnielsen @dmick - thoughts?

leseb:
LGTM, but @travisn or @dmick might have things to say :).

contrib/ceph-versioned-image-descriptions.md
```diff
@@ -0,0 +1,55 @@
...
+10 February 2080, the example image above will get a new image built with tag `v12.2.7-20800210`.
+
+**Versions** </br>
+There are a few ways to choose the Ceph version you desire:
+ - Full semantic version with build date, e.g., `v12.2.9-20181026`
+   - These tags are intended for use when precise control over image upgrade is desired and are
+     recommended for production use.
+ - Major version, e.g. `v12` (a.k.a. _Ceph Luminous_)
+   - These tags are always the most recent build of the newest Ceph major release matching the tag.
+ - Minor version (e.g., `v12.1`)
```
travisn: seems like we don't need the `v12.1` or `v12.2` tags. With ceph's versioning scheme, all the stable releases have the minor version of `.2`. Thus, `v12` would already cover the latest stable release and would be redundant with the tag `v12.2`.

contrib/ceph-versioned-image-descriptions.md
```diff
@@ -0,0 +1,55 @@
...
+   - These tags are always the most recent build of the newest Ceph minor release matching the tag
+     for environments where more precise control is needed than a major version but where bug fixes
+     both in Ceph and in the base image are desired.
+
```
travisn: what about the tag `v12.2.9`, which represents the latest build for that patch release? So if we build that patch again in the future with a different base os update, this tag can still be used. Not high priority, but seems like a nice-to-have.

BlaineEXE: Tags that are just `v12.2.9` don't exist, my assumption being if a user wants to be that specific about the Ceph version, they likely want to be that specific about the build version. This can be changed, but that involves code changes.

contrib/ceph-versioned-image-descriptions.md
BlaineEXE: I agree that they don't seem terribly useful. But they exist and so are documented here.

contrib/ceph-versioned-image-descriptions.md
dmick: Those are the public versions that everyone uses.  I'm baffled as to why they're not the common case.

leseb:
@dmick why did you close this?

dmick:
No intention to do so, I must have hit a wrong button somehow.  Bizarre.

contrib/ceph-versioned-image-descriptions.md
BlaineEXE: This, again, can be changed, but is out of scope here, and this seems to be delaying us from getting documentation of any kind on these images on Docker Hub.

dmick: I don't see why it's delaying putting the doc up, but, I was asked for comments, and I didn't believe those comments needed to be limited to just the documentation.  Issues are issues, regardless of the particular file they arise from.

Since the subject has come up, however, I believe the current tagging system is so deficient and confusingly done that I believe it would be better to fix the tagging system than to document it, as broken as it is.  Others may, of course, override my opinion, and I'm not going to fight to the death about it.

leseb:
@BlaineEXE @dmick @travisn can we go ahead with this? Or does anyone have something else to add?

dmick:
My preference is to fix the tagging and then document; if that's not palatable then I'd probably not document the questionable tags.  But of course I'm just one opinion. Regardless, there should be an issue capturing the questions/suggestions above.

leseb:
@BlaineEXE @travisn any plan to change the tagging system? Thanks

BlaineEXE:
No immediate plans to change it. I don't see that anything here is so
radically deficient as to merit delaying documentation considering people
are going to start using the images yesterday after the release of Rook
v0.9 (three days ago).

The pull request for the script that built and tagged the images was open
for over a month, and this feedback wasn't brought up with such language as
"confusing" and "deficient" at that time.

Blaine

leseb:
@BlaineEXE alright, merging then! Thanks!

travisn:
Great to have this documented now, thanks. I'm just slow to the party!

PR #1236, opened 2018-11-05
rhel7: enable ODCS

yselkowitz:
Signed-off-by: Yaakov Selkowitz <yselkowi@redhat.com>

Description of your changes:
Allow building rhel7 containers with ODCS (iow w/o --repo-url).

Checklist:
- [n/a] Documentation has been updated, if necessary.
- [n/a] Pending release notes updated with breaking and/or notable changes, if necessary.

cc: @ktdreyer

### DIFF
--- a/ceph-releases/ALL/rhel7/daemon/container.yaml
+++ b/ceph-releases/ALL/rhel7/daemon/container.yaml
@@ -1,0 +1,17 @@
+compose:
+  packages:
+  - ceph-common
+  - ceph-mon
+  - ceph-osd
+  - ceph-mds
+  - rbd-mirror
+  - ceph-mgr
+  - ceph-radosgw
+  - libradosstriper1
+  - nfs-ganesha
+  - nfs-ganesha-ceph
+  - nfs-ganesha-rgw
+  - tcmu-runner
+  - ceph-iscsi-config
+  - ceph-iscsi-cli
+  pulp_repos: true


leseb:
@yselkowitz are you going to do the sync downstream too?

ceph-releases/ALL/rhel7/daemon/container.yaml
```diff
@@ -0,0 +1,17 @@
+compose:
+  packages:
```
ktdreyer: Keeping this list up-to-date by hand is going to be painful. I'd rather we omitted the `packages` list entirely here and composed the entire contents of the Brew tag. It should not take that much longer, and it's computer time vs the human cost of keeping this updated.

If we had something that could parse our Dockerfile and determine this list automatically from "yum install" statements, that would be great, but that is not easy.

yselkowitz: Without compose:packages but with pulp_repos:true, only content sets (IOW released content) will be pulled in.  This would mean that the RPMs -- even the Ceph ones -- would have to be *released* before you could even *build* a container with them.  Therefore, using ODCS is not practical without compose:packages.

The other alternative to a repo file is to rely solely on the build tag inheritance, which would need to be adjusted even more.

yselkowitz:
Some tooling changes need to happen before this will work as intended without a specific list.

ktdreyer:
We can now specify `packages: []` in the `compose` key, and ODCS will compose the entire contents of the build target's build tag. https://osbs.readthedocs.io/en/latest/users.html#compose

ktdreyer:
The OSBS team is eager for us to try ODCS again. Now that we have this `packages: []` feature in production, I've re-submitted this change at https://github.com/ceph/ceph-container/pull/1516

PR #1237, opened 2018-11-05
rbd-mirror: add new key

leseb:
On Nautilus, we now have a new key for rbd-mirror so let's use it
instead of the rbd key. It's basically the same but the rbd-mirror cap
gives the daemon access to read private config-keys from the mon.


leseb:
Comment for myself, I think we should stop generating the keys and let the mon created them and just fetch them once they reach quorum.

src/daemon/config.static.sh
```diff
@@ -96,10 +96,15 @@ ENDHERE
 
   if [ ! -e "$RBD_MIRROR_BOOTSTRAP_KEYRING" ]; then
     # Generate the RBD Mirror bootstrap key
-    ceph-authtool "$RBD_MIRROR_BOOTSTRAP_KEYRING" --create-keyring --gen-key -n client.bootstrap-rbd --cap mon 'allow profile bootstrap-rbd'
+    ceph-authtool "$RBD_MIRROR_BOOTSTRAP_KEYRING" --create-keyring --gen-key -n client.bootstrap-rbd-mirror --cap mon 'allow profile bootstrap-rbd-mirror'
   fi
-    # Apply proper permissions to the keys
-    chown "${CHOWN_OPT[@]}" ceph. "$MON_KEYRING" "$OSD_BOOTSTRAP_KEYRING" "$MDS_BOOTSTRAP_KEYRING" "$RGW_BOOTSTRAP_KEYRING" "$RBD_MIRROR_BOOTSTRAP_KEYRING"
+
+  if [ ! -e "$RBD_BOOTSTRAP_KEYRING" ]; then
```
guits: icitte

PR #1238, opened 2018-11-06
mon: remove key generation

leseb:
As for Nautilus, keys get automatically generated so we don't need to
create them during the initial bootstrap.
We still keep the ability to generate our how admin key though.


leseb:
This is for N and above only so we need to add an exception for L and M.

src/daemon/config.static.sh
```diff
@@ -79,32 +79,8 @@ ENDHERE
...
-
-  if [ ! -e "$RBD_BOOTSTRAP_KEYRING" ]; then
-    # Generate the RBD Mirror bootstrap key
-    ceph-authtool "$RBD_BOOTSTRAP_KEYRING" --create-keyring --gen-key -n client.bootstrap-rbd --cap mon 'allow profile bootstrap-rbd'
-  fi
   # Apply proper permissions to the keys
-  chown "${CHOWN_OPT[@]}" ceph. "$MON_KEYRING" "$OSD_BOOTSTRAP_KEYRING" "$MDS_BOOTSTRAP_KEYRING" "$RGW_BOOTSTRAP_KEYRING" "$RBD_MIRROR_BOOTSTRAP_KEYRING"
+  chown "${CHOWN_OPT[@]}" ceph. "$MON_KEYRING" "$ADMIN_KEYRING"
 
   if [ ! -e "$MONMAP" ]; then
```
leseb: Remove this too?

src/daemon/config.static.sh
```diff
@@ -79,32 +79,8 @@ ENDHERE
...
-    ceph-authtool "$RBD_MIRROR_BOOTSTRAP_KEYRING" --create-keyring --gen-key -n client.bootstrap-rbd-mirror --cap mon 'allow profile bootstrap-rbd-mirror'
-  fi
-
-  if [ ! -e "$RBD_BOOTSTRAP_KEYRING" ]; then
-    # Generate the RBD Mirror bootstrap key
-    ceph-authtool "$RBD_BOOTSTRAP_KEYRING" --create-keyring --gen-key -n client.bootstrap-rbd --cap mon 'allow profile bootstrap-rbd'
-  fi
   # Apply proper permissions to the keys
-  chown "${CHOWN_OPT[@]}" ceph. "$MON_KEYRING" "$OSD_BOOTSTRAP_KEYRING" "$MDS_BOOTSTRAP_KEYRING" "$RGW_BOOTSTRAP_KEYRING" "$RBD_MIRROR_BOOTSTRAP_KEYRING"
+  chown "${CHOWN_OPT[@]}" ceph. "$MON_KEYRING" "$ADMIN_KEYRING"
```
ErwanAliasr1: I just don't get why the ADMIN_KEYRING is poping here while it wasn't in the previous code.
What was it a missing  ?
Can you detail the reason of this change in the commit message please ?

leseb: Actually, you're right, there is no need to add `ADMIN_KEYRING` to this.

src/daemon/demo.sh
```diff
@@ -58,6 +58,12 @@ if [[ "$RGW_FRONTEND_TYPE" == "beast" ]]; then
 fi
 
 
+function fetch_initial_keys {
+  for key in $(ceph -n mon. -k /var/lib/ceph/mon/ceph-mon0/keyring auth list |& grep ^client); do
```
ErwanAliasr1: why mon0 ? isn't like a little bit of hardcoding here ?

src/daemon/demo.sh
```diff
@@ -58,6 +58,12 @@ if [[ "$RGW_FRONTEND_TYPE" == "beast" ]]; then
 fi
 
 
+function fetch_initial_keys {
```
ErwanAliasr1: for keyring in /var/lib/ceph/mon/*/keyring; do 
for key in $(ceph -n mon. -k $keyring  auth list |& grep ^client); do
 ceph -n mon. -k $keyring auth get "$key" -o "$key"

src/daemon/demo.sh
leseb: Yeah this is wrong :)

src/daemon/demo.sh
```diff
@@ -58,6 +58,19 @@ if [[ "$RGW_FRONTEND_TYPE" == "beast" ]]; then
...
+  # This is only valid starting from nautilus
+  if [[ "$CEPH_VERSION" == "luminous" || "$CEPH_VERSION" == "mimic" ]]; then
```
ErwanAliasr1: @leseb just added that to handle the luminous|mimic case

PR #1240, opened 2018-11-08
Pull in nfs-ganesha-rados-grace package

jtlayton:
In order to deploy ganesha in a clustered configuration, we need the ganesha-rados-grace command. That's packaged separately in the nfs-ganesha-rados-grace package.

I'm not sure whether this is the correct fix however, as older versions of ganesha (before v2.7) didn't have this package, and so the installation attempt might fail.

### DIFF
--- a/ceph-releases/luminous/daemon-base/__GANESHA_PACKAGES__
+++ b/ceph-releases/luminous/daemon-base/__GANESHA_PACKAGES__
@@ -1,0 +1,1 @@
+nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw
--- a/ceph-releases/mimic/daemon-base/__GANESHA_PACKAGES__
+++ b/ceph-releases/mimic/daemon-base/__GANESHA_PACKAGES__
@@ -1,0 +1,1 @@
+nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw
--- a/src/daemon-base/__GANESHA_PACKAGES__
+++ b/src/daemon-base/__GANESHA_PACKAGES__
@@ -1,1 +1,1 @@
-nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw
+nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw nfs-ganesha-rados-grace


jtlayton:
Retest this please

jtlayton:
Second attempt: just install the package for "master" and not the other ceph releases. This failed CI test but it's not clear to me why. "make build" on my own machine succeeded.

A more future-proof way to do this might be to add the package to ```./src/daemon-base/__GANESHA_PACKAGES__``` but add override ```__GANESHA_PACKAGES__``` files for all of the older releases.

jtlayton:
Rebase onto newer master.

ErwanAliasr1:
I'm a little bit confused here.
I don't see any version of that package on shaman or download.ceph.com. Where did you found that package ?

Is it available on all OSes we support here ? Centos / Suse / Debian ?

Is this packages should be specific to nautilus and following as your patch suggest ?

jtlayton:
It's here:

http://download.ceph.com/nfs-ganesha/rpm-V2.7-stable/mimic/x86_64/nfs-ganesha-rados-grace-2.7.1-0.1.el7.x86_64.rpm

This package will only be present in nfs-ganesha v2.7 and above as the binary didn't exist before then.

It looks like whoever maintains the debian package hasn't updated their spec to build it.

ErwanAliasr1:
@leseb Are you fine if we upgrade to 2.7 for mimic+ ?

leseb:
I have no idea :/ but I guess it makes sens to have this for Nautilus and above only.

batrick:
I think we just need this for Nautilus. @jtlayton ?

jtlayton:
Yes, nautilus and later. It won't hurt to add it to earlier releases, but it won't be of much use there. It's also only present in ganesha v2.7 and up, so if it's an older version of ganesha it (likely) won't have the package.

ErwanAliasr1:
As long a nautilus = master, we get the packages from shaman, so the package is avail.
As the time nautilus will be a release, we'll have to bump the release which is rpm-V2.6-stable today.

ErwanAliasr1:
@herbert-li @BlaineEXE Any idea on the impacts on your side ?

jtlayton:
> As long as nautilus = master, we get the packages from shaman, so the package is avail.
> As the time nautilus will be a release, we'll have to bump the release which is rpm-V2.6-stable today.

We'll be wanting this in later releases too, so I think the approach I'm proposing, where we add it to the common daemon-base packages but exempt earlier releases is probably the right one. That said, let me know if you think this needs changes.

ErwanAliasr1:
@jtlayton your approach is good. Just wonder the impact on debian & opensuse btw.

jtlayton:
I imagine SuSE will be fine assuming they build packages using the specfile in the nfs-ganesha tree. There's no debian packaging info in there though. so not sure on that end.

jtlayton:
@ErwanAliasr, It appears that the only place ceph-container is building anything besides luminous is for centos7:



We don't really care about this tool for ceph releases earlier than nautilus, so I'm not terribly concerned about anything but centos at the moment. This link shows that the package is present on opensuse, so once there is a flavor in ceph-container for nautilus + opensuse, it should naturally pick this up:

http://download.opensuse.org/repositories/filesystems:/ceph:/nautilus/openSUSE_Tumbleweed/x86_64/

Debian I'm not sure about. I don't know where or how the packaging is done there. I'm happy to help them update it though if someone steps forward.

leseb:
@jtlayton do you need a backport on stable-3.2 for downstream?

jtlayton:
@leseb , I don't think so. We're mostly interested in nautilus and up for this. That said, I haven't followed how ceph-container is versioned vs. everything else, so let me know if you think we should.

leseb:
@jtlayton no backport if it's for N and above.

PR #1248, opened 2018-11-13
daemon-base: install kubernetes python module

jtlayton:
This module is necessary for the ceph-mgr rook orchestrator module,
so we should ensure that it's installed in this case.

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -30,4 +30,6 @@ bash -c ' \
     REPO_URL="http://download.ceph.com/rpm-${CEPH_VERSION}/el__ENV_[BASEOS_TAG]__/"; \
   fi && \
   rpm -Uvh "$REPO_URL/noarch/ceph-release-1-${RELEASE_VER}.el__ENV_[BASEOS_TAG]__.noarch.rpm" ' && \
-yum install -y __CEPH_BASE_PACKAGES__
+yum install -y __CEPH_BASE_PACKAGES__ python-pip && \
+pip install kubernetes && \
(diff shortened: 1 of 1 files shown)


jtlayton:
The k8s module does not seem to be separately packaged for centos 7, unfortunately so we can't just install an RPM there. We could remove pip afterward if that would help?

I know that @jcsp was installing 6.0.0 as a hardcoded version in kubejacker, but the latest version in epel7 seems to work just fine for me. I'd suggest that we avoid specifying a specific version until it's clear that it's necessary.

leseb:
@jtlayton yes let's remove `pip` after the installation of kube client.
Thanks

jtlayton:
The dependency comes from the ceph-mgr rook module. AIUI, rook is moving toward running stock container images for the actual daemons, so I suppose we'll need to have rook do this at runtime after starting the pod.

leseb:
@jtlayton what do you mean by "stock container images for the actual daemons"? Thanks

jtlayton:
Recently the rook project moved to decouple the ceph version from rook itself so that you could select the ceph version to use at runtime:

    https://github.com/rook/rook/blob/master/design/decouple-ceph-version.md

With that, you can specify the images to be used for running the actual ceph daemon pods. Rook doesn't maintain a special image per version, so you just get whatever is in the image to which you point it. The expectation is that most people will use "official" ceph/ceph:* images from the ceph-container project.

leseb:
Oh that's what you meant, ok I'm aware of that then :).
However, these ceph/ceph images are actually daemon-base images, plus the version granularity.

jtlayton:
Yes, but the key point here is that they all lack the python-kubernetes module, so we need to come up with some way to ensure that we can get it if we want the orchestrator to work in these images.

leseb:
Agreed.

PR #1254, opened 2018-11-21
src/daemon/demo.sh: Dynamic sleep instead of harcoded value

ErwanAliasr1:
The current code was waiting for 5 secs before trying to create the bucket.

The whole container usually starts in 25 secs meaning that a 1/5th is spent here.

This patch is poking the s3 service to see if it's available.
The loop is waiting up to 5 seconds.

Once the radosgw is ready, the bucket is created.

Tests on a private server showed that a single loop is enough, saving 5 seconds on the start process.

Description of your changes:

Resolves #

Checklist:

### DIFF
--- a/src/daemon/demo.sh
+++ b/src/daemon/demo.sh
@@ -220,9 +220,9 @@ function bootstrap_demo_user {
 
     if [ -n "$CEPH_DEMO_BUCKET" ]; then
       log "Creating bucket..."
-      # waiting for rgw to be ready, 5 seconds is usually enough
-      sleep 5
-      s3cmd mb s3://"$CEPH_DEMO_BUCKET"
+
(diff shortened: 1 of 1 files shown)


src/daemon/demo.sh
```diff
@@ -220,8 +220,13 @@ function bootstrap_demo_user {
 
     if [ -n "$CEPH_DEMO_BUCKET" ]; then
       log "Creating bucket..."
-      # waiting for rgw to be ready, 5 seconds is usually enough
-      sleep 5
+
+      # waiting up to 5 secs that rgw is ready
+      for i in $(seq 50); do
```
leseb: Can't we do a `until s3cmd mb s3://"$CEPH_DEMO_BUCKET"` instead?

PR #1256, opened 2018-11-21
ceph-disk: bluestore wait for all partition

leseb:
To fix a mutation of the race condition where you create partitions and
udev takes time to inform the kernnel to create the node. We basically
wait for the device to appear and eventually fail if not showing up.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1648010


ErwanAliasr1:
What if the device never appear ? Does this makes an infinite loop ?

leseb:
We fail after 10sec if nothing shows up

Le mer. 21 nov. 2018 à 18:58, Erwan Velu <notifications@github.com> a
écrit :

PR #1261, opened 2018-11-28
ceph_volume_active: auto discover of osd type

leseb:
Instead of requiring OSD_OBJECTSTORE to be set for activation we now
discover the objectstore on the fly.

### DIFF
--- a/src/daemon/osd_scenarios/osd_volume_activate.sh
+++ b/src/daemon/osd_scenarios/osd_volume_activate.sh
@@ -4,16 +4,26 @@ set -e
 function osd_volume_activate {
   : "${OSD_ID:?Give me an OSD ID to activate, eg: -e OSD_ID=0}"
 
+  CEPH_VOLUME_LIST_JSON="$(ceph-volume lvm list --format json)"
+
+  if ! echo "$CEPH_VOLUME_LIST_JSON" | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"])" &> /dev/null; then
+    log "OSD id $OSD_ID does not exist"
+    exit 1
+  fi
+
   # Find the OSD FSID from the OSD ID
-  OSD_FSID="$(ceph-volume lvm list --format json | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"tags\"][\"ceph.osd_fsid\"])")"
+  OSD_FSID="$(echo "$CEPH_VOLUME_LIST_JSON" | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"tags\"][\"ceph.osd_fsid\"])")"
+
+  # Find the OSD type
+  OSD_TYPE="$(echo "$CEPH_VOLUME_LIST_JSON" | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"type\"])")"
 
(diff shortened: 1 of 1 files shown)


src/daemon/osd_scenarios/osd_volume_activate.sh
```diff
@@ -7,13 +7,16 @@ function osd_volume_activate {
   # Find the OSD FSID from the OSD ID
   OSD_FSID="$(ceph-volume lvm list --format json | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"tags\"][\"ceph.osd_fsid\"])")"
 
+  # Find the OSD type
+  OSD_TYPE="$(ceph-volume lvm list --format json | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"type\"])")"
```
ErwanAliasr1: That makes a 2nd call to c-v lvm list with almost the same code.
I would suggest the following.
make a single c-v lvm list call and store the result. Then parse it to extract your two fields while checking they actually exist before printing to avoid a assert.
Why not having a small python script that do that and output the two info you are looking at _or_ report a failure.
That would be more secured & offer options to add more detection & reporting if required later without impact the performance.

ErwanAliasr1: This could be merged as-is if the performance impact isn't too high and update it later so it can be in the release.

src/daemon/osd_scenarios/osd_volume_activate.sh
```diff
@@ -4,16 +4,26 @@ set -e
 function osd_volume_activate {
   : "${OSD_ID:?Give me an OSD ID to activate, eg: -e OSD_ID=0}"
 
+  CEPH_VOLUME_LIST_JSON="$(ceph-volume lvm list --format json)"
+
+  if ! echo "$CEPH_VOLUME_LIST_JSON" | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"]"; then
```
ErwanAliasr1:
```suggestion
  if ! echo "$CEPH_VOLUME_LIST_JSON" | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"])"; then
```

src/daemon/osd_scenarios/osd_volume_activate.sh
```diff
@@ -4,16 +4,26 @@ set -e
...
+    log "OSD id $OSD_ID does not exist"
+    exit 1
+  fi
+
   # Find the OSD FSID from the OSD ID
-  OSD_FSID="$(ceph-volume lvm list --format json | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"tags\"][\"ceph.osd_fsid\"])")"
+  OSD_FSID="$(echo "$CEPH_VOLUME_LIST_JSON" | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"tags\"][\"ceph.osd_fsid\"])")"
+
+  # Find the OSD type
+  OSD_TYPE="$(echo "$CEPH_VOLUME_LIST_JSON" | python -c "import sys, json; print(json.load(sys.stdin)[\"$OSD_ID\"][0][\"type\"])")"
```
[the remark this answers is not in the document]
ErwanAliasr1: Except that, I'm fine.

PR #1272, opened 2018-12-06
contrib: fix bug w/ `comm` in manifest script

BlaineEXE:
There was a bug with the input to `comm` in the manifest creation script causing `comm` to fail with the introduction of Ceph `12.2.10`. This resolves that issue.


leseb:
I don't think so given that this is happening every day.

PR #1274, opened 2018-12-06
daemon-base: add gdisk to ceph base packages

BlaineEXE:
Rook expects sgdisk to be in images alongside Ceph. Add this to the base
packages list.

**Note:** The ideal fix for the situation is to allow Rook to use its own sgdisk, but this is the quickest option to get it fixed in Rook v0.9

### DIFF
--- a/src/daemon-base/__CEPH_BASE_PACKAGES__
+++ b/src/daemon-base/__CEPH_BASE_PACKAGES__
@@ -9,6 +9,7 @@
         __CEPH_MGR_PACKAGE__\
         kmod \
         lvm2 \
+        sgdisk \
         __RADOSGW_PACKAGE__ \
         __GANESHA_PACKAGES__ \
         __ISCSI_PACKAGES__


leseb:
@BlaineEXE why doesn't Rook get `sgdisk` as part of its current container image? Basically, on top of this one.

ErwanAliasr1:
Dependencies must come from the packages and never directly in the Dockerfile. Btw, we have some cleaning to do around that like all the lvm2, e2fs packages.

ErwanAliasr1:
Isn't the solution to use ceph-volume instead of the deprecated ceph-disk ?

Le ven. 7 déc. 2018 à 18:24, Travis Nielsen <notifications@github.com> a
écrit :

travisn:
Rook actually never used ceph-disk. Instead room implemented partitioning and everything at a low level based on sgdisk and lsblk.

travisn:
To be more clear, Rook now is calling ceph-volume. However, we still need sgdisk to support upgraded clusters unless we can find a workaround.

PR #1286, opened 2019-01-08
Reenable ganesha builds

jtlayton:
Not sure why these got disabled in the first place, but let's bring them back in now. A local  FLAVORS="master,centos,7" build worked fine for me.


ErwanAliasr1:
The reason was about because the ceph.com packages where no more available.

jtlayton:
Yes, @ktdreyer says "it is probably because the nfs-ganesha builds were deleted, because they were (accidentally) deleted from ceph.com as well". They are back now though, so we should be able to reenable them.

jtlayton:
Retest this please

PR #1298, opened 2019-01-18
Allow the monitor interface to be provided

mterzo:
Don't use the detected interface if CEPH_ETH is defined.  Detected based
on the amount of traffic over the interface is not always the
wanted behaviour.

### DIFF
--- a/src/daemon/start_mon.sh
+++ b/src/daemon/start_mon.sh
@@ -92,7 +92,9 @@ function start_mon {
       fi
   else
     local nic_more_traffic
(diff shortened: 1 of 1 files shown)


mterzo:
I couldn't find any current documentation about various ENV values and how they are used.

mterzo:
That's what You needed?

leseb:
@mterzo almost, the sing-off typically comes after the commit message. Please put it at the end. Thanks.

leseb:
@mterzo thanks!

PR #1309, opened 2019-02-15
add rhel8 support

leseb:


ceph-releases/ALL/rhel8/daemon-base/__DOCKERFILE_PREINSTALL__
```diff
@@ -0,0 +1,20 @@
...
+EXPOSE 6789 6800 6801 6802 6803 6804 6805 80 5000
+
+# Atomic specific labels
+ADD install.sh /install.sh
+LABEL version=4
+LABEL run="/usr/bin/docker run -d --net=host -e MON_NAME=\${MON_NAME} -e MON_IP=\${MON_IP} -e CEPH_PUBLIC_NETWORK=\${CEPH_PUBLIC_NETWORK} -e CEPH_DAEMON=\${CEPH_DAEMON} -v /etc/ceph:/etc/ceph -v /var/lib/ceph:/var/lib/ceph \${IMAGE}"
+LABEL install="/usr/bin/docker run --rm --privileged -v /:/host -e MON_IP=\${MON_IP} -e CEPH_PUBLIC_NETWORK=\${CEPH_PUBLIC_NETWORK} -e CEPH_DAEMON=\${CEPH_DAEMON} -e MON_NAME=\${MON_NAME} -e OSD_DEVICE=\${OSD_DEVICE} -e HOST=/host -e IMAGE=\${IMAGE} --entrypoint=/install.sh \${IMAGE}"
+
+# Build specific labels
+LABEL com.redhat.component="rhceph-rhel7-container"
```
ktdreyer: Would you please change this from rhel7 to rhel8?

ceph-releases/ALL/rhel8/daemon/content_sets.yml
```diff
@@ -0,0 +1,25 @@
...
+- rhel-8-server-extras-rpms
+- rhel-8-server-rhceph-4-tools-rpms
+- rhel-8-server-rhceph-4-mon-rpms
+- rhel-8-server-rhceph-4-osd-rpms
+ppc64le:
+- rhel-8-for-power-le-rpms
+- rhel-8-for-power-le-extras-rpms
+- rhel-8-server-rhceph-4-for-power-le-tools-rpms
+- rhel-8-server-rhceph-4-for-power-le-mon-rpms
+- rhel-8-server-rhceph-4-for-power-le-osd-rpms
```
ktdreyer: From what I can tell of RCM's RHEL 8 Pulp plans, here is what we'll need in this file:

```yaml
---
x86_64:
- rhel-8-x86_64-rpms
- rhel-8-x86_64-extras-rpms
- rhceph-4-tools-for-rhel-8-x86_64-rpms
- rhceph-4-mon-for-rhel-8-x86_64-rpms
- rhceph-4-osd-for-rhel-8-x86_64-rpms
ppc64le:
- rhel-8-ppc64le-rpms
- rhel-8-ppc64le-extras-rpms
- rhceph-4-tools-for-rhel-8-ppc64le-rpms
- rhceph-4-mon-for-rhel-8-ppc64le-rpms
- rhceph-4-osd-for-rhel-8-ppc64le-rpms
```

if it turns out we need more changes here, we can include them later.

leseb:
@ktdreyer is this tracker by any BZ? I'd like to add that to the commit. Thanks.

ktdreyer:
Sure, you can add https://bugzilla.redhat.com/show_bug.cgi?id=1656129 as a reference.

leseb:
@ktdreyer we should be good now, thanks for the review.

ceph-releases/ALL/rhel8/daemon-base/__DOCKERFILE_INSTALL__
```diff
@@ -0,0 +1,3 @@
+yum update -y && \
+yum install -y wget unzip util-linux python-setuptools udev device-mapper && \
```
ktdreyer: I'm testing this more. `python-setuptools` is not in RHEL 8, and it's `python3-setuptools`.

ceph-releases/ALL/rhel8/daemon-base/__ISCSI_PACKAGES__
```diff
@@ -0,0 +1 @@
+tcmu-runner ceph-iscsi-config ceph-iscsi-cli
```
dillaman: ```ceph-iscsi``` for RHEL8 -- does this mean we need to create a builder for RHEL8 rpms?

ktdreyer: :+1: we will have `ceph-iscsi` packages here. (upstream builds TBD)

ceph-releases/ALL/rhel8/daemon-base/__DOCKERFILE_POSTINSTALL_CLEANUP__
```diff
@@ -0,0 +1,5 @@
+echo 'Postinstall cleanup' && \
+ ( yum clean all && \
+   rpmdb --rebuilddb && \
```
ktdreyer: This `rpmdb --rebuilddb` command fails with the following error:

```
error: failed to replace old database with new database!
error: replace files in /var/lib/rpm with files from /var/lib/rpmrebuilddb.1096 to recover
```

dsavineau: Same on fedora 29. I'm wondering if this could come from the overlayfs issue with rename command on that FS.
I think that was resolved on EL7 by using the yum ovl plugin but I might be wrong (I tried dnf-plugin-ovl with no difference).

ktdreyer: I think we should take this line out for RHEL 8. I've filed https://bugzilla.redhat.com/show_bug.cgi?id=1680124 to track this with the RPM team.

ceph-releases/ALL/rhel8/daemon-base/__ISCSI_PACKAGES__
ktdreyer: Would you please change `ceph-iscsi-config ceph-iscsi-cli` to `ceph-iscsi` ?

leseb: Done.

leseb:
@ktdreyer are you going to do the downstream sync too? Thanks

ktdreyer:
We don't have the downstream pieces all set up yet, so I have Jenkins building this from GitHub outside the usual process for now.

When we get closer to shipping, we'll need to sync with downstream dist-git. But for today it's a lighter weight process.

I'd like to work with you to get something better in place than we have today. For example, I have to run this in Jenkins in order to build this version:
```
sed -i 's/7/8/g' contrib/compose-rhcs.sh
sed -i 's/luminous/nautilus/g' contrib/compose-rhcs.sh
```

How would you like to make that more robust?

leseb:
@ktdreyer can you send a PR to modify `
contrib/compose-rhcs.sh` so it can read an `RHEL_VERSION` env var or pass an arg?

That doesn't seem to be too difficult to fix, I'll leave that up to you. Thanks.

PR #1310, opened 2019-02-18
Add ceph-mgr-rook to nautilus images

jtlayton:
We want ceph-mgr-rook to be installed on nautilus and later images, but
earlier releases didn't have that package. Add it to the base
`__CEPH_MGR_PACKAGE__` file and add override files for luminous and mimic.


leseb:
@jtlayton building here: https://2.jenkins.ceph.com/job/ceph-container-build-push-imgs/326/

jtlayton:
Thanks! We'll probably need to do a rebuild once a fixed python-kubernetes module is available (in a couple of days).

PR #1311, opened 2019-02-19
Add dashboard mgr module by default

jtlayton:
rook expects the dashboard to be available.


leseb:
@jtlayton is it true for all the Ceph releases? Or only Nautilus?

jtlayton:
Nautilus and up. The splitout of mgr modules was just done recently in nautilus.

leseb:
So basically if we put this in `src/daemon-base/` this will affect all the daemons, that's a problem.

jtlayton:
We already have override files for luminous and mimic:

ceph-releases/luminous/daemon-base/__CEPH_MGR_PACKAGE__
ceph-releases/mimic/daemon-base/__CEPH_MGR_PACKAGE__

Won't they supersede this file on those releases?

leseb:
@jtlayton you're right, I just forgot about that!

Issue #1312, opened 2019-02-21
New flavor for Fedora 29 container


leseb:
Ability to build a f29 container.

leseb:
@dsavineau This one is for you when you're ready! :-)

dsavineau:
@leseb I'll take a look !

PR #1314, opened 2019-02-22
Add fedora 29 support

dsavineau:
Fedora 29 only supports luminous release.

Fixes: #1312

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,7 @@ ALL_BUILDABLE_FLAVORS := \
 	luminous,centos,7 \
 	luminous,opensuse,42.3 \
 	luminous,debian,9 \
+	luminous,fedora,29 \
 	mimic,centos,7 \
 	master,centos,7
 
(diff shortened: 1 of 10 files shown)


dsavineau:
AFAIK the upstream ceph repositories don't provide rpm packages for fedora (only EL6/7) [1][2].
Luminous packages are present natively in f29 distro so that's why I specify only that release.

[1] https://download.ceph.com/rpm-luminous/
[2] https://download.ceph.com/rpm-mimic/

leseb:
@dsavineau perhaps el7 packages will work on f29?

dsavineau:
@leseb I'll give it a try. My only concern is python 2 vs 3 stuff

ktdreyer:
There are too many changes between Fedora 29 and RHEL 7.

leseb:
@ktdreyer thanks Ken. Let's abandon this then, there is no real benefit of a Luminous container.

PR #1325, opened 2019-03-16
daemon/osd: Migrate ceph-disk to ceph-volume

dsavineau:
In Nautilus release, the ceph-disk command isn't supported anymore.
Instead we should migrate to ceph-volume.
To keep the backward compatibility with old container images using
the osd_ceph_disk_activate entrypoint, we need to replace the
osd_activate function with ceph-volume simple scan/activate commands
otherwise the container won't be able to start after upgrading from
Luminous or Mimic container image.

The new version of osd_disk_activate.sh is only available for Nautilus
and above releases. Older releases (Luminous and Mimic) keep the
previous one.

http://docs.ceph.com/docs/master/ceph-volume/#migrating

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1656506

### DIFF
--- a/ceph-releases/luminous/daemon/osd_scenarios/osd_disk_activate.sh
+++ b/ceph-releases/luminous/daemon/osd_scenarios/osd_disk_activate.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+# shellcheck disable=SC2034
+set -e
+source /opt/ceph-container/bin/disk_list.sh
+
+function osd_activate {
+  if [[ -z "${OSD_DEVICE}" ]] || [[ ! -b "${OSD_DEVICE}" ]]; then
+    log "ERROR: you either provided a non-existing device or no device at all."
+    log "You must provide a device to build your OSD ie: /dev/sdb"
+    exit 1
+  fi
+
+  CEPH_DISK_OPTIONS=()
+
+  if [[ ${OSD_FILESTORE} -eq 1 ]] && [[ ${OSD_DMCRYPT} -eq 0 ]]; then
+    if [[ -n "${OSD_JOURNAL}" ]]; then
+      CLI+=("${OSD_JOURNAL}")
+    else
+      CLI+=("${OSD_DEVICE}")
+    fi
+    export DISK_LIST_SEARCH=journal
+    start_disk_list
+    JOURNAL_PART=$(start_disk_list)
+    unset DISK_LIST_SEARCH
+    JOURNAL_UUID=$(get_part_uuid "${JOURNAL_PART}")
+  fi
+
+  # creates /dev/mapper/<uuid> for dmcrypt
+  # usually after a reboot they don't go created
+  udevadm trigger
+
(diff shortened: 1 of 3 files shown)


leseb:
It'd be nice to attach a BZ to that.

randomnord:
Hi,

Any update on when this fix is going to be released? i can see that the travis build failed

leseb:
@randomnord no worries, the failure is irrelevant.

randomnord:
@leseb any idea when this change might get pushed to DockerHub?

dsavineau:
@randomnord This should be done during this week

guits:
@leseb shouldn't we make this change available only for nautilus and after ?

leseb:
@dsavineau are you going to do the downstream sync too?

dsavineau:
@leseb I'm just wondering if we need to create the stable-4.0 branch before the sync downstream. wdyt ?

leseb:
@dsavineau I think it makes to branch stable-4.0, yes. Please go ahead :).

Tydus:
Thanks for the PR. However I think `osd_disk_prepare.sh` should also be migrated. I am blocked at this step without manually prepare the disks outside the containers.

dsavineau:
@Tydus I don't think `osd_disk_prepare.sh` needs to be migrated. In nautilus, ceph-disk isn't available anymore so you should deploy new OSDs with ceph-volume.
The update in `osd_disk_activate.sh` is only for the upgrade path from luminous/mimic to nautilus with OSDs already deployed with ceph-disk.

osegarra:
Hi, 

I'm installed the latest version available but I'm facing the same problem:

Error: /dev/sdb: unrecognised disk label,
/opt/ceph-container/bin/osd_disk_prepare.sh: line 46: ceph-disk: command not found

Is there any workarround for this issue?

Thanks a lot.

guits:
@osegarra please, open an issue and share more details.

osegarra:
https://tracker.ceph.com/issues/39976

dsavineau:
@osegarra this PR was only about to be able to start an OSD container that was deployed with ceph-disk in luminous/mimic via ceph-volume simple scan (osd_disk_activate.sh).
The full ceph-disk removal will be done in https://github.com/ceph/ceph-container/pull/1349

osegarra:
@dsavineau, I don't know if I understand correctly what you are trying to say but the final question is that from the image provided as a result of the make process of the nautilus flavour fails when deploying an OSD.

We will see the evolution of the opened issue.

Any comment or clarifications will be welcome,

dsavineau:
@osegarra the current `osd` entrypoint is still using the ceph-disk command that's why the container is failing to start. The purpose of #1349 is to remove this (or replace some part) by ceph-volume only.
Currently you need to use the `osd_ceph_volume_activate` entrypoint [1] to be able to start an OSD container. This requires to create the OSD with ceph-volume before the container.

[1] https://github.com/ceph/ceph-container/blob/master/src/daemon/entrypoint.sh.in#L102-L108

PR #1327, opened 2019-03-19
Introduce new environment variable to make use of public bind address.

ne0h:
Introducing the newly-added parameter '--public-bind-address' as an
envorinment variable to allow exposing ceph monitors via ClusterIP
in Kubernetes.

Checklist:


src/daemon/variables_entrypoint.sh
```diff
@@ -67,6 +67,7 @@ fi
 : "${MGR_IP:=0.0.0.0}"
 : "${CEPH_ARCH:=$(uname -m)}"
 : "${MON_PORT:=6789}"
+: "${PUBLIC_BIND_ADDRESS:=${PUBLIC_BIND_ADDRESS}}"
```
dsavineau: This should not be needed since you check the variable in start_mon.sh

src/daemon/start_mon.sh
```diff
@@ -192,7 +192,11 @@ function start_mon {
     # DO NOT TOUCH IT, IT MUST BE PRESENT
     DAEMON_OPTS+=(--mon-cluster-log-to-stderr "--log-stderr-prefix=debug ")
     log "SUCCESS"
-    exec /usr/bin/ceph-mon "${DAEMON_OPTS[@]}" -i "${MON_NAME}" --mon-data "$MON_DATA_DIR" --public-addr "${MON_IP}"
+    # Set PUBLIC_BIND_ADDRESS to MON_IP if unset
+    if [ -z "${PUBLIC_BIND_ADDRESS}" ]; then
+        PUBLIC_BIND_ADDRESS=${MON_IP}
+    fi
+    exec /usr/bin/ceph-mon "${DAEMON_OPTS[@]}" -i "${MON_NAME}" --mon-data "$MON_DATA_DIR" --public-addr "${PUBLIC_BIND_ADDRESS}" --public-bind-addr "${MON_IP}"
```
dsavineau: This is a little bit confusing because you're using the PUBLIC_BIND_ADDRESS variable for --public-addr option.
It would be better to rename the variable to something like MON_PUBLIC_ADDRESS and keep MON_IP for --public-bind-addr.
What do you think ?

src/daemon/variables_entrypoint.sh
ne0h: This has only been added to keep the "LIST OF ALL VARIABLES" up-to-date. Can easily be removed if desired.

src/daemon/start_mon.sh
ne0h: You are right, this is a little bit confusing. If the "--public-bind-address" is used, the monitor tries to bind the network interface which is provided by this parameter. The address passed by "--public-addr" is used for external authentication (https://github.com/ceph/ceph-client/blob/master/net/ceph/messenger.c#L2020).

ne0h:
push

ne0h:
Since the issue is not fixed, this PR should be re-opened!

PR #1328, opened 2019-03-19
daemon: Add support for python3

dsavineau:
Currently some entrypoint scripts were using python command. This
will in most of case refers to python 2 and not python 3.
Instead of hardcoding the python binary in the shell scripts we can
test if the binary exist and set the PYTHON variable.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1690093

### DIFF
--- a/src/daemon/variables_entrypoint.sh
+++ b/src/daemon/variables_entrypoint.sh
@@ -102,6 +102,12 @@ if [[ "$KV_TYPE" == "etcd" ]]; then
   ETCDCTL_OPTS=(--peers ${ETCD_SCHEMA}${KV_IP}:${KV_PORT})
 fi
 
+if is_available python; then
+  PYTHON=python
+else
+  PYTHON=python3
+fi
+
 # Internal variables
 MDS_KEYRING=/var/lib/ceph/mds/${CLUSTER}-${MDS_NAME}/keyring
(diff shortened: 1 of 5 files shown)


src/daemon/variables_entrypoint.sh
```diff
@@ -102,6 +102,12 @@ if [[ "$KV_TYPE" == "etcd" ]]; then
   ETCDCTL_OPTS=(--peers ${ETCD_SCHEMA}${KV_IP}:${KV_PORT})
 fi
 
+if is_available python; then
+  PYTHON=python
+else
+  PYTHON=python3
```
leseb: Why are we assuming that if `python` is not present then `python3` is?
Shouldn't we do something like this perhaps?

```
PYTHON=python
if is_available python3; then
  PYTHON=python3
fi
```

Or check for the existence of each and then fail if not present?

dsavineau: > Why are we assuming that if python is not present then python3 is?

because on distribution (like rhel 8, fedora 29) where python3 is the default python interpreter there's no python command (or symlink).

> Or check for the existence of each and then fail if not present?

We could but it's probably safe to say that base distro container images have at least one python interpreter

leseb: Looks acceptable.

PR #1329, opened 2019-03-20
Add environment variable to provide network fallback address.

ne0h:
Allow passing a fallback network address to monitor startup. This
address will be used if auto-detecting the network fails.

Solely relying on the configuration of the routing can lead to startup errors of monitors in some Kubernetes environments.
If so, the network of the pod is not detected properly which breaks the startup of the whole ceph cluster.

Checklist:


ne0h:
push

ne0h:
Since the issue is not fixed, this PR should be re-opened!

PR #1332, opened 2019-03-21
daemon/variables_entrypoint Fix python detection

dsavineau:
778c7f9 introduces the python binary detection based on the
is_available function. Unfortunately this function is source after
variables_entrypoint.sh in the entrypoint so it's not possible to
use it.
Also Luminous and Mimic release have a different variables entrypoint
script they're modified to reflect the python change.


guits:
seems like only @leseb can make the CI passing since it's linked to his docker hub account

PR #1337, opened 2019-03-28
base: add packages for ceph-csi

leseb:
We want to use our base image for ceph-csi. ceph-csi needs extra
packages so we are adding them.

### DIFF
--- a/src/daemon-base/__CEPH_BASE_PACKAGES__
+++ b/src/daemon-base/__CEPH_BASE_PACKAGES__
@@ -12,4 +12,5 @@
         gdisk \
         __RADOSGW_PACKAGE__ \
         __GANESHA_PACKAGES__ \
-        __ISCSI_PACKAGES__
+        __ISCSI_PACKAGES__ \
+        __CSI_PACKAGES__
--- a/src/daemon-base/__CSI_PACKAGES__
+++ b/src/daemon-base/__CSI_PACKAGES__
@@ -1,0 +1,1 @@
+ceph-fuse attr rbd-nbd


src/daemon-base/__CEPH_BASE_PACKAGES__
```diff
@@ -12,4 +12,7 @@
         gdisk \
         __RADOSGW_PACKAGE__ \
         __GANESHA_PACKAGES__ \
-        __ISCSI_PACKAGES__
+        __ISCSI_PACKAGES__ \
+        ceph-fuse \
```
dsavineau: I'm wondering if it makes sense to put those packages under __CSI_PACKAGES__ (to be created) wdyt ?
Also I'm not sure those packages are available for all distro/release, did you checked that ?

leseb: Yes, I thought about `CSI_PACKAGES`, I guess it makes sense to have it if packages are named differently in other distros, I haven't checked that yet.

dsavineau: I've checked and the three packages are available for all distro/releases
I've updated the PR because there was a typo on `__CSI_PACKAGES__` (missing S) and also remove the other empty occurence in ceph-release directory

ktdreyer:
Why doesn't the ceph-csi Dockerfile include these?

leseb:
@ktdreyer the goal is for ceph-csi to use that our ceph base image so that they don't have to care about the underlying Ceph image. They simply build their binary and put it on top of our image.

leseb:
@dsavineau good to go?

ktdreyer:
They can run `yum install`, right?

leseb:
@ktdreyer yes, but I don't think we want that either.

PR #1340, opened 2019-03-28
travis: Remove civetweb references for demo

dsavineau:
We now are using beast as radosgw frontend so it doesn't make sense
to keep the civetweb environment variables.


leseb:
@dsavineau please rebase.

PR #1341, opened 2019-03-29
base: fix lang.sh removal on RHEL

ktdreyer:
Commit 15d8573c1d9dfedbda9ca012f3609497704e34af introduced a syntax error because we omitted the `&&`, so `sed` was trying to interpret the `rm`, `-f` etc. as additional arguments.

Add the `&&` so the shell interprets the `sed` and `rm` as two separate commands.


dsavineau:
@ktdreyer thanks for spotting the issue.

PR #1350, opened 2019-04-15
daemon-base: Add ceph-mgr-diskprediction-local pkg

dsavineau:
Add diskprediction_local mgr module in order to have the same default
modules present in ceph-ansible and ceph-container.


dsavineau:
Nope Nautilus+ only, but we already take care about mgr modules (like rook) for pre and post Nautilus release

https://github.com/ceph/ceph-container/blob/master/ceph-releases/luminous/daemon-base/__CEPH_MGR_PACKAGE__
https://github.com/ceph/ceph-container/blob/master/ceph-releases/mimic/daemon-base/__CEPH_MGR_PACKAGE

leseb:
Just ran the CI again.

PR #1363, opened 2019-05-01
rhel8: drop "-rhel8" from component label

ktdreyer:
For the latest RH Ceph Storage on RHEL 8, we're going to simplify the component name to avoid the "rhel" string and just go with "rhceph". This matches what other products are doing for el8.


dsavineau:
@ktdreyer could you amend you commit and add a signoff ?

otherwise the CI error on mimic build is unrelated, I'm working on that.

ktdreyer:
Sure, I've added my signed-off-by line.

Issue #1343, opened 2019-04-01
Unable to determine osd id when dmcrypt is enabled


michalskalski:
Is this a bug report or feature request?
* Bug Report

**Bug Report**

What happened:

Im deploying ceph with rhosp 13 and dmcrypt enabled together with filestore. Ceph container is restarting constantly on execution of this command https://github.com/ceph/ceph-container/blob/stable-3.2/src/daemon/osd_scenarios/osd_disk_activate.sh#L84 because OSD_ID is empty. It looks like the reason of this is that in these 2 lines https://github.com/ceph/ceph-container/blob/stable-3.2/src/daemon/osd_scenarios/osd_disk_activate.sh#L58-L59 /proc/mounts is searched for the device to which /dev/mapper/x symlink is pointing to, but in my environment the /proc/mounts has the /dev/mapper/x device listed instead of /dev/dm-x:

```
[root@overcloud8yi-cephstoragehw0-0 /]# ls -l /dev/mapper/
total 0
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 24006e80-8d21-44c4-9cf2-a1c5cf012e1b -> ../dm-7
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 2ee04908-9a17-40a4-945d-a57a89ac93d6 -> ../dm-6
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 42a02812-ec16-4da8-966b-aa539142229e -> ../dm-1
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 52018de6-0f71-4947-be3a-fad4b5760c46 -> ../dm-2
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 6e239bef-d614-47bf-9c2d-ca7e05bb1416 -> ../dm-3
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 9e161dc5-c4f3-4f85-a1af-27f3d6008b13 -> ../dm-4
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 c193d6af-0959-47c5-b75e-96958834200b -> ../dm-0
crw-------. 1 root root 10, 236 Apr  1 21:59 control
lrwxrwxrwx. 1 root root       7 Apr  1 21:59 f95acaa3-749d-424c-b3c9-21ae1f0e37c4 -> ../dm-5
[root@overcloud8yi-cephstoragehw0-0 /]# cat /proc/mounts | grep osd
[...]

PR #1380, opened 2019-05-15
With dmcrypt expect /dev/mapper/device to be used

dsavineau:
When try to determine osd id with dmcrypt enabled expect that
/dev/mapper/device will be visible in /proc/mounts table.

Closes: #1343
Signed-off-by: Michal Skalski <mskalski@juniper.net>


dsavineau:
This change was proposed and merged in stable-3.2 branch but we're still using the master branch to build latest-* container images so the latest-luminous tag doesn't have this fix.

PR #1381, opened 2019-05-15
contrib/build-push-ceph: Add arm64 for nautilus

dsavineau:
12ab4d2 adds arm64 to ceph-build-config script but there's some
missing pieces in order to build arm64 nautilus image.
This also adds nautilus to the registry manifest creation because
we now have multiple arch for that release too (the same way than
luminous and mimic)

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -201,7 +201,7 @@ function create_registry_manifest {
   # This should normally work, by the time we get here the arm64 image should have been built and pushed
   # IIRC docker manisfest will fail if the image does not exist
   for image in daemon-base daemon; do
-    for ceph_release in luminous mimic; do
+    for ceph_release in ${CEPH_RELEASES[@]:1}; do
       TARGET_RELEASE="ceph/${image}:${RELEASE}-${ceph_release}-centos-7"
       DOCKER_IMAGES="$TARGET_RELEASE ${TARGET_RELEASE}-x86_64"
 
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ ALL_BUILDABLE_FLAVORS := \
 	luminous,opensuse,42.3 \
 	luminous,debian,9 \
 	mimic,centos,7 \
+	nautilus,centos,7 \
 	master,centos,7
 
 # ==============================================================================
--- a/contrib/build-push-ceph-container-imgs-arm64.sh
+++ b/contrib/build-push-ceph-container-imgs-arm64.sh
@@ -5,7 +5,7 @@ set -ex
 # VARIABLES #
 #############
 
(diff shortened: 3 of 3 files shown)


contrib/build-push-ceph-container-imgs.sh
```diff
@@ -201,7 +201,7 @@ function create_registry_manifest {
   # This should normally work, by the time we get here the arm64 image should have been built and pushed
   # IIRC docker manisfest will fail if the image does not exist
   for image in daemon-base daemon; do
-    for ceph_release in luminous mimic; do
+    for ceph_release in luminous mimic nautilus; do
```
leseb: Maybe we can use https://github.com/ceph/ceph-container/pull/1381/files#diff-9bd28db8c71e92c690fefd2d189c6b46R15 and remove the first element of the list?

dsavineau: This will only work if CEPH_RELEASES is configured with the default value. https://github.com/ceph/ceph-container/blob/master/contrib/build-push-ceph-container-imgs.sh#L13-L16

If someone set CEPH_RELEASES with only one value then this solution won't work (but I don't know if it's a valid scenario)

leseb: It's not intended to change `CEPH_RELEASES` so I think we are good.

dsavineau: ack

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -201,7 +201,7 @@ function create_registry_manifest {
   # This should normally work, by the time we get here the arm64 image should have been built and pushed
   # IIRC docker manisfest will fail if the image does not exist
   for image in daemon-base daemon; do
-    for ceph_release in luminous mimic; do
+    for ceph_release in ${CEPH_RELEASES[@]:1}; do
```
leseb: I'm doubting now, if we call `contrib/build-push-ceph-container-imgs-arm64.sh ` that's going to override`CEPH_RELEASES` with the one in `contrib/build-push-ceph-container-imgs-arm64.sh` so removing the first element will remove `luminous` in this case right?

dsavineau: I'll take a closer look but I think you're right so we probably need to set the list explicitly and not reuse the CEPH_RELEASES variable

dsavineau: in fact this has no side effect for arm64 because the `create_registry_manifest` function is only executed from `contrib/build-push-ceph-container-imgs.sh` (so with the right list).
With the arm64 script the function does nothing. https://github.com/ceph/ceph-container/blob/master/contrib/build-push-ceph-container-imgs-arm64.sh#L45-L47

PR #1410, opened 2019-06-27
daemon/start_mon.sh: add mon without port value

dsavineau:
When bootstrapping an initial cluster, we don't need to add the mon
to the monmap via 'ceph mon add' command.
But when we want to scale up the monitors then we need it.
Currently the new mon added will be out of quorum because the addr
in the ceph config and in the monmap won't match.

start_mon(): ceph --cluster ceph mon add xxx 192.168.100.5:3300
mon.xxx at v2:192.168.100.5:3300/0 already exists

Even if the ceph conf uses v2+v1 then the new mon will have only v2
address configured in the monmap creating a conflict.

0 mon.xxx@0(probing) e3  monmap addrs for rank 4 changed, i am
  [v2:192.168.100.5:3300/0,v1:192.168.100.5:6789/0], monmap is
  v2:192.168.100.5:3300/0, respawning
0 mon.xxx@0(probing) e3 respawn

Removing the port value on the 'ceph mon add' command allows to add
a monitor with the msgr configuration defined in the ceph config.

This change is nautilus+ only.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1719013
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1722066

### DIFF
--- a/src/daemon/start_mon.sh
+++ b/src/daemon/start_mon.sh
@@ -170,7 +170,14 @@ function start_mon {
       ceph-mon --setuser ceph --setgroup ceph --cluster "${CLUSTER}" -i "${MON_NAME}" --inject-monmap "$MONMAP" --keyring "$MON_KEYRING" --mon-data "$MON_DATA_DIR"
     fi
     if [[ "$CEPH_DAEMON" != demo ]]; then
-      timeout 7 ceph "${CLI_OPTS[@]}" mon add "${MON_NAME}" "${MON_IP}":"${MON_PORT}" || true
+      v2v1=$(ceph-conf -c /etc/ceph/${CLUSTER}.conf 'mon host' | tr ',' '\n' | grep -c ${MON_IP})
+      # in case of v2+v1 configuration : [v2:xxxx:3300,v1:xxxx:6789]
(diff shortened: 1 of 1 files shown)


guits:
@leseb couldn't it be a breaking change for existing deployment where `$MON_PORT` would have been previously overridden ?

leseb:
@guits do we allow that in ceph-ansible?

guits:
@leseb yes, I guess using `ceph_mon_docker_extra_env: -e MON_PORT=1234` would allow it.

leseb:
Even though we specify a different port, I don't think this will actually work, (the mon map does not have that port). If this is supposed to work in ceph-ansible then the mon endpoint should have that vars, I don't think it does.

fultonj:
@dsavineau @guits 
The scratch container with this change [1] resolved our issue of scaling from 3 to 5 mons. 

[1] docker.io/ceph/daemon:bz1719013-nautilus-centos-7-x86_64

yorabl:
@fultonj is right, the scale up of the monitors was successful, but the scale up of rgw services failed

dsavineau:
@fultonj @yorabl thanks for the update !

Issue #1413, opened 2019-07-03
remove vim from ceph-container


ktdreyer:
**Feature Request**

_What should the feature do:_
Please consider removing vim from the ceph container image.

_What would be solved through this feature:_
More packages = More security risks. In particular, Vim in RHEL 7 was recently vulnerable to [CVE-2019-12735](https://access.redhat.com/errata/RHSA-2019:1619). This shows up as a bad mark on security scanner reports for our container image, and requires us to rebuild the container for this package's security updates.

dsavineau:
> More packages = More security risks.

vim-minimal is part of the base image (Centos 7 or RHEL 7) and not added by ceph-container so you will have the same issue with any security updates on based packages.

ktdreyer:
I can run `rpm -e vim-minimal` in the `ubi8:latest` container, so we should be able to uninstall it during the cleanup part of our ceph container build process.

dsavineau:
I know we can remove it and that was not my point.
Why should we only do this for vim ? What about other based packages not useful that can be impacted by security updates too ?
If it's only about a bad mark on security scanner reports then RHEL 7 is impacted too. So if the base image changes then we should rebuild the Ceph image too right ? 
I thought this process was automatically done for RHCS but I might be wrong.
If we rebuild the Ceph image when the base image change then the vim removal doesn't make sense.

ktdreyer:
> If it's only about a bad mark on security scanner reports then RHEL 7 is impacted too. So if the base image changes then we should rebuild the Ceph image too right ?

The rebuild process for RHCS uses a system called [Freshmaker](https://pagure.io/freshmaker/). This system queries another datastore called [Lightblue](https://pagure.io/freshmaker/blob/master/f/freshmaker/lightblue.py). LightBlue has information about all the RPMs in all our container images. This allows Freshmaker to know which images to rebuild. In other words, if Vim is not in present in the RHCS image, Freshmaker will not rebuild RHCS for a Vim CVE.

[...]

PR #1420, opened 2019-07-05
centos/rhel: remove vim-minimal package

dsavineau:
The vim package isn't required on the ceph container image and could
shows a bad mark on security scanner reports when CVE are discovered.

Resolves: #1413


guits:
-1  see https://github.com/ceph/ceph-container/issues/1413#issuecomment-523941029

dsavineau:
Closing this. It's not the right way of managing update/security issue in the container image.
This should be done by updating the base image with the update/security fixes or using a custom base image without the package.

ktdreyer:
I haven't made it clear how much work it for the product team to rebuild and re-test for these ancillary security problems. I guess we don't worry about it upstream.

mykaul:
> Closing this. It's not the right way of managing update/security issue in the container image.
> This should be done by updating the base image with the update/security fixes or using a custom base image without the package.

Both options make sense to me - especially the latter. Do we have an issue tracking the move to RHEL UBI8 minimal, for example?

PR #1425, opened 2019-07-16
zap_device: zap dmcrypt partitions as well

guits:
if the device passed is a partition, we never call
`zap_dmcrypt_device()`, it means dmcrypt partitions aren't wipped properly.
This commit moves the call to `zap_dmcrypt_device()` so it applies for
root devices and partitions.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1572933

### DIFF
--- a/src/daemon/zap_device.sh
+++ b/src/daemon/zap_device.sh
@@ -110,22 +110,22 @@ function zap_device {
       exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


guits:
as usual, nfs-ganesha repo in chacra seems to be broken :expressionless:

PR #1428, opened 2019-07-16
zap_device: zap dmcrypt partitions as well (bp #1425)

dsavineau:
if the device passed is a partition, we never call
`zap_dmcrypt_device()`, it means dmcrypt partitions aren't wipped properly.
This commit moves the call to `zap_dmcrypt_device()` so it applies for
root devices and partitions.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1572933
Backport: #1425

(cherry picked from commit f285b23b75f749edd2bab9a8796fe9b90d5e80b6)

### DIFF
--- a/src/daemon/zap_device.sh
+++ b/src/daemon/zap_device.sh
@@ -110,22 +110,22 @@ function zap_device {
       exit 1
     fi
 
+    # if the disk passed is a raw device AND the boot system disk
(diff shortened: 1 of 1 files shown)


guits:
@dsavineau I know this is definitely not elegant, but any chance we could merge this even if `master` and `stable-4.0` can't be merged so we can proceed with the downstream backport in `ceph-3.3-rhel-7`?

dsavineau:
FYI this won't be part of latest-luminous tag

PR #1436, opened 2019-07-23
ceph: unblock master images

leseb:
Let's also pin master to ganesha 2.8 since ganesha master is still
broken.


guits:
Although I'm unsure nfs-ganesha 2.8 is supposed to work with ceph@master, I'm merging this so we can build images.

guits:
@dsavineau  ^

dsavineau:
nfs-ganesha 2.8 couldn't be build against ceph@master due to change in the rgw lib.
Even if we're able to install nfs-ganesha 2.8 (nautilus) on ceph@master image, I don't think everything will wok on the ganesha side.

guits:
> nfs-ganesha 2.8 couldn't be build against ceph@master due to change in the rgw lib.
> Even if we're able to install nfs-ganesha 2.8 (nautilus) on ceph@master image, I don't think everything will wok on the ganesha side.

this is well what I was afraid of

dsavineau:
> this is well what I was afraid of

:+1: maggle

PR #1442, opened 2019-07-30
daemon/start_mon.sh: fix failure handling in ceph-conf 'mon host'

CoRfr:
In commit 'daemon/start_mon.sh: add mon without port value' the
following code was added:
  v2v1=$(ceph-conf -c /etc/ceph/${CLUSTER}.conf 'mon host' | tr ',' '\n' | grep -c ${MON_IP})

If that command fails, the process stops because of set -e, while it is
expected that we would reach the 'else' case.

Description of your changes:

Resolves #

Checklist:

### DIFF
--- a/src/daemon/start_mon.sh
+++ b/src/daemon/start_mon.sh
@@ -170,13 +170,14 @@ function start_mon {
       ceph-mon --setuser ceph --setgroup ceph --cluster "${CLUSTER}" -i "${MON_NAME}" --inject-monmap "$MONMAP" --keyring "$MON_KEYRING" --mon-data "$MON_DATA_DIR"
     fi
     if [[ "$CEPH_DAEMON" != demo ]]; then
-      v2v1=$(ceph-conf -c /etc/ceph/${CLUSTER}.conf 'mon host' | tr ',' '\n' | grep -c ${MON_IP})
-      # in case of v2+v1 configuration : [v2:xxxx:3300,v1:xxxx:6789]
-      if [ ${v2v1} -eq 2 ]; then
-        timeout 7 ceph "${CLI_OPTS[@]}" mon add "${MON_NAME}" "${MON_IP}" || true
+      v2v1=$(ceph-conf -c /etc/ceph/${CLUSTER}.conf 'mon host' | tr ',' '\n' | grep -c ${MON_IP} || true)
       # with v2 only : [v2:xxxx:3300]
-      else
+      if [ ${v2v1} -eq 1 ]; then
         timeout 7 ceph "${CLI_OPTS[@]}" mon add "${MON_NAME}" "${MON_IP}":"${MON_PORT}" || true
+      # in case of explicit v2+v1 configuration : [v2:xxxx:3300,v1:xxxx:6789]
+      # or implicit v2+v1 configuration : xxxx
+      else
+        timeout 7 ceph "${CLI_OPTS[@]}" mon add "${MON_NAME}" "${MON_IP}" || true
       fi
     fi
   fi


CoRfr:
This is for stable-4.0, but the issue is also valid on master branch.

dsavineau:
What was the scenario where the ceph-conf command was failing ?

CoRfr:
I tested and {{ceph-conf -c /etc/ceph/ceph.conf 'mon host'}} would return nothing.

Partial content of my ceph.conf is:
{noformat}
[mon]


    [mon.carmd-ev-lxcim1]
      host = carmd-ev-lxcim1
      mon_addr = 10.1.60.1
      mon_allow_pool_delete = true
    [mon.carmd-ev-lxcim2]
      host = carmd-ev-lxcim2
      mon_addr = 10.1.60.2
    [mon.carmd-ev-lxlb01]
      host = carmd-ev-lxcim3
      mon_addr = 10.1.60.3
{noformat}

dsavineau:
Ok that's because you're using a `mon.` entry for each monitor instead of the global `mon` section.
So in this case the `|| true` changes the way the node will be added in the cluster because the `else` statement will add the mon to the crushmap only with v2 configuration but you're ceph configuration is using v2+v1 implicitly (ie: without v2:xxxx:3300,v1:xxxx:6789)

CoRfr:
I think I see what you mean, I updated the logic of the change.
I'm still not sure it handle the per-host declaration properly though.

doransmestad:
I just ran into this issue as well.  My ceph config file looks very similar to @CoRfr's - just have the individual mon entries, nothing in the global section as that is what ceph-container generated for me when I originally set the cluster up.  Using the code in this PR fixed it and I was able to start up the mon node again.

@dsavineau  are you indicating that we should update our config file to use the global section instead of the individual entries?   Or will the format continue to be supported and this will get merged in?   Thanks!

CoRfr:
I look at this issue today again to try to provide a proper fix, and actually I don't know how the command `ceph-conf -c /etc/ceph/${CLUSTER}.conf 'mon host'` from commit `8763b14a "daemon/start_mon.sh: add mon without port value"` can work.

At the moment my ceph.conf is the following:
{noformat}
[mon]
  
    [mon.carmd-ev-lxcim1]
      host = carmd-ev-lxcim1
      mon_addr = 10.1.60.1
    [mon.carmd-ev-lxcim2]
      host = carmd-ev-lxcim2
      mon_addr = 10.1.60.2
    [mon.carmd-ev-lxcim3]
      host = carmd-ev-lxcim3
      mon_addr = 10.1.11.231
{noformat}

Even if I convert the mon_addr values to v2+v1 configuration format [v2:xxxx:3300,v1:xxxx:6789], I don't even see what the configuration should look like to behave properly.

It looks like `ceph-conf -c /etc/ceph/${CLUSTER}.conf 'mon host'` should return something like:
`[v2:10.1.60.1:3300,v1:10.1.60.1:6789],[v2:10.1.60.2:3300,v1:10.1.60.2:6789],[v2:10.1.11.231:3300,v1:10.1.11.231:6789]` in my case, meaning a comma-separated list of address declarations, but I couldn't find the code that updates the ceph.conf to provide this, and even if set manually I tried with:
```
[mon]
  host = [v2:10.1.60.1:3300,v1:10.1.60.1:6789],[v2:10.1.60.2:3300,v1:10.1.60.2:6789],[v2:10.1.11.231:3300,v1:10.1.11.231:6789]
```
and that doesn't work.

So I'm still lost at this stage. @dsavineau could you please provide a sample `ceph.conf` that works?

PR #1446, opened 2019-08-05
osd_volume_activate: manage ceph-disk based OSD

dsavineau:
After upgrading to nautilus, the OSD deployed by ceph-disk needs to be
manage by ceph-volume with the simple subcommand.
This was part of the osd_ceph_disk_activate entrypoint but there's no
need to have a separate entrypoint because at the end we are using the
same command : ceph-volume.


guits:
what about osd_disk_activate.sh ?

PR #1457, opened 2019-08-24
Add ability to execute from Ceph build and build containers for CI/testing

dmick:
This branch modifies contrib/build-push-ceph-container-imgs.sh to allow use from a Ceph build to push a centos7 daemon-base container to a repo of choice.  There are also some cleanups and refactors for clarity, which I've tried to isolate in separate commits.

I expect this not to change the behavior of this script in other contexts.

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -68,7 +68,7 @@ function install_docker {
     sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes docker-ce
   fi
   sudo systemctl start docker
-  sudo systemctl status docker
+  sudo systemctl status --no-pager docker
   sudo chgrp "$(whoami)" /var/run/docker.sock
 }
 
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -59,10 +59,14 @@ function cleanup_previous_run {
 
 declare -F install_docker ||
 function install_docker {
-  curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-  sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
-  sudo apt-get update
-  sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes docker-ce
+  if [[ -x /usr/bin/yum ]] ; then
+    sudo yum install -y docker
+  else
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+    sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+    sudo apt-get update
+    sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes docker-ce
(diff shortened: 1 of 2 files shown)


contrib/build-push-ceph-container-imgs.sh
```diff
@@ -171,6 +196,15 @@ function build_and_push_latest_bis {
 declare -F push_ceph_imgs_latest ||
 function push_ceph_imgs_latest {
   local latest_name
+
+  if [[ ${CI_CONTAINER} ]] ; then
```
leseb: `if ${CI_CONTAINER}; then` so we stay consistent.

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -6,6 +6,20 @@ set -ex
 #############
 # VARIABLES #
 #############
+CI_CONTAINER=${CI_CONTAINER:=false}
+if ${CI_CONTAINER} ; then
+  # save the ceph branch, from the parent job
+  CEPH_BRANCH=${BRANCH}
+  # must be set by caller (perhaps another Jenkins build job)
+  # CONTAINER_REPO_HOSTNAME="quay.io"
```
leseb: Should we do some validation or are we 100% sure these vars will be present if `CI_CONTAINER` is true?

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -122,26 +141,37 @@ function create_head_or_point_release {
...
+  if ${CI_CONTAINER}; then
+    FLAVORS="${CEPH_BRANCH},centos,7"
+    make FLAVORS=${FLAVORS} \
+         CEPH_DEVEL="true" \
+         RELEASE=${RELEASE} \
+         TAG_REGISTRY=${CONTAINER_REPO_ORGANIZATION} \
+         IMAGES_TO_BUILD=daemon-base \
+         build.parallel
+  else
+    make CEPH_DEVEL=${DEVEL} RELEASE=${RELEASE} TAG_REGISTRY=${CONTAINER_REPO_ORGANIZATION} build.parallel
```
leseb: I don't see `CONTAINER_REPO_ORGANIZATION` declare anywhere if `CI_CONTAINER` is false, so I think that going to break. Can we just use `make CEPH_DEVEL=${DEVEL} RELEASE="$RELEASE" build.parallel` since we are in the `else`? Thanks

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -6,15 +6,28 @@ set -ex
...
   # save the ceph branch, from the parent job
   CEPH_BRANCH=${BRANCH}
   # must be set by caller (perhaps another Jenkins build job)
+  # BRANCH (as above, the Ceph branch)
+  # SHA1 (sha1 corresponding to the Ceph branch)
   # CONTAINER_REPO_HOSTNAME="quay.io"
   # CONTAINER_REPO_ORGANIZATION="cephci"
   # CONTAINER_REPO_USERNAME=user
   # CONTAINER_REPO_PASSWORD=password
+  for v in BRANCH SHA1 CONTAINER_REPO_HOSTNAME CONTAINER_REPO_ORGANIZATION \
```
leseb: If the list might change we could make it an array and iterate over it, if not let's keep it this way.

dmick:
Repushed with squashing/fixes as above; let me know if there's anything else you think needs fixing.

PR #1458, opened 2019-08-26
Fix lum

leseb:
Description of your changes:

Resolves #

Checklist:


leseb:
> This PR breaks the current iscsi content for nautilus because the URL for the repo is the using 2.x instead of 3

Actually no because `https://download.ceph.com/ceph-iscsi/2/rpm/el__ENV_[BASEOS_TAG]__/ceph-iscsi.repo` points to `latest` which points to ceph-iscsi 3.

leseb:
> Well that's not intuitive at all :)
> Maybe we should create the yum repo file for 3.x the same way than you did for 2.x to get the content more readable

I know I asked to change this here: https://github.com/rook/rook/issues/3662#issuecomment-524320596
But yes we could generate the repo file to clarify. It'd be better if someone could move the repo file... @djgalloway? Can you help with that? Thanks

PR #1459, opened 2019-08-26
centos: Update ceph-iscsi repositories

dsavineau:
Download the dedicated ceph-iscsi yum repository file for 2.x and 3.x
releases.


leseb:
I think there is one more thing though, it looks like some packages are missing, so I think we need to adapt `__ISCSI_PACKAGES__` for Nautilus and above.

I see failure like `No package ceph-iscsi available.` on Nautilus here: https://2.jenkins.ceph.com/job/ceph-container-build-ceph-base-push-imgs/102/console

PR #1466, opened 2019-08-29
tweaks: disable obtain_device_list_from_udev lvm

dsavineau:
Because we're using lvm commands from the container but udev isn't
running then we also need to disable obtain_device_list_from_udev.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1676612


leseb:
We need to force build so that Rook image can get this

PR #1470, opened 2019-09-04
exec: Remove ERR trap so other traps function

nakato:
As far as I can tell, the ERR trap is overriding all other traps defined
here, so they are not being called.  Meanwhile the ERR trap seems to only
cover the few lines that follow it until wait happens, for as soon as the
trap is entered the ERR trap is unset.

As the ERR trap is only covering a few lines that should be completely
reliable, and isn't handling cleanup for any errors that occur before
`exec` is called, this should be safe to remove.

This allows further development of the traps at hand here, such as
passing SIGHUP through.

### DIFF
--- a/src/daemon/docker_exec.sh
+++ b/src/daemon/docker_exec.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-# use -E so that trap ERR works with set -e
-set -E
-
 function do_stayalive {
   if [ -z "$STAYALIVE" ]; then
     return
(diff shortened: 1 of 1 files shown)


nakato:
Build failure is unrelated, not sure how to re-run it.

```
Downloading packages:

http://download.ceph.com/rpm-nautilus/el7/x86_64/librados2-14.2.3-0.el7.x86_64.rpm: [Errno 14] HTTP Error 404 - Not Found
```

nakato:
Ping for stalebot.

nakato:
stalebot

nakato:
Is there something I've missed with the submission of this patch?

nakato:
ping

PR #1474, opened 2019-09-11
fix: delete temporary mount point

### LINKED ISSUE #1473: temp dirs used to mount devices are never deleted
Is this a bug report or feature request?
* Bug Report

**Bug Report**

What happened:

The disk_list.sh script that runs during container startup never removes the temp dir it creates to mount and manage data volumes. We ended up in a situation where we had 100 000 empty folders on our OSD nodes.

It is not preventing ceph from working properly but restarting an OSD takes several minutes (chmod ceph:ceph runs on each folder).

What you expected to happen:

The temporary folders should have be deleted when the volume is umounted.

How to reproduce it (minimal and precise):

1. Deploy ceph luminous (containerized)
2. Check the /var/lib/ceph/tmp folders on an OSD node
3. Restart one or several OSDs on the node
4. Check the tmp folder again, more folders empty temp will be there

**Environment**:
* OS (e.g. from /etc/os-release): CentOS 7
* Kernel (e.g. `uname -a`): Linux 3.10.0-957.27.2.el7.x86_64
* Docker version (e.g. `docker version`): Docker version 1.13.1
* Ceph version (e.g. `ceph -v`): ceph version 12.2.12

gpocentek:
Description of your changes:
[file] disk_list.sh: delete the temporary folder created as a mount point for
Ceph devices.

Resolves #1473

Checklist:

### DIFF
--- a/src/daemon/disk_list.sh
+++ b/src/daemon/disk_list.sh
@@ -6,8 +6,6 @@ set -e
 # VARIABLES #
 #############
 
-mkdir -p /var/lib/ceph/tmp/
-tmp_dir="$(mktemp --directory --tmpdir=/var/lib/ceph/tmp/)"
 DOCKER_ENV=""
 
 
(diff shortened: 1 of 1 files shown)


gpocentek:
Thanks for the review @dsavineau 

`tmp_dir` being a global variable it didn't feel right to delete the associated folder in a function. The change also makes the re-use of mount/unmount functions more secure, even though it's not needed right now.

Would you prefer to have `rmdir`called after the `umount_ceph_data`without any other change?

src/daemon/disk_list.sh
```diff
@@ -89,6 +89,7 @@ function get_docker_env () {
 
   cd || return
   umount_ceph_data
```
dsavineau: missing "$tmp_dir" arg on the umount_ceph_data call

gpocentek:
I've reworked the code a bit to make it more readable (hopefully).

PR #1481, opened 2019-09-27
daemon-base: add ceph-mgr-ssh package

liewegas:


dsavineau:
@liewegas @leseb does this need to be backported to stable-4.0 branch ? (ie: for nautilus stable image upstream and rhcs 4.0 downstream)

liewegas:
I don't think we need it for nautilus.  Not now at least!

PR #1488, opened 2019-10-08
ubi8: use quotes for LABEL version

dsavineau:
This was done downstream and to avoid conflict during sync between
upstream and downstream we should use the same syntax.


ktdreyer:
@b-ranto FYI changes need to go into GitHub so Dimitri's reconciliation process is easier

PR #1489, opened 2019-10-09
src/daemon-base: add smartmontools and nvme-cli packages

liewegas:
The mon and osd need smartctl and nvme to scrape metrics from HDDs/SSDs.

### DIFF
--- a/src/daemon-base/__CEPH_BASE_PACKAGES__
+++ b/src/daemon-base/__CEPH_BASE_PACKAGES__
@@ -11,6 +11,8 @@
         kmod \
         lvm2 \
         gdisk \
+	smartmontools \
+	nvme-cli \
         __RADOSGW_PACKAGES__ \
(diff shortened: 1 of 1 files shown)


dsavineau:
What's the reason to add this package ? because it's not explained in the commit message

liewegas:
The OSD and mon use it to scrape smart metrics from devices.  I'll update the comit

liewegas:
I get smartmontools-7.0-1.el7.x86_64 in the ceph/daemon-base container.  not sure how to tell which repo it's coming from?

liewegas:
It drags in mailx as a dependency, this is annoying; not sure if it's worth doing some rpm-fu to force removal of that.  (smartmontools has a smartd daemon that sends email; we don't use it as we only need the smartctl utility.)

liewegas:
jenkins test this please

PR #1492, opened 2019-10-17
Rework OSD memory tuning

dsavineau:
This moves the tune_memory function from the start_mon.sh file to the common_functions files.
There's no reason to do the memory tunning on the MON nodes. This should be done on the OSD nodes instead.
Also this process is done when bluestore is used and for demo scenario.


dsavineau:
Should we drop it from ceph-ansible too ? or it still makes sense for HCI deployment ?

https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/templates/ceph.conf.j2#L92-L107

leseb:
> Should we drop it from ceph-ansible too ? or it still makes sense for HCI deployment ?
> 
> https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/templates/ceph.conf.j2#L92-L107

If the Ceph version is at least Nautilus, yes drop it.

jdurgin:
> Should we drop it from ceph-ansible too ? or it still makes sense for HCI deployment ?
> 
> https://github.com/ceph/ceph-ansible/blob/master/roles/ceph-config/templates/ceph.conf.j2#L92-L107

We still need ansible to set the osd_memory_target like it does there to match the host - it's just the other options that are no longer necessary.

PR #1504, opened 2019-11-04
ceph: temporarily use remoto from pip

leseb:
Ceph is in the process of moving to python 3 only but in the meantime we need a higher version of remoto which has a better support of python 2.

This can get reverted in a few weeks when that transition is finished.

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -33,9 +33,13 @@ bash -c ' \
   if [[ "${CEPH_VERSION}" =~ master|^wip* ]] || ${CEPH_DEVEL}; then \
     REPO_URL=$(curl -s "https://shaman.ceph.com/api/search/?project=ceph&distros=centos/__ENV_[BASEOS_TAG]__&flavor=default&ref=${CEPH_VERSION}&sha1=latest" | jq -a ".[0] | .url"); \
     RELEASE_VER=0 ;\
+    yum install -y python-pip ; \
+    pip install -U remoto ; \
+    yum remove -y python-pip ; \
   else \
     RELEASE_VER=1 ;\
     REPO_URL="http://download.ceph.com/rpm-${CEPH_VERSION}/el__ENV_[BASEOS_TAG]__/"; \
   fi && \
   rpm -Uvh "$REPO_URL/noarch/ceph-release-1-${RELEASE_VER}.el__ENV_[BASEOS_TAG]__.noarch.rpm" ' && \
 yum install -y __CEPH_BASE_PACKAGES__
+


leseb:
@liewegas can you elaborate why it's needed and when we can expect to revert this? Thanks!

liewegas:
The centos 7 ceph container current uses python 2, and the version of remoto packaged for python 2 is missing support for the ssh_options argument to the connection (which the ssh orchestrator needs).  The python3 remoto package is new enough that this works out, but we aren't quite ready to switch to python3-only due to some missing packages in EPEL (see https://github.com/ceph/ceph/pull/25969).

liewegas:
Note that this doesn't really have anything to do with teuthology, except that I can't write a teuthology test to exercise ceph-daemon + ssh orchestrator until the ssh orchestrator works when running out of the ceph-container image.  (For now I'm testing with a vstart cluster and then deploying new daemons from there.)

dsavineau:
> The centos 7 ceph container current uses python 2, and the version of remoto packaged for python 2 is missing support for the ssh_options argument to the connection (which the ssh orchestrator needs)

So the ceph-mgr-ssh dependency is missing some version control on remoto ? [1]

Why not just bump the python2 remoto version on EPEL ? because using pip in the container is the last thing we want to do.

[1] https://github.com/ceph/ceph/blob/master/ceph.spec.in#L600

liewegas:
> > The centos 7 ceph container current uses python 2, and the version of remoto packaged for python 2 is missing support for the ssh_options argument to the connection (which the ssh orchestrator needs)
> 
> So the ceph-mgr-ssh dependency is missing some version control on remoto ? [1]
> 
> Why not just bump the python2 remoto version on EPEL ? because using pip in the container is the last thing we want to do.

That would also solve the problem.  Whatever is easiest!

dsavineau:
@alfredodeza I don't know if you're the right person for that but because you're the remoto author maybe you can help on that one.
Is it possible to update the remoto pkg for py2 on epel ? If yes what would be the delay ?

I don't mind to update remoto  via pip but only as a temporary solution until we either update the py2 version or switch to py3.

liewegas:
Since it's temporary I'd lean toward updating the container image with pip and letting @alfredodeza focus on the py3 dependencies!

alfredodeza:
We are blocked on getting the latest version of remoto on EPEL. A few things need to happen:

* apipkg needs to be updated https://src.fedoraproject.org/rpms/python-apipkg/pull-request/4
* execnet needs to get bumped once `apipkg` is available at the newer version

Once those two are in, we can try to get the latest version (1.1.4 as of today) in EPEL. 

@ktdreyer does this sound accurate to you? 

I am OK with installing remoto via pip _for development purposes only_. We should **not** release an official image without dependencies being available.

liewegas:
Having this only on the (ceph) master branch and wip containers would be ideal!

dsavineau:
The failure in the CI (master only) is unrelated

ktdreyer:
The maintainer (@thmo) has merged our [epel7 update to apipkg](https://src.fedoraproject.org/rpms/python-apipkg/pull-request/4) today. When he builds that, we can build the updates to execnet and remoto for epel 7 as well.

dsavineau:
Thanks for the update ! @ktdreyer

thmo:
Which is the minimal version of execnet needed in epel7?

alfredodeza:
@thmo 1.7.1

PR #1506, opened 2019-11-06
centos: run pip after rpm package installation

dsavineau:
To update remoto via pip we need to run the upgrade process after the
package installation otherwise we will still have the old remoto version
from the rpm package.


liewegas:
can we merge this before you head out for the day?

PR #1507, opened 2019-11-06
ceph: need openssh-clients for remoto to work

liewegas:
I seem to need openssh-clients installed for remoto to work.
This should probably be a dependency of the package?

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -41,7 +41,7 @@ bash -c ' \
 yum install -y __CEPH_BASE_PACKAGES__ && \
 bash -c ' \
   if [[ "${CEPH_VERSION}" =~ master|^wip* ]] || ${CEPH_DEVEL}; then \
-    yum install -y python-pip ; \
+    yum install -y python-pip openssh-clients ; \
     pip install -U remoto ; \
     yum remove -y python-pip ; \
   fi '


dsavineau:
@alfredodeza can you confirm that ?

> This should probably be a dependency of the package?

It doesn't seem to come from remoto but execnet python library.
But I'm wondering if this should be fixed in ceph-mgr-ssh instead.

ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
```diff
@@ -44,4 +44,5 @@ bash -c ' \
     yum install -y python-pip ; \
     pip install -U remoto ; \
     yum remove -y python-pip ; \
+    yum install -y openssh-clients ; \
```
dsavineau: I think we can combined the openssh-clients package installation with python-pip instead of using yum install twice

liewegas: fixed

liewegas:
> @alfredodeza can you confirm that ?
> 
> > This should probably be a dependency of the package?
> 
> It doesn't seem to come from remoto but execnet python library.
> But I'm wondering if this should be fixed in ceph-mgr-ssh instead.

I'm guessing the remoto or execnet package would be the right one to add the dependency and not ceph-mgr-ssh?  In any case, still going for the quick fix here.

dsavineau:
> I'm guessing the remoto or execnet package would be the right one to add the dependency and not ceph-mgr-ssh?

yes and no, it depends if the libraries manage multiple backends or not.
remoto seems to support ssh, docker, k8s, openshift.. [1] so I don't think this package should depend on all backend method.
On the other hand, the ceph-mgr-ssh package is exclusif to ssh so...

[1] https://github.com/alfredodeza/remoto/blob/master/remoto/connection.py#L25-L34

PR #1509, opened 2019-11-08
customize root's prompt inside the container

liewegas:
This makes things less confusing for humans shelling into the container.


liewegas:
@pcuznor

dsavineau:
I don't see why this should be done in the ceph container.
daemon-base is only about ceph binaries (+ deps) but not custom configuration.
Also this should be done by the users via the environment variables in docker/podman.

```console
$ docker run --rm -it -e PS1='[ceph: \u@\h:\l \W]\\$ ' ceph/daemon-base:latest-master bash
[ceph: root@f0087d00cc77:0 /]$
```

liewegas:
Yes--good suggest, I can just do this in ceph-daemon when i start a shell container.

PR #1511, opened 2019-11-12
daemon-base: include less

liewegas:
This is purely for human users of the container.  AFAICS the other staples
are there (grep, jq, watch), but not less... and more suuuuucks.


dsavineau:
> This is purely for human users of the container. AFAICS the other staples
> are there (grep, jq, watch)

`grep` and `watch` are coming from the centos 7 based image.
`jq` is used at build time for retrieving information from shaman api.

From my point of view, we shouldn't include `less`. `more` is already present in the container image and if you absolutely want to use `less` then you can still install it with yum.

PR #1517, opened 2019-11-21
nfs: run a dedicated dbus daemon for nfs-ganesha

guits:
This runs a dedicated dbus daemon for nfs-ganesha in order to avoid
using the host dbus socket so we don't have to launch nfs-ganesha with
--privileged option.

Related ceph-ansible PR: ceph/ceph-ansible#4760

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1725254

### DIFF
--- a/src/daemon/start_nfs.sh
+++ b/src/daemon/start_nfs.sh
@@ -8,12 +8,19 @@ function start_rpc {
 
 }
 
+function start_dbus {
+  mkdir -p /run/dbus
+  dbus-daemon --system
+}
+
 function start_nfs {
   get_config
   check_config
 
   # Init RPC
   start_rpc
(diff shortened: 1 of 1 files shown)


src/daemon/start_nfs.sh
```diff
@@ -33,6 +38,8 @@ function start_nfs {
   # create ganesha log directory since the package does not create it
   mkdir -p /var/log/ganesha/
 
+  start_dbus
```
dsavineau: could you move this just after the start_rpc call ?

guits: non maggle

dsavineau: :-1:

PR #1524, opened 2019-11-26
ubi8: add ceph-mgr-k8sevents package

dsavineau:
This adds the ceph-mgr-k8sevents rpm package to the RHEL 8 container
image for OCS/rook-ceph usage.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1763918

### DIFF
--- a/ceph-releases/ALL/ubi8/daemon-base/__CEPH_MGR_PACKAGES__
+++ b/ceph-releases/ALL/ubi8/daemon-base/__CEPH_MGR_PACKAGES__
@@ -1,0 +1,6 @@
+ceph-mgr__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-dashboard__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-diskprediction-local__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-k8sevents__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-rook__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-ssh__ENV_[CEPH_POINT_RELEASE]__


ceph-releases/ALL/ubi8/daemon-base/__CEPH_MGR_PACKAGES__
```diff
@@ -0,0 +1,6 @@
+ceph-mgr__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-dashboard__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-diskprediction-local__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-k8sevents__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-rook__ENV_[CEPH_POINT_RELEASE]__ \
+ceph-mgr-ssh__ENV_[CEPH_POINT_RELEASE]__
```
ktdreyer: There are many more packages in this list than `ceph-mgr-k8sevents`. Would you please help me understand why we list all the other packages here? The commit log does not mention any of these.

b-ranto: This does seem to override all the ceph-mgr packages that are being installed in the ubi8 image. Some of these were already included in our previous builds (all apart from k8sevents and ssh package). I suppose the ssh package was recently included to the list as well.

dsavineau: That's how ceph-container overrides are working. [1]

Before that PR ubi8 was using the common ceph mgr file (`src/daemon-base/__CEPH_MGR_PACKAGES__`) [2].

When we want to override it for ubi8 we need to create a dedicated copy in `ceph-releases/ALL/ubi8/daemon-base`.
We can't just add the specific k8sevents package, we have to redefine the whole list.

The ceph-mgr-ssh package in ubi8 list is a mistake and I have already fix it via https://github.com/ceph/ceph-container/pull/1526

[1] https://github.com/ceph/ceph-container/blob/master/CONTRIBUTING.md#where-does-should-source-code-live
[2] https://github.com/ceph/ceph-container/blob/master/src/daemon-base/__CEPH_MGR_PACKAGES__

ktdreyer: Thanks @dsavineau . The reason I ask about this is that I wanted to understand 1) When we would revert this change (when we release v14.2.5 upstream?) and 2) What exactly the revert should look like.

dsavineau: > 1) When we would revert this change (when we release v14.2.5 upstream?) 

Yes, when v14.2.5 upstream will be release

> 2) What exactly the revert should look like.

Removing the `ceph-releases/ALL/ubi8/daemon-base/__CEPH_MGR_PACKAGES__` file and add the k8sevents package in `src/daemon-base/__CEPH_MGR_PACKAGES__`

Note that this won't impact downstream (ie: we won't have to resync for this because the dockerfile won't change for ubi8)

Issue #1528, opened 2019-12-02
ceph/ceph: add vX.Y.Z tag


liewegas:
We have ``x.y.z-$date`` tags, but not ``x.y.z`` tags, which means we can't (easily) programatically just ask for a particular version of ceph.

PR #1529, opened 2019-12-02
contrib: add vX.Y.Z container tag on ceph/ceph

dsavineau:
Only vX and vX.Y were created on ceph/ceph.
The full vX.Y.Z tag was present but with a date suffix.
Let's add it by default.

Closes: #1528


leseb:
I forced a build here https://2.jenkins.ceph.com/job/ceph-container-build-ceph-base-push-imgs/, images are already showing up.

dsavineau:
Thanks Seb !

PR #1533, opened 2019-12-11
ceph-mgr-ssh -> ceph-mgr-cephadm (for master/octopus only!)

liewegas:
Remove ceph-mgr-ssh entirely from nautilus (and earlier).


liewegas:
@dsavineau am I doing this right?

liewegas:
https://github.com/ceph/ceph-container/pull/1534 is simpler and passes the ci check; we can do something more elegant later?

PR #1534, opened 2019-12-12
daemon-base: drop ceph-mgr-ssh, except for the master/wip-* special case

liewegas:
This is an ugly kludge but it will work for the time being until something
more elegant is sorted out.

Signed-off-by: Sage Weil <sage@redhat.com>

PR #1535, opened 2019-12-12
daemon-base: add cephadm mgr package for octopus

dsavineau:
This is a follow up on #1534.
Because ceph-mgr-cephadm is only available on octopus then we need a
dedicated copy of the ceph mgr packages for nautilus.


liewegas:
This brings up (again) the handling for the various release branches.. what happens if i have wip-foo?  Is that the ALL variant?  What about wip-foo-nautilus?  (We use the convention that a -$release suffix is used to tell the build/ci to use the various rules for that release.. so in this case wip-foo-nautilus should follow the nautilus template.)

leseb:
Thanks for the followup Dim!

Issue #1536, opened 2019-12-13
Containers restart causes lot of unnecessary IO


gfidente:
It takes ~4 minutes in upgrades PER container restart and when deploying, it stops all containers and restart ALL of them at the same time so 

find -L /var/lib/ceph/ -mindepth 1 -maxdepth 3 -exec chown ceph. {}  is running, in this case 16x in parallel and causing lots of IOs.

gfidente:
@dsavineau @guits not sure if you guys can help get this going and backported in stable-3 ?

dsavineau:
> It takes ~4 minutes in upgrades PER container restart and when deploying, it stops all containers and restart ALL of them at the same time so

Do you have some logs to see what directories/files are the root cause ?

Because each OSD container has only access to its own OSD data even if the /var/lib/ceph directory is bind mount for every container (ie: each container mount its OSD partition inside the container).

david-hill:
Customer was using 3-37 which is the latest available.   We do see a chown for all ceph containers on each deploy/update.

david-hill:
I can easily reproduce this issue by simply running "openstack overcloud deploy" in this customer environment.  It'll stop all ceph containers on a node and then restart all of them .   They will appear as "down" in "ceph -s" for about 4 minutes and if we login that node while this is happening, we can see the chown running for those ~4 minutes.   I can ask the customer to provide the logs if we need this...

dsavineau:
> They will appear as "down" in "ceph -s" for about 4 minutes and if we login that node while this is happening, we can see the chown running for those ~4 minutes.

Even if I agree with the fix on the find/chown command to speed up the restart, I really want to understand what's the underlying issue instead of just hidde it.

So it would be nice too see what's the output of `find -L /var/lib/ceph/ -mindepth 1 -maxdepth 3 -ls`

dsavineau:
@david-hill any news on this ?

david-hill:
I have to find back the case ...

gfidente:
@dsavineau it looks to me that in a few places, including daemon startup scripts, we are calling the `create_mandatory_directories` function as a safety measure to make sure the directories are all there and owned by `ceph:` ... as a side effect, we run `chown` against a directory which is already populated with lots of files (data) for example when containers are restarted

dsavineau:
> as a side effect, we run chown against a directory which is already populated with lots of files (data) for example when containers are restarted

[...]

PR #1537, opened 2019-12-13
daemon: speed up create_mandatory_directories

gfidente:
Makes find skip files already owned by 'ceph' user when launching
chown command; chown will unconditionally re-apply the ownership
causing a lot of writes.

It looks like we use this function not only when it is necessary
to create the directories but also when there is existing data into
those, as a safety measure.

Resolves: #1536
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1784047
Signed-off-by: Giulio Fidente <gfidente@redhat.com>


gfidente:
thanks for the review, will do

PR #1547, opened 2019-12-18
centos: disable python3 update for CentOS 8

dsavineau:
On some GNU/Linux distribution, when updating the CentOS 8 packages in
the container base image then the platform-python package is updated.
After the update the python interpreter is broken so we can run yum/dnf
commands anymore

yum: /usr/libexec/platform-python: bad interpreter: Permission denied

As a temporary solution we can disable the platform-python and
python3-libs package update.


dsavineau:
@ktdreyer @liewegas @leseb I don't know if one of you already saw this.
Note that we only have the issue on the ubuntu xenial nodes on jenkins.

PR #1549, opened 2019-12-19
contrib: fix centos release for ceph-ci build

dsavineau:
This also add a new function to get the centos release from a ceph
release/branch

### DIFF
--- a/contrib/build-push-ceph-container-imgs-arm64.sh
+++ b/contrib/build-push-ceph-container-imgs-arm64.sh
@@ -22,7 +22,8 @@ function install_docker {
 function build_ceph_imgs {
   echo "Build Ceph container image(s)"
   for ceph_release in "${CEPH_RELEASES[@]}"; do
-    make DAEMON_BASE_TAG=daemon-base:"$RELEASE"-"${ceph_release}"-centos-7-aarch64 DAEMON_TAG=daemon:"$RELEASE"-"${ceph_release}"-centos-7-aarch64 RELEASE="$RELEASE" FLAVORS="${ceph_release},centos-arm64,7" BASEOS_REPO=centos build
+    CENTOS_RELEASE=$(_centos_release ${ceph_release})
+    make DAEMON_BASE_TAG=daemon-base:"$RELEASE"-"${ceph_release}"-centos-${CENTOS_RELEASE}-aarch64 DAEMON_TAG=daemon:"$RELEASE"-"${ceph_release}"-centos-${CENTOS_RELEASE}-aarch64 RELEASE="$RELEASE" FLAVORS="${ceph_release},centos-arm64,${CENTOS_RELEASE}" BASEOS_REPO=centos build
   done
   docker images
 }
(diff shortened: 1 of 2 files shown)


contrib/build-push-ceph-container-imgs-arm64.sh
```diff
@@ -22,15 +22,17 @@ function install_docker {
 function build_ceph_imgs {
   echo "Build Ceph container image(s)"
   for ceph_release in "${CEPH_RELEASES[@]}"; do
-    make DAEMON_BASE_TAG=daemon-base:"$RELEASE"-"${ceph_release}"-centos-7-aarch64 DAEMON_TAG=daemon:"$RELEASE"-"${ceph_release}"-centos-7-aarch64 RELEASE="$RELEASE" FLAVORS="${ceph_release},centos-arm64,7" BASEOS_REPO=centos build
+    CENTOS_RELEASE=$(_centos_release ${ceph_release})
```
djgalloway: Forgive my lack of BASH knowledge but:

1. If this is set at the top of the script, will `CENTOS_RELEASE` carry on through the rest of the script?
2. Is the `build_ceph_imgs` function *always* called?  Just want to make sure we don't accidentally end up with an empty `CENTOS_RELASE`

dsavineau:
> I learned there are no docker systemd unit files in CentOS8. This morning, I created a CentOS8 container using this script by just commenting out these three lines. So can we put an `if` statement only running them if we're not on CentOS8?
> 
> [295730c#diff-9bd28db8c71e92c690fefd2d189c6b46R79-R81](https://github.com/ceph/ceph-container/commit/295730cba647817ce0f545ffc806b8a657a20631#diff-9bd28db8c71e92c690fefd2d189c6b46R79-R81)
> 
> ```
>   sudo systemctl start docker
>   sudo systemctl status --no-pager docker
>   sudo chgrp "$(whoami)" /var/run/docker.sock
> ```

Yes that's because CentOS 8 only uses `podman` instead of `centos` but that's not the purpose of this PR.
AFAIK the sepia nodes used for the container build jobs are using ubuntu so that's fine for the moment.

If we plan to use another distribution on those nodes then we will create a dedicated PR for that.

djgalloway:
> > I learned there are no docker systemd unit files in CentOS8. This morning, I created a CentOS8 container using this script by just commenting out these three lines. So can we put an `if` statement only running them if we're not on CentOS8?
> > [295730c#diff-9bd28db8c71e92c690fefd2d189c6b46R79-R81](https://github.com/ceph/ceph-container/commit/295730cba647817ce0f545ffc806b8a657a20631#diff-9bd28db8c71e92c690fefd2d189c6b46R79-R81)
> > ```
> >   sudo systemctl start docker
> >   sudo systemctl status --no-pager docker
> >   sudo chgrp "$(whoami)" /var/run/docker.sock
> > ```
> 
> Yes that's because CentOS 8 only uses `podman` instead of `centos` but that's not the purpose of this PR.
> AFAIK the sepia nodes used for the container build jobs are using ubuntu so that's fine for the moment.
> 
> If we plan to use another distribution on those nodes then we will create a dedicated PR for that.

Actually, as far as containers built from ceph-ci.git packages, we are *only* using CentOS Jenkins slaves.  I could agree that perhaps this should be fixed in a separate PR.  Merging this one will at least get CentOS7 containers built again.

contrib/build-push-ceph-container-imgs-arm64.sh
dsavineau: > 1. If this is set at the top of the script, will `CENTOS_RELEASE` carry on through the rest of the script?

True, unless I set `CENTOS_RELEASE` as a local variable for each function. But this shouldn't be a problem because we always call the `_centos_release` function before using that variable so we shouldn't get a `CENTOS_RELEASE` value from a previous call.

> 2. Is the `build_ceph_imgs` function _always_ called?

Not always, it depends if we're building devel containers or using the ceph-ci scenario.

https://github.com/ceph/ceph-container/blob/master/contrib/build-push-ceph-container-imgs.sh#L319-L320

> Just want to make sure we don't accidentally end up with an empty `CENTOS_RELASE`

This isn't possible because the `CENTOS_RELASE` variable is always set by the `_centos_release` function which always returns a value.

PR #1550, opened 2019-12-19
contrib: Don't attempt to start systemd services on CentOS8

djgalloway:

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -76,9 +76,14 @@ function install_docker {
     sudo apt-get update
     sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes docker-ce
   fi
-  sudo systemctl start docker
-  sudo systemctl status --no-pager docker
-  sudo chgrp "$(whoami)" /var/run/docker.sock
+
+  # Podman doesn't have systemd unit files so we can't start them on CentOS8
+  CENTOS_RELEASE=$(_centos_release ${CEPH_BRANCH})
+  if [ $CENTOS_RELEASE != 8 ] || [[ -x /usr/bin/apt-get ]]; then
+    sudo systemctl start docker
+    sudo systemctl status --no-pager docker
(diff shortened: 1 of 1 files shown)


contrib/build-push-ceph-container-imgs.sh
```diff
@@ -76,9 +76,14 @@ function install_docker {
     sudo apt-get update
     sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes docker-ce
   fi
-  sudo systemctl start docker
-  sudo systemctl status --no-pager docker
-  sudo chgrp "$(whoami)" /var/run/docker.sock
+
+  # Podman doesn't have systemd unit files so we can't start them on CentOS8
+  CENTOS_RELEASE=$(_centos_release ${CEPH_BRANCH})
+  if [ $CENTOS_RELEASE != 8 ] || [[ -x /usr/bin/apt-get ]]; then
```
dsavineau: I'm not sure to understand why we're using the `CENTOS_RELEASE` here because it's only for the container image and not the OS of the node where the build is running.

djgalloway: This Jenkins job builds packages from branches in ceph-ci.git: https://github.com/ceph/ceph-build/blob/master/ceph-dev-new-build/build/build_rpm#L208

After those packages are pushed and a repo is created, the Jenkins slave that built the packages then runs `build-push-ceph-container-imgs.sh`.  If the Jenkins slave is running CentOS8, the container build fails because the docker systemd unit doesn't exist (and there is no podman equivalent).

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -76,9 +76,14 @@ function install_docker {
     sudo apt-get update
     sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes docker-ce
   fi
-  sudo systemctl start docker
```
liewegas: how about just ``|| true`` on all of these?

long term, we should switch to buildah, probably, right?

djgalloway:
https://github.com/ceph/ceph-container/pull/1551

PR #1552, opened 2019-12-19
contrib: add podman support

dsavineau:
This is for CentOS 8 distribution.

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -76,10 +76,13 @@ function install_docker {
     sudo apt-get update
     sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y --force-yes docker-ce
   fi
-  # ignore failure here; some distros (el8) don't need this
-  sudo systemctl start docker || true
-  sudo systemctl status --no-pager docker || true
-  sudo chgrp "$(whoami)" /var/run/docker.sock || true
+  sudo systemctl start docker
+  sudo systemctl status --no-pager docker
+  sudo chgrp "$(whoami)" /var/run/docker.sock
+}
+
+function install_podman {
+  sudo dnf install -y podman podman-docker
 }
 
 function login_docker_hub {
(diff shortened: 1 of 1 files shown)


contrib/build-push-ceph-container-imgs.sh
```diff
@@ -306,7 +309,11 @@ function create_registry_manifest {
 # MAIN #
 ########
 
-install_docker
+if [[ -x /usr/bin/dnf ]] ; then
+  install_podman
```
liewegas: we need to also s/docker/$DOCKER/ throughout and set DOCKER=podman here, right?

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -76,10 +76,13 @@ function install_docker {
...
-  sudo systemctl start docker || true
-  sudo systemctl status --no-pager docker || true
-  sudo chgrp "$(whoami)" /var/run/docker.sock || true
+  sudo systemctl start docker
+  sudo systemctl status --no-pager docker
+  sudo chgrp "$(whoami)" /var/run/docker.sock
+}
+
+function install_podman {
+  sudo dnf install -y podman
```
dsavineau: I think I need podman-docker package too right ?

djgalloway:
This makes way more sense.  I hadn't thought of checking for `/usr/bin/dnf` first.

Thanks for fixing and sorry for not taking a bigger stand against `|| true` and playing fast-and-loose with merges.

contrib/build-push-ceph-container-imgs.sh
djgalloway: For `docker` commands to work, yes.

contrib/build-push-ceph-container-imgs.sh
dsavineau: See my comment below but I think installation podman-docker is enough for now because this package provide the docker command as a wrapper to podman

```console
# docker version
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Version:       1.0.5
Go Version:    go1.11.6
OS/Arch:       linux/amd64
```

liewegas: much easier :)

PR #1554, opened 2019-12-20
centos: fix __GANESHA_PACKAGES__ check

liewegas:
Is this right?  I'm not sure what -n "a string" would accomplish
since that should (always) be the length of the string.

But if this should be -e, then the iscsi one below is also broken?

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -1,7 +1,7 @@
 yum install -y epel-release && \
 yum install -y jq && \
 bash -c ' \
-  if [ -n "__GANESHA_PACKAGES__" ]; then \
+  if [ -e "__GANESHA_PACKAGES__" ]; then \
     echo "[ganesha]" > /etc/yum.repos.d/ganesha.repo ; \
     echo "name=ganesha" >> /etc/yum.repos.d/ganesha.repo ; \
     if [[ "${CEPH_VERSION}" =~ master|^wip* ]]; then \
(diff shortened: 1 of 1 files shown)


liewegas:
> AFAIR the variable `__GANESHA_PACKAGES__` always exists, so using `-e` would also work. What we are checking with `-n` is that the variable has content (packages to install), so I'm not sure what's broken at this point.

``-n "$FOO"`` could do that, but AFAICS ``-n "FOO"`` is always true

leseb:
> > AFAIR the variable `__GANESHA_PACKAGES__` always exists, so using `-e` would also work. What we are checking with `-n` is that the variable has content (packages to install), so I'm not sure what's broken at this point.
> 
> `-n "$FOO"` could do that, but AFAICS `-n "FOO"` is always true

You're right, I guess that if this is true the whole things has been broken almost for the very beginning. @dsavineau thoughts?

liewegas:
> > > AFAIR the variable `__GANESHA_PACKAGES__` always exists, so using `-e` would also work. What we are checking with `-n` is that the variable has content (packages to install), so I'm not sure what's broken at this point.
> > 
> > 
> > `-n "$FOO"` could do that, but AFAICS `-n "FOO"` is always true
> 
> You're right, I guess that if this is true the whole things has been broken almost for the very beginning. @dsavineau thoughts?

hmm it is substituting magically somehow, though, because for master build i see

```
STEP 22: RUN     yum install -y epel-release && yum install -y jq && bash -c '   if [ -n "" ]; then     echo "[ganesha]" > /etc/yum.repos.d/ganesha.repo ; ....
```
but for the ceph-ci.git build i see
```
STEP 22: RUN     yum install -y epel-release && yum install -y jq && bash -c '   if [ -n "nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw nfs-ganesha-rados-grace" ]; then     echo "[ganesha]" > /etc/yum.repos.d/ganesha.repo ;     ...
```
Trying to figure out why they are different

liewegas:
aha,
```
gnit:ceph-container (master) 07:33 AM $ find ceph-releases/master/
ceph-releases/master/
ceph-releases/master/daemon
ceph-releases/master/daemon/__ETCD_PACKAGE__
ceph-releases/master/daemon/__DOCKERFILE_INSTALL__
ceph-releases/master/daemon/__KUBECTL_PACKAGE__
ceph-releases/master/daemon/__S3CMD_PACKAGE__
ceph-releases/master/daemon-base
ceph-releases/master/daemon-base/__GANESHA_PACKAGES__
ceph-releases/master/daemon-base/__DOCKERFILE_POSTINSTALL_CLEANUP__
```

liewegas:
see https://github.com/ceph/ceph-container/pull/1555

PR #1555, opened 2019-12-20
ALL: no ganesha packages for now

liewegas:

### DIFF
--- a/ceph-releases/luminous/daemon-base/__GANESHA_PACKAGES__
+++ b/ceph-releases/luminous/daemon-base/__GANESHA_PACKAGES__
@@ -1,0 +1,1 @@
+nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw nfs-ganesha-rados-grace
--- a/ceph-releases/mimic/daemon-base/__GANESHA_PACKAGES__
+++ b/ceph-releases/mimic/daemon-base/__GANESHA_PACKAGES__
@@ -1,0 +1,1 @@
+nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw nfs-ganesha-rados-grace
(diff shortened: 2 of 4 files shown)


liewegas:
This matches the empty file in the ``master`` dir.  What's weird is that I do'nt see ganesha packages defined for nautilus or most of hte other targets.. i'm not sure where the list is coming from.  In any case, though, I think this will fix our ceph-ci.git container builds

dsavineau:
> This matches the empty file in the `master` dir. What's weird is that I do'nt see ganesha packages defined for nautilus or most of hte other targets.. i'm not sure where the list is coming from. In any case, though, I think this will fix our ceph-ci.git container builds

https://github.com/ceph/ceph-container/blob/master/src/daemon-base/__GANESHA_PACKAGES__

I suppose the issue is when using ceph branches like "wip*" ?

Because right now the fix proposed breaks nautilus, mimic and luminous releases.

We should probably do the opposite thing:
  - create `ceph-releases/{nautilus,mimic,luminous}/daemon-base/__GANESHA_PACKAGES__` with the content of `src/daemon-base/__GANESHA_PACKAGES__`
  - remove `ceph-releases/master/daemon-base/__GANESHA_PACKAGES__`
  - empty `src/daemon-base/__GANESHA_PACKAGES__`

liewegas:
Like so?

dsavineau:
> Like so?

Yes this should fix the ganshe issue but considering that there's other stuff using `ceph-releases/master` [1] I think the wip branches will still be broken.

I guess a better approach for all of this would be to set the content not for a ceph release but for the centos 8 release instead.
That way it will work for both master and wip* branches.

[1] https://github.com/ceph/ceph-container/tree/master/ceph-releases/master

liewegas:
Can we merge this in the meantime and hopefully i'll get a working wip-* build?

liewegas:
For wip, the other thing to keep in mind is that we really want any unrecognized branch (master, wip-*) to use the 'master' release.  Then nautilus or *nautilus (e.g., wip-foo-nautilus) shoudl use nautilus.  ANd wip-*-mimic should use mimic.. etc.

PR #1556, opened 2019-12-20
daemon-base: Use `dnf` on CentOS8

djgalloway:

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -29,8 +29,8 @@ yum update -y --exclude=platform-python --exclude=python3-libs && \
 rpm --import 'https://download.ceph.com/keys/release.asc' && \
 bash -c ' \
   if [[ __ENV_[BASEOS_TAG]__ -eq 8 ]]; then \
-    yum install -y dnf-plugins-core ;\
-    yum copr enable -y ktdreyer/ceph-el8 ;\
+    dnf install -y dnf-plugins-core ;\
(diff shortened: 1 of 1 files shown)


djgalloway:
Seen in https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos8,DIST=centos8,MACHINE_SIZE=huge/34539/

```
No such command: copr. Please use /usr/bin/yum --help
It could be a DNF plugin command, try: "dnf install 'dnf-command(copr)'"
```

I don't think it's the only reason that container build failed but surely using the proper package manager can't hurt here?

dsavineau:
> Seen in https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos8,DIST=centos8,MACHINE_SIZE=huge/34539/
> 
> ```
> No such command: copr. Please use /usr/bin/yum --help
> It could be a DNF plugin command, try: "dnf install 'dnf-command(copr)'"
> ```
> 
> I don't think it's the only reason that container build failed but surely using the proper package manager can't hurt here?

Even if I agree that we should use dnf command for CentOS 8, yum is still available as an dnf alias

```console
# ls -hl /usr/bin/yum /bin/yum 
lrwxrwxrwx 1 root root 5 May 13  2019 /bin/yum -> dnf-3
lrwxrwxrwx 1 root root 5 May 13  2019 /usr/bin/yum -> dnf-3
```

If we only change this two commands by dnf then we will still use yum at other places. If we don't want to have too much difference between CentOS 7 and 8, I think we can keep yum.
This will avoid code duplication.

PR #1561, opened 2019-12-30
ubi8: daemon-base: install without docs

mykaul:
In a container, there's no need to install documentation.
Pass the relevant option for yum to skip documentation installation

Signed-off-by: Yaniv Kaul <ykaul@redhat.com>

### DIFF
--- a/ceph-releases/ALL/ubi8/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/ubi8/daemon-base/__DOCKERFILE_INSTALL__
@@ -1,3 +1,3 @@
-yum update -y && \
-yum install -y wget unzip util-linux python3-setuptools udev device-mapper && \
-yum install -y __CEPH_BASE_PACKAGES__
+yum update --setopt=tstflags=nodocs -y && \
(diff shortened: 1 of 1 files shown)


mykaul:
CI failure is on:
"Error: Failed to synchronize cache for repo 'lab-extras'"
which is true - I can't reach http://apt-mirror.front.sepia.ceph.com/ either :(

So some infra issue.

dsavineau:
> CI failure is on:
> "Error: Failed to synchronize cache for repo 'lab-extras'"
> which is true - I can't reach http://apt-mirror.front.sepia.ceph.com/ either :(
> 
> So some infra issue.

this has been resolved

mykaul:
Asked the CentOS team about 8 - https://github.com/CentOS/sig-cloud-instance-build/issues/167

Of course it should be added to CentOS 8 - thought I'd start with UBI first though.

Not sure why it should be a configuration item, though.

dsavineau:
> Not sure why it should be a configuration item, though.

To avoid adding new yum/dnf commands without this option and because that's the way it's done on ubi 7.

PR #1564, opened 2020-01-15
build: only build stable releases

leseb:
Description of your changes:

Currently the regex builds every package, which leads to issue when not
building the stable minor '2' because some packages do not exist.
For example, we hit:

No package ceph-mgr-k8sevents-14.1.0 available.

It doesn't make sense to build non-stable packages because we have dev
builds.
So this will avoid packages issues as well as speeding up the build
time.

Checklist:

### DIFF
--- a/contrib/ceph-build-config.sh
+++ b/contrib/ceph-build-config.sh
@@ -229,14 +229,15 @@ function get_ceph_download_url () {
 }
 
 # Return a list of the ceph version strings available on the server as:
-#   "12.0.0 12.0.1 12.0.2 12.1.0 ..."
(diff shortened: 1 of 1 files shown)


dsavineau:
> Currently the regex builds every package, which leads to issue when not
> building the stable minor '2' because some packages do not exist.
> For example, we hit:
> 
> No package ceph-mgr-k8sevents-14.1.0 available.

I think we will have the same issue for stable releases too.
As you mentioned with ceph-mgr-k8sevents, this package is only available since 14.2.5 so rebuilding releases before that one would fail.

PR #1567, opened 2020-01-16
el8: add target-restore package

dsavineau:
This is a workaround for the python3-rtslib package that doesn't create
the /etc/target and /var/target directories compared to python-rtslib.
Installing target-restore package will create those without pulling a
lot a directories


dsavineau:
not needed anymore

PR #1578, opened 2020-01-28
Fix osd-volume sigterm-cleanup

j20d:
Description of your changes:
Made sure that `findmnt ... | grep`-construction always returns 0

`findmnt ... | grep` would return 1 when there's no mounts to be found, terminiating the sigterm_cleanup_post-function immediately, skipping the cryptsetup-close.


j20d:
So ... what should I have done to keep it from getting flagged as 'wontfix'?
The script is clearly broken and the fix rather trivial, if this is due to the CI-Test failing I'd suggest someone looks into why the pipeline was running out of resources when I submitted the PR.

j20d:
really? Closed without anyone actually taking a look?

PR #1582, opened 2020-02-02
Fix activating legacy osd-disk partitioned nvme device

### LINKED ISSUE #1253: Can't use NVMe disks as OSDs.
Is this a bug report or feature request?
* Bug Report

**Bug Report**

What happened:
When the daemon tries to mount an nvme mount, /dev/nvme0n1p1, it instead tries to mount /dev/nvme0n11.

What you expected to happen:
It mounted /dev/nvme0n1p1

How to reproduce it (minimal and precise):
Using ceph-helm with 
ceph/daemon:tag-stable-3.0-luminous-centos-7 on Kubernetes with an NVMe disk attached to the host.

**Environment**:
* OS (e.g. from /etc/os-release): Amazon Linux 2
* Kernel (e.g. `uname -a`): 4.14.72-73.55.amzn2.x86_64
* Docker version (e.g. `docker version`): 17.06.2-ce
* Ceph version (e.g. `ceph -v`) :12.2.4 (52085d5249a80c5f5121a76d6288429f35e4e77b) luminous (stable)

It looks like this issue lies in https://github.com/ceph/ceph-container/blob/master/src/daemon/common_functions.sh#L135, which doesn't seem to have a special case for nvme devices.

hex43ver:
Description of your changes:

Resolves #1253

Checklist:

### DIFF
--- a/src/daemon/osd_scenarios/osd_disk_activate.sh
+++ b/src/daemon/osd_scenarios/osd_disk_activate.sh
@@ -20,12 +20,14 @@ function osd_activate {
     exit 1
   fi
 
-  if ! test -d /etc/ceph/osd || ! grep -q ${OSD_DEVICE}1 /etc/ceph/osd/*.json; then
-    log "INFO: Scanning ${OSD_DEVICE}"
-    ceph-volume simple scan ${OSD_DEVICE}1
+  data_part=$(dev_part "${OSD_DEVICE}" 1)
+
+  if ! test -d /etc/ceph/osd || ! grep -q ${data_part} /etc/ceph/osd/*.json; then
+    log "INFO: Scanning ${data_part}"
+    ceph-volume simple scan ${data_part}
   fi
 
-  CEPH_VOLUME_SCAN_FILE=$(grep -l ${OSD_DEVICE}1 /etc/ceph/osd/*.json)
+  CEPH_VOLUME_SCAN_FILE=$(grep -l ${data_part} /etc/ceph/osd/*.json)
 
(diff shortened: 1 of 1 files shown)


src/daemon/osd_scenarios/osd_disk_activate.sh
```diff
@@ -21,8 +21,9 @@ function osd_activate {
   fi
 
   if ! test -d /etc/ceph/osd || ! grep -q ${OSD_DEVICE}1 /etc/ceph/osd/*.json; then
-    log "INFO: Scanning ${OSD_DEVICE}"
-    ceph-volume simple scan ${OSD_DEVICE}1
+    data_part=$(dev_part "${OSD_DEVICE}" 1)
+    log "INFO: Scanning ${data_part}"
+    ceph-volume simple scan ${data_part}
```
dsavineau: The data_part should be set before the condition and used with the grep command.

src/daemon/osd_scenarios/osd_disk_activate.sh
```diff
@@ -21,8 +21,9 @@ function osd_activate {
...
   fi
 
   CEPH_VOLUME_SCAN_FILE=$(grep -l ${OSD_DEVICE}1 /etc/ceph/osd/*.json)
```
dsavineau: The same partition is also used later

dsavineau:
it would also be nice to amend your commit and add a commit body to explain the reason of the change.

hex43ver:
Thanks, it has been fixed

PR #1585, opened 2020-02-05
daemon-base: add nfs-ganesha-rados-urls

mgfritch:
needed in `ceph/daemon-base:latest-master-devel` for nfs-ganesha
deployment by `bin/cephadm` (https://github.com/ceph/ceph/pull/33064)

Signed-off-by: Michael Fritch <mfritch@suse.com>

Checklist:

### DIFF
--- a/ceph-releases/ALL/centos/8/daemon-base/__GANESHA_PACKAGES__
+++ b/ceph-releases/ALL/centos/8/daemon-base/__GANESHA_PACKAGES__
@@ -1,0 +1,1 @@
+nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw nfs-ganesha-rados-grace nfs-ganesha-rados-urls


mgfritch:
this is the right path for the master release:  `ceph-releases/ALL/centos/8/daemon-base/__GANESHA_PACKAGES__` ?

dsavineau:
> this is the right path for the master release: `ceph-releases/ALL/centos/8/daemon-base/__GANESHA_PACKAGES__` ?

Even if this works because ceph@master is using CentOS 8, this should be a specific configuration per ceph release rather a per distro.

That said I'll merge this PR as is for cephadm purpose and I'll probably do some refactoring on this later.

mgfritch:
> > this is the right path for the master release: `ceph-releases/ALL/centos/8/daemon-base/__GANESHA_PACKAGES__` ?
> 
> Even if this works because ceph@master is using CentOS 8, this should be a specific configuration per ceph release rather a per distro.
> 
> That said I'll merge this PR as is for cephadm purpose and I'll probably do some refactoring on this later.

Yep, I was looking for an `octopus` related release, but I don't think one exists yet ...please, refactor however makes sense. Thanks!

PR #1600, opened 2020-03-05
ubi8: add gating.yaml

ktdreyer:
Our internal Greenwave instance will use these settings to determine overall pass/fail test results.


ktdreyer:
cc @XDsubhash since he is the original author of this commit.

ktdreyer:
It would be nice to cherry-pick this back to "3" as well.

dsavineau:
> It would be nice to cherry-pick this back to "3" as well.

We're not using ubi8 in RHCS 3 so should this be in rhel7 instead ?

ktdreyer:
Whoops, yes, you're right.

PR #1616, opened 2020-03-17
contrib: add octopus to ceph/ceph repository

dsavineau:
This commit adds the Ceph Octopus release to the ceph/ceph repository
but only for x86_64 architecture.
When the arm64 build will be available we will add it to the mix.
This also allows using RC build (X.1.Z) in addition of stable build
(X.2.Z).


dsavineau:
> Looks good but until we get the arm build we won't have the `v15` build.

@liewegas do you know how long we should wait to get the arm64 build ?

PR #1620, opened 2020-03-19
octopus: nfs-ganesha bits from download.ceph.com

dsavineau:
We now have nfs-ganesha 3.2 builds available on download.ceph.com so we
should use this source for the Ceph Octopus release.

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -4,13 +4,20 @@ bash -c ' \
   if [ -n "__GANESHA_PACKAGES__" ]; then \
     echo "[ganesha]" > /etc/yum.repos.d/ganesha.repo ; \
     echo "name=ganesha" >> /etc/yum.repos.d/ganesha.repo ; \
-    if [[ "${CEPH_VERSION}" =~ master|octopus|^wip* ]]; then \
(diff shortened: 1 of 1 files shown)


dsavineau:
@dmick is it was you were looking for ?

dmick:
I don't know.  Was the decision for which version of nfs-ganesha to use discussed by people who know which version of nfs-ganesha is appropriate for every release?  Is this as simple as we can make this code, and can we not make it so that new releases of ceph and nfs-ganesha can be assembled without hacking this code every time?  Can we write any of the relevant information about release matching down?  @mattbenjamin @dang

dsavineau:
> I don't know. Was the decision for which version of nfs-ganesha to use discussed by people who know which version of nfs-ganesha is appropriate for every release?

Not really, we're just following what was provided upstream on shaman. During the el8 migration we discussed about changing the source from shaman to download.nfs-ganesha.org with Sage and Kaleb.
Also we always try to align the release of external projects like ceph-iscsi and nfs-ganesha to downstream.

> Is this as simple as we can make this code, and can we not make it so that new releases of ceph and nfs-ganesha can be assembled without hacking this code every time?

If you have a better implementation about this then feel free to send a PR but considering that projects like ceph-iscsi and nfs-ganesha have their own releases then I don't see how we can have a generic code for every new ceph releases without doing the mapping between the ceph release and the ceph-iscsi/nfs-ganesha releases.

dsavineau:

dsavineau:

dsavineau:

dsavineau:

PR #1621, opened 2020-03-19
contrib: Aggressively clean up leftover podman artifacts

djgalloway:
In our CI, we're occasionally seeing https://github.com/containers/libpod/issues/5306

This PR is an attempt to fix those issues based on the suggestions in the comments.

Description of your changes:
- Cleanup leftover artifacts from previous builds

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -64,6 +64,13 @@ function _centos_release {
 
 function cleanup_previous_run {
   make clean.all || true
+  # https://github.com/containers/libpod/issues/5306#issuecomment-591661572
+  # https://github.com/containers/libpod/issues/5306#issuecomment-595737952
+  if ${CI_CONTAINER}; then
+    docker system migrate || true
+    rm -rf ~/.config/containers || true
+    rm -rf ~/.local/share/containers || true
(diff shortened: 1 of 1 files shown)


tchaikov:
as an alternative probably we can also install the latest 1.8.2 podman to address this issue. the prebuilt latest upstream releases can be found at https://podman.io/getting-started/installation#centos

djgalloway:
@dsavineau what do you recommend?

dsavineau:
Considering that this is an infra issue and not a ceph-container issue I would say that using a podman release that includes the fix should be the best solution.

That said, if this isn't possible we could do something here. The only thing about the change is that I don't really like to change the Makefile for this kind of solution.
A better approach could be to add the changes in [1] and based on a `CI_CONTAINER` flag condition.

[1] https://github.com/ceph/ceph-container/blob/master/contrib/build-push-ceph-container-imgs.sh#L65-L67

djgalloway:
Yep, I like the latter suggestion better.  I just wasn't sure it made sense to add bash to a function that just calls a function in the makefile.  I'll adjust accordingly.

dsavineau:
unrelated CI failure

djgalloway:
@liewegas FYI

PR #1636, opened 2020-03-27
start_mon.sh: fix mon bootstrap bug

alram:
The current code combines `--mkfs` and `--inject-monmap` when bootstrapping the mon which is not supported. 
We end up with a default monmap containing the mons from `mon_host` with names `noname-<x>`.This is problematic when using CNI with k8s because the ceph-mon service is resolvable only after the Pod is ready so the mons won't form a quorum.
This injects the monmap after running mkfs.

I'm thinking that this will break a few things down the line when multiple mons are bootstrapped since it's been there forever. I'll do more tests over the weekend, please don't merge yet.

I'm working in an environment without rook, but I believe this could fix https://github.com/rook/rook/issues/5065

### DIFF
--- a/src/daemon/start_mon.sh
+++ b/src/daemon/start_mon.sh
@@ -144,7 +144,8 @@ function start_mon {
     done
 
     # Prepare the monitor daemon's directory with the map and keyring
-    ceph-mon --setuser ceph --setgroup ceph --cluster "${CLUSTER}" --mkfs -i "${MON_NAME}" --inject-monmap "$MONMAP" --keyring "$MON_KEYRING" --mon-data "$MON_DATA_DIR"
+    ceph-mon --setuser ceph --setgroup ceph --cluster "${CLUSTER}" --mkfs -i "${MON_NAME}" --keyring "$MON_KEYRING" --mon-data "$MON_DATA_DIR"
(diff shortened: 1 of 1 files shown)


alram:
On second thoughts this might not be enough. It looks like this might require all the mons IP not just the current running container one. I'll have a better look tomorrow

dsavineau:
> but I believe this could fix [rook/rook#5065](https://github.com/rook/rook/issues/5065)

I don't think so because rook doesn't use the container image with the extra scripts (ceph/daemon) you're modifying but the container image with the ceph binaries only (ceph/ceph).

alram:
> > but I believe this could fix [rook/rook#5065](https://github.com/rook/rook/issues/5065)
> 
> I don't think so because rook doesn't use the container image with the extra scripts (ceph/daemon) you're modifying but the container image with the ceph binaries only (ceph/ceph).

Oh thanks. yeah, I just checked the code you're right. It's all rewritten as part of the operator

alram:
Since rook and ceph-container do not share the scripts, I've switched to rook for my testing env. so I'm going to close this.

The bug is only affecting k8s+ cni so it should not hit a bunch of people since Rook is the way forward for this setup.

PR #1642, opened 2020-04-01
contrib: Use newer version of podman when building CI containers

djgalloway:
- Revert attempts (dcdf268 and 8f7c363) to workaround a bug by cleaning up cached podman artifacts
- Use newer version of podman when building CI containers
- Use newer version of fuse-overlayfs when building CI containers

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -64,13 +64,6 @@ function _centos_release {
 
 function cleanup_previous_run {
   make clean.all || true
-  # https://github.com/containers/libpod/issues/5306#issuecomment-591661572
-  # https://github.com/containers/libpod/issues/5306#issuecomment-595737952
-  if ${CI_CONTAINER}; then
-    docker system migrate || true
-    sudo rm -rf ~/.config/containers || true
-    sudo rm -rf ~/.local/share/containers || true
-  fi
 }
 
 declare -F install_docker ||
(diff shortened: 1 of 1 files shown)


contrib/build-push-ceph-container-imgs.sh
```diff
@@ -91,6 +84,14 @@ function install_docker {
 function install_podman {
   # https://tracker.ceph.com/issues/44242
   sudo dnf localinstall -y http://apt-mirror.front.sepia.ceph.com/lab-extras/8/fuse-overlayfs-0.7.6-2.0.dev.gitf1da779.el8.x86_64.rpm
```
dsavineau, line 86: I guess we don't need to install this specific fuse-overlayfs anymore with the need repository as I can see 0.7.6-10.1 available.

djgalloway: Oh I didn't notice.  I'll remove that.  Thanks!

liewegas:
Unfortunately this didn't work.. see https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos8,DIST=centos8,MACHINE_SIZE=gigantic/38934//consoleFull

```
+ sudo dnf -y module disable container-tools
...
+ docker login -u **** -p **** quay.io
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Error: could not get runtime: open /proc/13238/ns/user: no such file or directory
```

liewegas:
Can we just do the entire container build as root?

dsavineau:
I don't see any reason why you shouldn't be able to run the build script as root

Issue #1579, opened 2020-01-30
Bug in common_functions.sh?


frans42:
In the image tagged "v3.2.10-stable-3.2-mimic-centos-7-x86_64" (mimic 13.2.8) in the script common_functions.sh:278-279 I believe the code block

```
  if [[ ${OSD_BLUESTORE} -eq 1 ]]; then
    dev_real_path=$(resolve_symlink "$OSD_BLUESTORE_BLOCK_WAL" "$OSD_BLUESTORE_BLOCK_DB")
    for partition in $(list_dev_partitions "$OSD_DEVICE" "$dev_real_path"); do
```

should be something like this:

```
  if [[ ${OSD_BLUESTORE} -eq 1 ]]; then
    dev_real_path=( $(resolve_symlink "$OSD_BLUESTORE_BLOCK_WAL" "$OSD_BLUESTORE_BLOCK_DB") )
    for partition in $(list_dev_partitions "$OSD_DEVICE" "${dev_real_path[@]}"); do
```

The expansion `"$dev_real_path"` seems not to provide the desired result.

dsavineau:
> I believe the code block should be something like this

Do you just believe or did you ran into a real error ?

frans42:
I run into an error a long time ago, when trying to deploy co-located bluestore devices where WAL and DB are actually co-located instead of separate partitions on the same disk as was deployed in older versions by default. This has been fixed recently here:

```
  if [[ ${OSD_BLUESTORE} -eq 1 ]]; then
    CEPH_DISK_CLI_OPTS+=(--bluestore)
    if [[ "${OSD_BLUESTORE_BLOCK_WAL}" != "${OSD_DEVICE}" ]]; then
      CEPH_DISK_CLI_OPTS+=(--block.wal "${OSD_BLUESTORE_BLOCK_WAL}" --block.wal-uuid "${OSD_BLUESTORE_BLOCK_WAL_UUID}")
    fi
    if [[ "${OSD_BLUESTORE_BLOCK_DB}" != "${OSD_DEVICE}" ]]; then
      CEPH_DISK_CLI_OPTS+=(--block.db "${OSD_BLUESTORE_BLOCK_DB}" --block.db-uuid "${OSD_BLUESTORE_BLOCK_DB_UUID}")
    fi
    ceph-disk -v prepare "${CEPH_DISK_CLI_OPTS[@]}" \
    --block-uuid "${OSD_BLUESTORE_BLOCK_UUID}" \
[...]

PR #1667, opened 2020-05-04
common_function: use dev_real_path as array

dsavineau:
When applying ceph owner/group for Bluestore OSD with dedicated DB
and/or WAL partitions then the blank spaces in the dev_real_path
variable will be preserve. The result will be a single string which
won't be evaluated correctly with the for loop.
    
Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1831273
Closes: #1579
    

dsavineau:
unrelated CI failure

PR #1669, opened 2020-05-05
ubi8: ignore absent pulp repositories

ktdreyer:
OSBS introduced a new "`ignore_absent_pulp_repos`" setting. This setting tells ODCS to proceed when it cannot find some of the expected repositories in pulp.

The reason we must enable this setting is that Red Hat release engineering has not created all the pulp repositories in `content_sets.yml` yet. We can revert this commit when all of those pulp repositories are ready.

### DIFF
--- a/ceph-releases/ALL/ubi8/daemon/container.yaml
+++ b/ceph-releases/ALL/ubi8/daemon/container.yaml
@@ -4,3 +4,4 @@
 compose:
   packages: []
   pulp_repos: true
+  ignore_absent_pulp_repos: true


ktdreyer:
@tserlin FYI

dsavineau:
@ktdreyer so this doesn't need to be backported to 3 and 4 branches ?

ktdreyer:
That's correct because "5" is the first branch that has `container.yaml`.

ktdreyer:
(Also, we can revert this once we have all the "5" pulp repos in place.)

PR #1678, opened 2020-05-21
ubi8: add s390x entry to content_sets.yml

ktdreyer:
We're building the RHCS 5 container for s390x. Add the relevant repositories to the `content_sets.yml` file that the [OSBS](https://osbs.readthedocs.io/) and [Freshmaker](https://pagure.io/freshmaker) systems read.


ktdreyer:
We'll need in `stable-5.0`, and we'll also need a version of this change for RHCS 4, but the Pulp content set names are different ("4" vs "5").

dsavineau:
unrelated CI failure

PR #1704, opened 2020-07-02
build-push-ceph-container-imgs.sh: change container tag

ideepika:
with teuthology supporting crimson flavor build differentiating
container on basis of only sha1 is not unique anymore.
This change will push container with tag format:

<>:sha1:os-os_version-shaman_osd_flavor

Signed-off-by: Deepika Upadhyay <dupadhya@redhat.com>

Description of your changes:

Resolves #

Checklist:

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -20,6 +20,7 @@ if ${CI_CONTAINER} ; then
   # must be set by caller (perhaps another Jenkins build job)
   # BRANCH (as above, the Ceph branch)
   # SHA1 (sha1 corresponding to the Ceph branch)
+  # OSD_FLAVOR (choose between default and crimson flavor)
   # CONTAINER_REPO_HOSTNAME="quay.io"
   # CONTAINER_REPO_ORGANIZATION="ceph-ci"
   # CONTAINER_REPO_USERNAME=user
@@ -39,7 +40,8 @@ CONTAINER_BRANCH="${GIT_BRANCH#*/}"
 CONTAINER_SHA=$(git rev-parse --short HEAD)
 TAGGED_HEAD=false # does HEAD is on a tag ?
 DEVEL=${DEVEL:=false}
-OSD_FLAVOR=${FLAVOR:="default"}
+# flavor based on OSD type proporgated by ceph-build
+OSD_FLAVOR=${OSD_FLAVOR:=default}
 
 if [ -z "$CEPH_RELEASES" ]; then
   # NEVER change 'master' position in the array, this will break the 'latest' tag
@@ -255,6 +257,11 @@ function push_ceph_imgs_latest {
     docker push $full_repo_tag
     docker push $branch_repo_tag
     docker push $sha1_repo_tag
+    if [[ "${OSD_FLAVOR}" == "crimson" ]]; then
+      sha1_flavor_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}-${OSD_FLAVOR}
+      docker tag $local_tag $sha1_flavor_repo_tag
+      docker push $sha1_flavor_repo_tag
+    fi
     return
   fi
 


contrib/build-push-ceph-container-imgs.sh
```diff
@@ -245,13 +250,13 @@ function push_ceph_imgs_latest {
     local_tag=${CONTAINER_REPO_ORGANIZATION}/daemon-base:${RELEASE}-${BRANCH}-centos-${CENTOS_RELEASE}-${HOST_ARCH}
     full_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${RELEASE}-centos-${CENTOS_RELEASE}-${HOST_ARCH}-devel
     branch_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${BRANCH}
-    sha1_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}
+    sha1_os_flavor_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}:${OS}-${CENTOS_RELEASE}:${OSD_FLAVOR}
```
dsavineau, line 253: using extra `:`  in the tag name isn't supported

ideepika: aah, I see. We wanted to identify a container based on shaman flavor(OSD_FLAVOR) since sha1 alone does not act as a unique identifier.
reference from irc:
```
otoh, would be great to have the OS there as well. evenutally I'd like to get leap containers as well. 
7:14 PM <kefu> Kefu Chai os_type-os_ver-flavor.
7:15 PM <SebastianW> Sebastian Wagner +1
7:37 PM <ideepika> Deepika Upadhyay kefu Sebastian: would it be sha1-os_type-os_ver-flavor as a whole, since we would want to a uniqueID for a particular branch? 
7:38 PM SebastianW
7:38 PM <SebastianW> Sebastian Wagner yes
7:38 PM <SebastianW> Sebastian Wagner we need to identify the branch. sha1 was the least problematic
7:38 PM <SebastianW> Sebastian Wagner otherwise you'll end up testing the wrong build. 
7:39 PM so yes
7:39 PM sha1-os-ver-flavor
```

dsavineau: so you can just add the OSD_FLAVOR to the tag but with a dash or underscore between both value (ie: not `:`)

ideepika:
https://github.com/ceph/ceph/pull/35913

contrib/build-push-ceph-container-imgs.sh
ideepika: @dsavineau would it be okay to use include `sha1-centos-8-flavor` tag as Sebastian mentioned in future they might want to use leap container as well.
Updated to dash from colon

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -248,13 +252,13 @@ function push_ceph_imgs_latest {
...
     branch_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${BRANCH}
-    sha1_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}
+    sha1_os_flavor_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}-${OS}-${CENTOS_RELEASE}-${OSD_FLAVOR}
     docker tag $local_tag $full_repo_tag
     docker tag $local_tag $branch_repo_tag
     docker tag $local_tag $sha1_repo_tag
     docker push $full_repo_tag
     docker push $branch_repo_tag
-    docker push $sha1_repo_tag
+    docker push $sha1_os_flavor_repo_tag
```
tchaikov, line 261: i don't think this works. you need to fix the previous command of 
```sh
docker tag $local_tag $sha1_repo_tag
```
first. namely to create a tag of `$sha1_os_flavor_repo_tag` referring the `$local_tag`. also, we are still overwriting the`$full_repo_tag` and `$branch_repo_tag` with the last flavor pushed to registry.

i am not sure who are the consumers of the other two tags, but if there are any of them. they will likely have random failures. as crimson-osd will be used in the place of ceph-osd, if the crimson flavored image is pushed later.

ideepika: aah, sorry for missing that. Even I am not sure about consumer for them but I don't think that will cause failure, as we only supply OSD_FLAVOR to push-container-img.sh script in Jenkins, and apart from that OSD_FLAVOR defaults to use `default` flavor, so I don't suspect it to fail.

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -19,7 +19,9 @@ if ${CI_CONTAINER} ; then
   CEPH_BRANCH=${BRANCH}
   # must be set by caller (perhaps another Jenkins build job)
   # BRANCH (as above, the Ceph branch)
+  # OS (choice of flavor of operating system: default: centos)
```
tchaikov, line 22:
```suggestion
  # OS (choice of operating system: default: centos)
```

ideepika: ack.

contrib/build-push-ceph-container-imgs.sh
dsavineau: > as Sebastian mentioned in future they might want to use leap container as well.

There's no reason to have multiple container images with the same Ceph version (ie: with different GNU/Linux base image). Container image delivers software (ceph in this case) not OS.

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -37,9 +39,11 @@ CONTAINER_REPO_PASSWORD=${CONTAINER_REPO_PASSWORD:-$DOCKER_HUB_PASSWORD}
 # GIT_BRANCH is typically 'origin/master', we strip the variable to only get 'master'
 CONTAINER_BRANCH="${GIT_BRANCH#*/}"
 CONTAINER_SHA=$(git rev-parse --short HEAD)
+OS=${OS:=centos}
```
dsavineau, line 42: Again, this isn't useful especially when we have variables named with CENTOS or function like `_centos_release`

contrib/build-push-ceph-container-imgs.sh
dsavineau: I'm not aware about the consumers, this has been design by @dmick for the ceph CI so it might have more information

contrib/build-push-ceph-container-imgs.sh
ideepika: since this script is move gravitated to make use of centos containers, I agree ... otherwise it would be a lot of work without any usecase for now. I will update my PR, thanks Dimitri!

ideepika:
related: https://github.com/ceph/ceph-build/pull/1588/commits

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -248,13 +250,13 @@ function push_ceph_imgs_latest {
     local_tag=${CONTAINER_REPO_ORGANIZATION}/daemon-base:${RELEASE}-${BRANCH}-centos-${CENTOS_RELEASE}-${HOST_ARCH}
     full_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${RELEASE}-centos-${CENTOS_RELEASE}-${HOST_ARCH}-devel
     branch_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${BRANCH}
-    sha1_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}
+    sha1_flavor_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}-${OSD_FLAVOR}
     docker tag $local_tag $full_repo_tag
     docker tag $local_tag $branch_repo_tag
-    docker tag $local_tag $sha1_repo_tag
+    docker tag $local_tag $sha1_flavor_repo_tag
     docker push $full_repo_tag
```
ideepika, line 257: @dmick can you help verify the use-case for full_repo_tag + branch_repo_tag or are they obsolete now?

dmick: They were always for convenience so a developer can configure his branch or a specific sha1.  I don't see anything that would make them obsolete.

tchaikov: @dmick can we add a flavor postfix to their tag names?  just to differentiate the default flavor build from the crimson flavored build? our concern is that it might break some users of these images, if we change the naming scheme of them.

dmick: It would probably be safer to add a few new tags that include crimson.  I don't know for sure who is using what tags for what reason, but adding the capability to select crimson specifically seems safest.

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -39,7 +40,8 @@ CONTAINER_BRANCH="${GIT_BRANCH#*/}"
 CONTAINER_SHA=$(git rev-parse --short HEAD)
 TAGGED_HEAD=false # does HEAD is on a tag ?
 DEVEL=${DEVEL:=false}
-OSD_FLAVOR=${FLAVOR:="default"}
+#shaman flavor based on OSD type proporgated by ceph-build
```
tchaikov, line 43: please note, shaman is just a RESTful API and web UI for accessing the builds and repos. the "flavor" is connected more closely to ceph-build and jenkins.

also, please add a space after `#`.

ideepika: aah, I see. amended PR :)

tchaikov:
@dsavineau do you want to take a final look before merging this change? some tests are failing, but i don't think they are related.

tchaikov:
jenkins retest this please.

PR #1709, opened 2020-07-09
DNM: release: Bump ceph-release RPM version

djgalloway:
Backstory: chacra.ceph.com got erased inadvertently (https://lists.ceph.io/hyperkitty/list/dev@ceph.io/thread/YQMAHTB7MUHL25QP7V5ZUJQSTOGY4GHX/).

The packages on download.ceph.com come from chacra.ceph.com, get signed, then pushed to download.ceph.com.

Since chacra.ceph.com got erased, I've had to rebuild the ceph-release RPM for each Ceph version.

The new ceph-release RPM version is 1-1.  See https://github.com/ceph/ceph-build/blob/master/ceph-release-rpm/build/build#L117-L125

container builds are currently failing because ceph-release-1-0.elX.noarch.rpm is missing.  e.g.,
```
[91merror: open of null/noarch/ceph-release-1-0.el8.noarch.rpm failed: No such file or directory
```

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -45,7 +45,7 @@ fi && \
 bash -c ' \
   if [[ "${CEPH_VERSION}" =~ master|^wip* ]] || ${CEPH_DEVEL}; then \
     REPO_URL=$(curl -s "https://shaman.ceph.com/api/search/?project=ceph&distros=centos/__ENV_[BASEOS_TAG]__&flavor=${OSD_FLAVOR}&ref=${CEPH_VERSION}&sha1=latest" | jq -a ".[0] | .url"); \
-    RELEASE_VER=0 ;\
+    RELEASE_VER=1 ;\
     if [[ "${OSD_FLAVOR}" == "crimson" ]]; then \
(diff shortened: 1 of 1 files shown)


dsavineau:
Looks like we still have 1-0 package version on chachra isn't it ?

I dont' remember exactly but 1-0 was for unreleased content and 1-1 for stable/release content right ?

djgalloway:
> Looks like we still have 1-0 package version on chachra isn't it ?

Yeah, 1-0 does still get created I think.  I misunderstood the root cause of this issue.

> I dont' remember exactly but 1-0 was for unreleased content and 1-1 for stable/release content right ?

No idea.

Regardless, I think this is the appropriate fix: https://github.com/ceph/ceph-container/pull/1706

What we're seeing is [this](https://github.com/ceph/ceph-container/blob/master/contrib/build-push-ceph-container-imgs.sh#L60-L67) function assumes if it's a wip branch that we should build on CentOS8.  That's not always true, however.  For example, Yuri has a branch called `wip-yuri5-testing-2020-07-08-2015-luminous`.  It's luminous-based so we should be creating CentOS7 containers.  So `REPO_URL` [here](https://github.com/ceph/ceph-container/blob/master/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__#L47) gets set to `null` because the dockerfile wrongly assumes we're on CentOS8.

PR #1729, opened 2020-09-22
DAEMON_OPTS: use --default-* for logging opts

clwluvw:
Description of your changes:
Using --default-* in DAEMON_OPTS to allow ceph.conf to override this value.

Signed-off-by: Seena Fallah <seenafallah@gmail.com>

Checklist:

### DIFF
--- a/src/daemon/start_mon.sh
+++ b/src/daemon/start_mon.sh
@@ -194,7 +194,7 @@ function start_mon {
     # enable cluster/audit/mon logs on the same stream
     # Mind the extra space after 'debug'
     # DO NOT TOUCH IT, IT MUST BE PRESENT
-    DAEMON_OPTS+=(--mon-cluster-log-to-stderr "--log-stderr-prefix=debug ")
+    DAEMON_OPTS+=("--default-mon-cluster-log-to-stderr=true" "--default-log-stderr-prefix=debug ")
     if [[ ! "${CEPH_VERSION}" =~ ^(luminous|mimic)$ ]]; then
       DAEMON_OPTS+=("--default-mon-cluster-log-to-file=false")
     fi
--- a/src/daemon/variables_entrypoint.sh
+++ b/src/daemon/variables_entrypoint.sh
@@ -80,7 +80,7 @@ CRUSH_LOCATION_DEFAULT=("root=default" "host=${HOSTNAME}")
 CLI_OPTS=(--cluster ${CLUSTER})
 
 # This is ONLY used for the daemon's startup, e.g: ceph-osd $DAEMON_OPTS
-DAEMON_OPTS=(--cluster ${CLUSTER} --setuser ceph --setgroup ceph --log-to-stderr=true --err-to-stderr=true --default-log-to-file=false)
+DAEMON_OPTS=(--cluster ${CLUSTER} --setuser ceph --setgroup ceph --default-log-to-stderr=true --err-to-stderr=true --default-log-to-file=false)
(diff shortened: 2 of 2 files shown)


clwluvw:

clwluvw:
Ping @dsavineau

clwluvw:
@dsavineau @guits Can you please take a look at it?

dsavineau:
Even if the change won't break anything I'm wondering in which situation we would like to disable the stderr log ?

That would mean, podman and docker won't be able to get any logs from the running container.

clwluvw:
> Even if the change won't break anything I'm wondering in which situation we would like to disable the stderr log ?

> That would mean, podman and docker won't be able to get any logs from the running container.

clwluvw:
> Even if the change won't break anything I'm wondering in which situation we would like to disable the stderr log ?

@dsavineau When turning `log_to_file` true there will be double logging! One in container and one in `/var/log/ceph`! It has performance regression and also needs bigger space for fs to save these logs!
> That would mean, podman and docker won't be able to get any logs from the running container.

Does it has any benefits? I think it's like `log-to-syslog` in systemd deployment that it makes sense to turn it off or don't have any logs on it :thinking:

dsavineau:
> Does it has any benefits?

One can prefer to have logs via journald because the containers are managed by systemd.

I think your patch doesn't cover the `mon_cluster_log_to_stderr` option because if you want log to file only then you need to disable both options.

```console
# grep stderr /etc/ceph/ceph.conf 
log to stderr = false
mon cluster log to stderr = false
# ceph --admin-daemon /var/run/ceph/ceph-mon.foo.asok config get log_to_stderr       
{
    "log_to_stderr": "false"
}
# ceph --admin-daemon /var/run/ceph/ceph-mon.foo.asok config get mon_cluster_log_to_stderr
{
    "mon_cluster_log_to_stderr": "true"
}
```

https://github.com/ceph/ceph-container/blob/master/src/daemon/start_mon.sh#L197

clwluvw:
> > Does it has any benefits?
> 
> One can prefer to have logs via journald because the containers are managed by systemd.
> 
> I think your patch doesn't cover the `mon_cluster_log_to_stderr` option because if you want log to file only then you need to disable both options.

@dsavineau this config can be configure via `ceph.conf` because it isn't specified as an argument in daemon ops but `log-to-stderr` is specified as an argument in daemon ops so `ceph.conf` can't override it!

dsavineau:
> > > Does it has any benefits?
> > 
> > 
> > One can prefer to have logs via journald because the containers are managed by systemd.
> > I think your patch doesn't cover the `mon_cluster_log_to_stderr` option because if you want log to file only then you need to disable both options.
> 
> @dsavineau this config can be configured via `ceph.conf` because it isn't specified as an argument in daemon opts but `log-to-stderr` is specified as an argument in daemon opts so `ceph.conf` can't override it!

Please read twice my comment.

clwluvw:
@dsavineau Sorry I miss the link you shared and I though it was a default ceph config! :sweat_smile: 
I have changed the mon variables too.

clwluvw:
@dsavineau I have also checked cephadm for these variables and it uses these `--defaults-*` for it's logging opts.

dsavineau (approved):

clwluvw:
@dsavineau Would you please add backport label for `stable-4.0` and `stable-5.0`?

guits (approved):

PR #1730, opened 2020-09-24
tests/tox: don't use sudo with make

dsavineau:
If we use sudo with make then we will generate some files with root:root
owner/group which can't be deleted by jenkins during the workspace clean
up.
Because we already allow the current user to use the docker socket then
there's no need to use sudo with docker commands too.


dsavineau:

PR #1732, opened 2020-10-01
tests/tox: don't use sudo with make (bp #1730)

dsavineau:
If we use sudo with make then we will generate some files with root:root
owner/group which can't be deleted by jenkins during the workspace clean
up.
Because we already allow the current user to use the docker socket then
there's no need to use sudo with docker commands too.

Backport: #1730

(cherry picked from commit c945b6c3d040f8a2a4e032840edd362d69fc1dd9)


dsavineau:

dsavineau:
jenkins test sloppy python

dsavineau:
jenkins flake8

dsavineau:

PR #1745, opened 2020-10-28
centos/8: add ktdreyer copr repository back

dsavineau:
Since smarmontools 7 won't be present in CentOS 8 until 8.3 then we will
be stuck with 6.6 which isn't compatible with the device health status.
This re-adds ktdreyer copr repository that contains a smartmontools 7
build.
This is a temporary solution until CentOS 8.3 release.

Fixes: https://tracker.ceph.com/issues/47107


yaarith:
Thanks, @dsavineau!
I was just about to open a PR with the exact change, but had obvious conflicts to merge it...

PR #1753, opened 2020-11-12
contrib: do not overwrite images created for el8

tchaikov:
before this change, we build octopus container images for both el7 and
el8, and tag them using the same set of tag names. so the last build
wins in the racing. the consumers of these images still expect the el8
builds by asking the images by names encoded using sha1.

after this change, the el7 container images are not tagged with names
without distro_release encoded, so the el7 images does not overwrite
the el8 ones even if they are built and pushed after el8 ones with
the same sha1 and branch names.

Fixes: https://tracker.ceph.com/issues/48162
Signed-off-by: Kefu Chai <tchaikov@gmail.com>

Description of your changes:

Resolves #

Checklist:

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -281,6 +281,9 @@ function push_ceph_imgs_latest {
       sha1_flavor_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}-${OSD_FLAVOR}
       docker tag "$local_tag" "$sha1_flavor_repo_tag"
       docker push "$sha1_flavor_repo_tag"
+    elif [[ "${distro_release}" == "7" ]]; then
+      docker tag "$local_tag" "$full_repo_tag"
+      docker push "$full_repo_tag"
     else
       docker tag "$local_tag" "$full_repo_tag"
       docker tag "$local_tag" "$branch_repo_tag"


dmick:
I don't think testing for octopus is helpful, and since it's not minimal, could be confusing to future readers.  The ceph release/container relationship is controlled by ceph-build and Jenkins.

tchaikov:
@dmick updated and repushed.

badone (approved):
This looks OK to me, thanks Kefu. We can still get to the el7 container using an alternative tag and maybe there's some change required teuthology if we do want to test el7 containers. I know that is separate to this change but I'd feel more comfortable if we had more eyes pass over this change. Having said that it gets a 'Yes' from me.

dmick (approved):
Lgtm.  The check failures seem to be docker.io rate limiting.  @djgalloway is working on a local repo mirror.

jdurgin:
as discussed this morning, we don't need el7 containers for octopus - the intent of the el7 octopus build was to provide an upgrade path from nautilus for bare metal. It gives some overlap before users need to upgrade to el8 for pacific.

The inside of the container doesn't come into play there - it should always be el8-based for octopus and pacific.

dsavineau:
> Lgtm. The check failures seem to be docker.io rate limiting. @djgalloway is working on a local repo mirror.

We shouldn't need of a local repo mirror with #1754

dsavineau (approved):

dsavineau:

dsavineau:

dsavineau:

dmick:
I saw #1754 @dsavineau, also a good idea.  I think keeping as much container traffic local as is practical is still a good move, though (assuming the mirroring doesn't introduce issues).  Good to know we have options.

PR #1754, opened 2020-11-12
tests/tox: consume docker registry credentials

dsavineau:
To avoid docker hub rate limit on PRs and nightlies jobs.


djgalloway:
Are these paid credentials?

dsavineau:
> Are these paid credentials? Like it's a paid dockerhub account?

No we're using a "standard" user account. There's still rate limit with the authencitated user but this isn't done at IP address level.

Also we're using access token instead of using user/pass credentials.

PR #1795, opened 2021-01-22
release: add pacific target

dsavineau:
Since pacific branch has been created then we can split pacific and
master.
There's still no pacific build on download.ceph.com so this target is
still using shaman/chacra bits for now. Same thing for nfs-ganesha.
However we can use the stable ceph-iscsi 3 repository instead of
shaman/chacra.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@
 FLAVORS ?= \
 	nautilus,centos,7 \
 	octopus,centos,8 \
+	pacific,centos,8 \
 	master,centos,8
 
 TAG_REGISTRY ?= ceph
@@ -55,6 +56,7 @@ ALL_BUILDABLE_FLAVORS := \
 	nautilus,centos,8 \
 	octopus,centos,7 \
 	octopus,centos,8 \
+	pacific,centos,8 \
 	master,centos,8
 
 # ==============================================================================
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -4,7 +4,7 @@ bash -c ' \
   if [ -n "__GANESHA_PACKAGES__" ]; then \
     echo "[ganesha]" > /etc/yum.repos.d/ganesha.repo ; \
     echo "name=ganesha" >> /etc/yum.repos.d/ganesha.repo ; \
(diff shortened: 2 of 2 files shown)


dsavineau:
@liewegas @dmick I would be nice to know what pacific should use in term of ganesha and iscsi release.

I suppose that ceph-iscsi release will be the same than octopus (3.4) because this didn't changed a lot.

But for ganesha, are we going to add pacific to the current V3.3 repository on download.ceph.com [1] or do we plan to have V3.4 created for pacific on download.ceph.com ?
Note that the current master image is already using 3.4.

[1] https://download.ceph.com/nfs-ganesha/rpm-V3.3-stable/

dmick:
> it would be nice to know what pacific should use in term of ganesha and iscsi release

agreed.  The last time I asked around I couldn't even get anyone to tell me who should know, much less answer the question.  It's one of the most depressing experiences I've ever had with the team.

jdurgin:
nfs-ganesha sounds like a question for @mattbenjamin 
iscsi for @dillaman

jdurgin (approved):
let's merge this so we can get container builds and pacific tests going for now, further iscsi/nfs details can be handled in another PR

dillaman:
I want a point release of upstream ceph-iscsi / tcmu-runner prior to Pacific 16.2.0 -- but just not quite yet.

dsavineau:
> I want a point release of upstream ceph-iscsi / tcmu-runner prior to Pacific 16.2.0 -- but just not quite yet.

If it's a 3.x point release (like 3.5) then we are fine.

dsavineau:
The next question is : do we need latest-pacific-devel tag on ceph/daemon-base for cephadm on the pacific branch ?

This will be a temporary solution until we get 16.1.x releases so we will be able to build ceph/ceph:v16

jdurgin:
yes, we'll need latest-pacific-devel for testing on the pacific branch anyway

jdurgin:
even after an RC and release, that is

dsavineau:
FYI docker.io/ceph/daemon-base:latest-pacific-devel is now available

dmick:
@jdurgin: what uses the latest-pacific-devel tag?  All the CI work uses ceph-ci/ceph tags (branch, SHA1, and branch-shortSHA1-os-version-arch-devel

dsavineau:
@dmick I think that's mainly for cephadm [1]

[1] https://github.com/ceph/ceph/blob/pacific/src/cephadm/cephadm#L3

PR #1804, opened 2021-02-03
contrib/build-push-ceph-container-imgs.sh: solve aarch64 pkg conflict

dmick:
installing podman-docker leads to:

 Problem: cannot install both podman-2.2.1-1.el8.aarch64 and podman-2.0.5-5.module_el8.3.0+512+b3b58dca.aarch64
  - package podman-catatonit-2.0.5-5.module_el8.3.0+512+b3b58dca.aarch64 requires podman = 2.0.5-5.module_el8.3.0+512+b3b58dca, but none of the providers can be installed

Adding --allowerasing as dnf suggests allows the dnf install to complete.

Signed-off-by: Dan Mick <dmick@redhat.com>

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -103,7 +103,11 @@ function install_podman {
     # We used to provide fuse-overlayfs-0.7.6-2.0 in lab-extras but a newer version is available in the kubic repo so we'll install/update from there
     sudo dnf install -y fuse-overlayfs
   fi
-  sudo dnf install -y podman podman-docker
+  ALLOWERASING=""
+  if [[ "$HOST_ARCH" == "aarch64" ]] ; then 
+    ALLOWERASING="--allowerasing"
+  fi
+  sudo dnf install ${ALLOWERASING} -y podman podman-docker
 }
 
 function login_docker_hub {


djgalloway (approved):

guits:

guits (approved):

guits:

guits:

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -103,7 +103,11 @@ function install_podman {
     # We used to provide fuse-overlayfs-0.7.6-2.0 in lab-extras but a newer version is available in the kubic repo so we'll install/update from there
     sudo dnf install -y fuse-overlayfs
   fi
-  sudo dnf install -y podman podman-docker
+  ALLOWERASING=""
+  if [[ "$HOST_ARCH" == "aarch64" ]] ; then 
+    ALLOWERASING="--allowerasing"
+  fi
```
dsavineau, line 109: I'm not sure to understand why this is needed only for `aarch64` and not `x86_64`.

It looks like the `podman` package is coming from the suse repository (2.2) which doesn't provide the `podman-docker` package and generates the failure mentioned. IMHO this should fail on both archs.

Do you still need to suse repository since centos container-tools module is now providing podman 2.0 and fixes most (or all) of the past issues ?

dmick: currently investigating deeper now.  I don't understand either.  So far it's led to knowing far too much about the new module organization in dnf packaging and some really odd UI behavior therein.

This is currently marked "Draft" and is aarch64 only because that's the place it was failing; I haven't diagnosed why it doesn't affect x86_64.  At this moment I'm attempting to build a test with all that hackery removed (since the default versions will be 2.0.5).

dmick: ooh, yay for moving targets.  the kubic repo now contains podman-3.0.0 for both aarch64 and x86_64 (as of today)

dmick: Building without the kubic repo finished successfully.  Whether the container is any good or not, I don't yet know, but I have to solve the issue of tag collision now (for the 'branch' and 'sha1' "shorter" tags, the image is going to need to be a 'fat manifest' image that refers to both base architecture images rather than the second arch overwriting the first one).  So I'll close this PR in favor of another than 1) undoes the kubic hack and 2) manages the manifests properly.

dmick:
restating here, as more appropriate: 

Building without the kubic repo finished successfully. Whether the container is any good or not, I don't yet know, but I have to solve the issue of tag collision now (for the 'branch' and 'sha1' "shorter" tags, the image is going to need to be a 'fat manifest' image that refers to both base architecture images rather than the second arch overwriting the first one). So I'll close this PR in favor of another than 1) undoes the kubic hack and 2) manages the manifests properly.

PR #1814, opened 2021-03-03
ceph-releases/ALL/centos: pip install scikit-learn

liewegas:
The mgr failure prediction module needs this, and there is no package
for centos or EPEL (or plans to produce one).

Signed-off-by: Sage Weil <sage@newdream.net>

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -59,4 +59,5 @@ bash -c ' \
   fi && \
   rpm -Uvh "$REPO_URL/noarch/ceph-release-1-${RELEASE_VER}.el__ENV_[BASEOS_TAG]__.noarch.rpm" ' && \
 yum install -y __GPERFTOOLS_LIBS__ && \
-yum install -y __CEPH_BASE_PACKAGES__
+yum install -y __CEPH_BASE_PACKAGES__ && \
+python3 -m pip install -U scikit-learn==0.19.2


ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
```diff
@@ -58,4 +58,5 @@ bash -c ' \
     REPO_URL="http://download.ceph.com/rpm-${CEPH_VERSION}/el__ENV_[BASEOS_TAG]__/"; \
   fi && \
   rpm -Uvh "$REPO_URL/noarch/ceph-release-1-${RELEASE_VER}.el__ENV_[BASEOS_TAG]__.noarch.rpm" ' && \
-yum install -y __CEPH_BASE_PACKAGES__
+yum install -y __CEPH_BASE_PACKAGES__ && \
+python3 -m pip install -U scikit-learn
```
dsavineau, line 62:
```suggestion
python3 -m pip install -U scikit-learn==0.19.2
```

To keep the same requirement than the one from the ceph sources.

https://github.com/ceph/ceph/blob/master/src/pybind/mgr/diskprediction_local/requirements.txt#L3

dsavineau (changes requested):
> and there is no package
for centos or EPEL (or plans to produce one).

I thought [1] was supposed to track this.

Also what are the releases that required this change ?

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1844636

yaarith:
> > and there is no package
> > for centos or EPEL (or plans to produce one).
> 
> I thought [1] was supposed to track this.
> 
> Also what are the releases that required this change ?

@liewegas pacific and octopus?

> [1] https://bugzilla.redhat.com/show_bug.cgi?id=1844636

Yes, thanks for pointing this out. It's not clear when this will happen (or if), so this is a workaround for containerized deployments.

liewegas:
> > and there is no package
> > for centos or EPEL (or plans to produce one).
> 
> I thought [1] was supposed to track this.

We have been unsuccessful in finding anyone willing to maintain the package.  Nobody sees any real need for an RPM.  AFAICT pip installing it in our container more closely aligns with what everyone else is doing...

> Also what are the releases that required this change ?
> 
> [1] https://bugzilla.redhat.com/show_bug.cgi?id=1844636

Just pacific is fine; octopus + pacific is better. Adding it to older releases is not useful.  Does this mean we need to override the __DOCKERFILE_INSTALL__  in all of the older release directories?

ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
liewegas: done!

tchaikov:
does a package (something like python3-scikit-learn hosted by copr help? if yes, I can help cook one.

yaarith:
@tchaikov sure, this would help! This way we can use it in the spec file directly.

I already built packages for centos 8, which worked fine, but did not have keys to sign them:
https://bugzilla.redhat.com/show_bug.cgi?id=1844636#c7

When I tried to rebuild the package after ceontos 8.3 was out I saw a few errors:
https://bugzilla.redhat.com/show_bug.cgi?id=1844636#c9

tchaikov:
#1821

guits:
@liewegas @tchaikov should we close this one in favor of #1821 ?

tchaikov:
@guits i think so. but i'd defer this to @yaarith and @liewegas .

liewegas:
yep!

PR #1821, opened 2021-03-11
daemon-base: add copr repo for python3-scikit-learn packages

tchaikov:
the mgr failure prediction module needs this, and there is no package
packaged for CentOS8 or EPEL8.

see also https://bugzilla.redhat.com/show_bug.cgi?id=1844636

Signed-off-by: Kefu Chai <kchai@redhat.com>

Description of your changes:

Resolves #

Checklist:

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -57,6 +57,11 @@ bash -c ' \
     RELEASE_VER=1 ;\
     REPO_URL="http://download.ceph.com/rpm-${CEPH_VERSION}/el__ENV_[BASEOS_TAG]__/"; \
   fi && \
-  rpm -Uvh "$REPO_URL/noarch/ceph-release-1-${RELEASE_VER}.el__ENV_[BASEOS_TAG]__.noarch.rpm" ' && \
+  rpm -Uvh "$REPO_URL/noarch/ceph-release-1-${RELEASE_VER}.el__ENV_[BASEOS_TAG]__.noarch.rpm" && \
+  if [[ __ENV_[BASEOS_TAG]__ -eq 8 ]]; then \
+    yum install -y dnf-plugins-core ; \
+    yum copr enable -y tchaikov/python-scikit-learn ; \
+    yum install -y python3-scikit-learn ; \
+  fi ' && \
 yum install -y __GPERFTOOLS_LIBS__ && \
 yum install -y __CEPH_BASE_PACKAGES__


yaarith:
@tchaikov awesome, thanks!

I'm a bit confused, I see we install `dnf-plugins-core` with:
```
yum install -y dnf-plugins-core ; \
```
then enable the repo with:
```
yum copr enable -y tchaikov/python-scikit-learn ; \
```
but I don't see where the installation of the sklearn package itself is; not sure if I miss anything.

ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
```diff
@@ -49,6 +49,8 @@ bash -c ' \
       ARCH="arm64"; \
     fi ; \
     REPO_URL=$(curl -s "https://shaman.ceph.com/api/search/?project=ceph&distros=centos/__ENV_[BASEOS_TAG]__/${ARCH}&flavor=${OSD_FLAVOR}&ref=${CEPH_REF}&sha1=latest" | jq -r .[0].url); \
+    yum install -y dnf-plugins-core ; \
+    yum copr enable -y tchaikov/python-scikit-learn ; \
```
liewegas, line 53: does this actually get the package installed?

dsavineau: No this only enables the copr repository so we need an explicit `yum install -y python3-scikit-learn` command as well

tchaikov: no, it does not. i assumed that the package list was located somewhere else.

dsavineau (changes requested):
We need to handle this for CentOS 8 only otherwise all CentOS 7 build (like Nautilus) will fail.

You can replicate what we did with ktdreyer's copr repository [1] for smartmontools package.

[1] https://github.com/ceph/ceph-container/commit/aa644a5fd03aecb9eda8b95002d25ded9c57b25e

ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
tchaikov: @liewegas @dsavineau @yaarith  fixed and repushed.

tchaikov:
@dsavineau updated as suggested.

dsavineau (changes requested):
sorry I didn't pay attention to this earlier but I think this block should be outside the `CEPH_VERSION =~ master|pacific` condition as we want this for all releases based on CentOS 8 like Octopus (if I'm right).

tchaikov:
@dsavineau thanks. could you take another look?

dsavineau (approved):

PR #1829, opened 2021-03-15
centos/pacific: use ganesha from download.ceph.com

dsavineau:
We can now use stable nfs-ganesha release (3.5) from download.ceph.com
instead of CentOS buildlogs.


dsavineau:

PR #1846, opened 2021-03-23
[skip ci] centos-arm64/8: enable ceph-iscsi packages

dsavineau:
Since we have tcmu-runner package available on shaman for arm64 which was
the last piece missing to be able to install ceph-iscsi, then we can
enable the ceph-iscsi installation on the arm64 architecture as well.
This is done only on CentOS 8 distribution.


dsavineau:
This PR needs #1859 first

PR #1851, opened 2021-03-24
[skip ci] daemon-base/centos: simplify tcmu shaman repo code

dsavineau:
Instead of doing two curl calls on shaman for retrieving the tcmu-runner
repository information then we can do it in a single one via the /latest
endpoint and the arch parameter (instead of the search API).
This also makes sure we're using the -o curl option to store the repository
information on disk instead of using output redirection.


dmick:
I wonder if we need to handle the case that the .repo file already exists

djgalloway:
> I wonder if we need to handle the case that the .repo file already exists

I quickly just tested and `-o` overwrites existing files.

dmick:
I was more concerned about root vs non-root, but @djgalloway points out it's run inside the container, so that doesn't matter.

dmick (approved):

djgalloway (approved):

PR #1885, opened 2021-04-20
[skip ci] contrib: restart docker if start fails

dsavineau:
58d7bdd only managed the change in the script for building the
daemon-base/daemon container images but not the ceph one.


dsavineau:
FYI @djgalloway

PR #1889, opened 2021-04-21
contrib: remove ceph-nano dependency

dsavineau:
We don't really need to use ceph nano for listing the container tags
from docker hub registry.
We can do it in bash and we already do that in other scripts.
Also ceph nano is not optimized and something the command returns
nothing without failing.


dsavineau:
jenkins lint

PR #1913, opened 2021-07-22
contrib: fix centos 7 origin

dsavineau:
5b5e7f9 changed the source registry for the CentOS base container images.

Unfortunately the CentOS 7 image on quay.io/centos/centos registry is
amd64 only and not a multiarch manifest like docker.io/centos (with five
different architectures).
This sucks and as a solution we will stick to docker.io registry for the
CentOS 7 container image.


dsavineau:
jenkins lint

PR #1916, opened 2021-07-27
[skip ci] workflow: add timeout on ceph command

dsavineau:
In case of ceph monitor failure but the container is still up then
the ceph command will timeout only after 300s.
Adding --connect-timeout parameter to have the job in failure earlier.


dsavineau:
the CI failure is unrelated to this change and tracked by https://tracker.ceph.com/issues/51816

PR #1930, opened 2021-08-11
Revert "contrib: fix centos 7 origin"

dsavineau:
This reverts commit 7cc3401351dd2762312f36cefb2e47451c0695f0.

Because we don't built the Nautilus container image anymore and all
other Ceph releases are based on CentOS 8 then we can revert this change.


dsavineau:
jenkins lint

PR #1933, opened 2021-08-23
[skip ci] tox/tests: use base image from quay.io

dsavineau:
Avoid as much as possible to use the docker.io registry so even the centos
base container image should be pulled from the quay.io registry.


dsavineau:

PR #1940, opened 2021-09-02
contrib/rhcs: set upstream branch explicitly

dsavineau:
Since the upstream branch doesn't follow the downstream RHCS branch
then we can set it statically.


dsavineau:
jenkins lint

PR #1950, opened 2021-09-17
ubi8: rebase on daemon-base image

dsavineau:
Starting RHCS 6, ceph-ansible won't be supported anymore so we don't
need the RHCS container image to be based on the daemon image (binaries
and shell scripts).
Instead we can rebase on the daemon-base image.
For now updating the ceph release to master. This will need to be updated
to quincy in the futur.


dsavineau:
jenkins lint

PR #1960, opened 2021-11-04
daemon: remove forego usage

dsavineau:
This removes the forego installation and osd scenarios using it since
it isn't maintained and used anymore.


ceph-releases/ALL/centos/daemon/__DOCKERFILE_PREINSTALL__
```diff
@@ -1 +0,0 @@
-COPY --from=builder /go/bin/forego /usr/local/bin/forego
```
djgalloway, line 1: Does this file need to exist but should be empty?

src/daemon/common_functions.sh
```diff
@@ -141,15 +141,11 @@ function dev_part {
...
-    log "Bootstrapped OSD(s) found; using OSD directory"
-    source /opt/ceph-container/bin/osd_directory.sh
-    osd_directory
-  elif parted --script "${OSD_DEVICE}" print | grep -sqE '^ 1.*ceph data'; then
+  if parted --script "${OSD_DEVICE}" print | grep -sqE '^ 1.*ceph data'; then
     log "Bootstrapped OSD found; activating ${OSD_DEVICE}"
     source /opt/ceph-container/bin/osd_disk_activate.sh
     osd_activate
-  else
+  elif [[ ${OSD_DEVICE} != "none" ]]; then
```
djgalloway, line 148: Why not `else`?

dsavineau: because we don't want to trigger the osd_disk function if `${OSD_DEVICE} == "none"`

ceph-releases/ALL/centos/daemon/__DOCKERFILE_PREINSTALL__
dsavineau: yes, it is included in the base dockerfile

see https://github.com/ceph/ceph-container/blob/master/src/daemon/Dockerfile#L16

src/daemon/common_functions.sh
djgalloway: Should there maybe be an `else` error scenario then?  I'm not familiar enough with these scenarios to know if these are the only two options and no other (for lack of a better word) scenario could exist.

djgalloway: I guess in other words, is it better than nothing is done or that the function errors out if neither condition is met?

dsavineau: both are valid solutions but I guess to make sense to log and exit in this situation

djgalloway:
[This](https://github.com/ceph/ceph-build/blob/master/ceph-container-lint/build/build#L28) doesn't work if the changed file is deleted.

dsavineau:
jenkins lint

dsavineau:
jenkins lint

PR #1964, opened 2021-11-08
daemon-base: add missing spaces

edevil:
Add missing space so that package names are not
clobbered together.

Otherwise we would obtain things like:
```
[docker_build]  [91mError: Unable to find a match: python3-samlceph-grafana-dashboards-16.2.6
```


edevil:
It's still a valid issue.

PR #1966, opened 2021-11-15
Dockerfile: remove Sree for ubi8 based image

guits:
this is only needed for ceph-nano, which has never been productized
downstream.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2020150


dsavineau:
> this is only needed for ceph-nano, which has never been productized downstream.

In fact, this is not totally true. Sree is used by the demo entrypoint which is called by ceph-nano.

As a consequence, the current change is breaking the demo scenario.

PR #1979, opened 2022-01-31
distro: use CentOS Stream 8 instead of deprecated CentOS 8

nixpanic:
CentOS 8 has been deprecated, and the contents of the repositories is
removed from the mirror network. Installing packages in a CentOS 8
container image is not possible anymore.

CentOS Stream 8 is the only CentOS 8 variant that is maintained.

See-also: https://centos.org/cl-vs-cs/

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -8,7 +8,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: build the ceph container image
-        run: make RELEASE="demo" FLAVORS="master,centos,8" build
+        run: make RELEASE="demo" FLAVORS="master,centos,stream8" build
 
       - name: run the ceph demo container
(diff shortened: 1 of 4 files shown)


Madhu-1:
@leseb can someone help on this one? cephcsi is completely blocked due to https://github.com/ceph/ceph-csi/issues/2836 we are not able to build cephcsi images and non of the cephcsi CI jobs will pass as the image building fails. we need this fix to get unblocked if we can get a official ceph image with this fix it will be great :+1:

humblec:
@dsavineau may be you could help ?

mykaul:
Has Ceph moved as well?

nixpanic:
> Has Ceph moved as well?

I do not know. But as there are no updates to CentOS 8 any longer, it is not an option to stay on CentOS 8. Everything that consumed CentOS 8 should have moved to Stream 8 or an alternative distribution.

nixpanic:
For Ceph-CSI we might be able to use ceph/ceph-csi#2837 as a workaround until the Ceph container-images become fully functional again.

guits:
@nixpanic i'm taking a look at this

guits:
closing this in favor of #1980

PR #1993, opened 2022-03-01
add quincy release

guits:
quincy v17.1.0 release is available.
Let's build quincy images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@
 FLAVORS ?= \
 	octopus,centos,8 \
 	pacific,centos,8 \
+	quincy,centos,8 \
 	master,centos,8
 
 TAG_REGISTRY ?= ceph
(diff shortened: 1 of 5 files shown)


guits:
`ceph-release` package is missing at https://download.ceph.com/rpm-quincy/el8/noarch/  😢

dsavineau:
You'll also need to allow the RC releases to be built [1] if you want to build 17.1.0 (assuming the ceph-release package issue is fixed)

```diff
-   local pkg_regex=">ceph-[0-9]+.[2].[0-9]+-[0-9]+"
+   local pkg_regex=">ceph-[0-9]+.[12].[0-9]+-[0-9]+"
```

The RC builds where removed from that regex to avoid rebuilding the nautilus RC releases back when the RPM release directory had all the releases in the same directory (<= nautilus).
Because we're building only for >=octopus then only the latest release is rebuilt so it's safe to keep the RC release from the regex.

https://github.com/ceph/ceph-container/blob/master/contrib/ceph-build-config.sh#L240

dsavineau (approved):

PR #2001, opened 2022-03-26
Use ceph-volume lvm deactivate to cleanup on sigterm

mstinsky:
As of octopus the ceph-volume command has a deactivate option which implements what is done in the sigterm_cleanup_post function. This MR removes the current cleanup code and just does a `ceph-volume lvm deactivate` to unmount the lvs and close the crypts. This helps sharing code with ceph/ceph.

While implementing this I found a bug with `ceph-volume lvm deactivate` which prevents closing the dmcrypts, so this MR should be blocked until https://github.com/ceph/ceph/pull/45659 is resolved.

Also I am not really familiar  with the repo structure and workflow I was reading a bit in the CONTRIBUTING.md and if I understand this correctly this MR would also need the old `osd_disk_activate.sh` inside `ceph-releases/nautilus/daemon/osd_scenarios/` correct? Because this would not be compatible with nautilus.

### DIFF
--- a/src/daemon/osd_scenarios/osd_volume_activate.sh
+++ b/src/daemon/osd_scenarios/osd_volume_activate.sh
@@ -46,14 +46,13 @@ function osd_volume_simple {
   fi
 }
 
-function get_dmcrypt_uuids {
-  dmsetup ls --target=crypt | cut -d$'\t' -f 1
+function get_osd_fsid {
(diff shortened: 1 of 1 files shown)


mstinsky:
Just figured that this MR as is breaks osds with simple setup.
Not sure if we would want to restructure the if else statements to use `ceph-volume lvm deactivate` for lvm volumes and keep the old method for simple volumes or just keep everything as is. 
Any opinions?

PR #2003, opened 2022-04-05
[skip ci] contrib: update API endpoint for existing tags

dsavineau:
The quay.io API endpoint `repository/<repo>/<ns>/tag/<tag>/images` doesn't
exist anymore.
Instead we can use the `repository/<repo>/<ns>/tag/` endpoint with the
`specificTag` parameter.
As a consequence we now need to test the tags array length from the response
because we can't only rely on the curl return code.
This also change the workflow when the curl command fails for any reason. The
current workflow was to continue the execution while the script stops now.


guits (approved):

guits:
thanks maggle!

PR #2013, opened 2022-06-01
base: add sg3_utils package

guits:
This adds the sg3_utils package.

### DIFF
--- a/src/daemon-base/__CEPH_BASE_PACKAGES__
+++ b/src/daemon-base/__CEPH_BASE_PACKAGES__
@@ -15,6 +15,7 @@
         nvme-cli \
         libstoragemgmt \
         systemd-udev \
+        sg3_utils \
         procps-ng \
         hostname \
         __RADOSGW_PACKAGES__ \


ktdreyer:
What problem does this change solve?

Should Ceph's packages depend on this package?

pcuzner:
@ktdreyer the reason we need the package is the disk failure handling workflow. When a disk fails, and it's still seen by the kernel the current lsmcli integration works...but, when the device stops responding to test-unit-ready, the kernel drops the device node, which means the lsmcli integration no longer works. We know we see this on some of the external enclosures for example. By included sg3_utils we get access to sg_ses, which allows us to talk directly to the slot in the enclosure to activate the LED.

So getting sg3_utils included is the first step in improving our failed drive handling with external enclosures.

guits:

guits:

guits:

guits:

guits:

guits:

guits:

guits:
@ktdreyer I personally don't mind pulling the package by dependency during the image build or installing it explicitly, as long as we end up with the package present in the image.
Which option do we prefer?

guits:
> @ktdreyer I personally don't mind pulling the package by dependency during the image build or installing it explicitly, as long as we end up with the package present in the image. Which option do we prefer?

@ktdreyer ^

ktdreyer:
Seems like we should put it into the RPM

PR #2032, opened 2022-07-15
ubi9: set ignore_absent_pulp_repos

ktdreyer:
ODCS fails to compose because it cannot find the `rhceph-6-tools-for-rhel-9-*-rpms` repositories in Pulp yet.

(Revert this when release engineering has created these repositories.)


ktdreyer:
cc @bigjust @tserlin

guits (approved):

PR #2044, opened 2022-08-08
ubi9: reinstall systemd package

ktdreyer:
The ubi9 base image deletes `libsystemd` from the `systemd` package. To fix this and reinstate the `libsystemd` shared library, we must reinstall the `systemd` package.

We can revert this change when the ubi9-minimal image does not ship a broken `systemd` package.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2114004


ktdreyer:
I've scratch-built an image with this change and I confirmed `/usr/lib/systemd/libsystemd-shared-250.so` is present. Simple commands like `udevadm` do not crash.

guits (approved):

ktdreyer:
I think we need to backport to `stable-7.0`, right?

PR #2041, opened 2022-08-04
base: add sssd-client to NFS-Ganesha packages

BlaineEXE:
In order for NFS-Ganesha to use System Security Services Daemon
(SSSD)[1], it requires the sssd-client packages to be installed. This
should add less than 500 kB to the image.

NFS-Ganesha can use SSSD for user ID management. SSSD supports LDAP,
Active Directory, and FreeIPA. Notably, SSSD is the primary tool used
to allow FreeIPA support.

[1] https://sssd.io

Signed-off-by: Blaine Gardner <blaine.gardner@redhat.com>

Description of your changes:

Which issue is resolved by this Pull Request:
Resolves #

Checklist:
- [ ] Documentation has been updated, if necessary.
- [ ] Pending release notes updated with breaking and/or notable changes, if necessary.

PR #2046, opened 2022-08-09
base: add sssd-client to centos 8 and ubi9 images

BlaineEXE:
The latest Ceph builds don't contain sssd-client along with the
nfs-ganesha packages. The upstream builds are from CentOS 8, which has
overrides defined. Add sssd-client to these overrides as well

Also add the package to the RHEL-based ubi9.

Signed-off-by: Blaine Gardner <blaine.gardner@redhat.com>

Description of your changes:

Resolves #

Checklist:

### DIFF
--- a/ceph-releases/ALL/centos/8/daemon-base/__GANESHA_PACKAGES__
+++ b/ceph-releases/ALL/centos/8/daemon-base/__GANESHA_PACKAGES__
@@ -1,1 +1,0 @@
-nfs-ganesha nfs-ganesha-ceph nfs-ganesha-rgw nfs-ganesha-rados-grace nfs-ganesha-rados-urls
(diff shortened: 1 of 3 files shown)


ktdreyer:
It's unclear to me why these files are different than what you've already updated in https://github.com/ceph/ceph-container/pull/2041 . Maybe that difference is accidental, and we could reconcile them.

BlaineEXE:
> It's unclear to me why these files are different than what you've already updated in #2041 . Maybe that difference is accidental, and we could reconcile them.

I assumed it was for some particular reason. The two files in this PR (that override the file I updated in #2041) add the package `nfs-ganesha-rados-urls`. The summary for the package says `This package contains the libganesha_rados_urls library used for handling RADOS URL configurations.`.

guits:
It seems to be some legacies or a leftover from a refactor.

I think we should just keep e875a2fdf678918ae9ab58c4617340fd6f58cb5a and drop 193a961a1aa41fb10d7fdec7ecf062ee6135f0bc from this PR.

BlaineEXE:
Done. Thanks @guits

guits (approved):

ktdreyer:
Excellent, thanks @BlaineEXE and @guits

PR #2058, opened 2022-08-31
build: Prevent `error: db5 error` errors

djgalloway:
https://access.redhat.com/solutions/6962401
https://access.redhat.com/solutions/5884221

Passed: https://jenkins.ceph.com/job/ceph-dev-container-only/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos8,DIST=centos8,MACHINE_SIZE=gigantic/10/console

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -88,4 +88,6 @@ bash -c ' \
     yum copr enable -y tchaikov/python-scikit-learn ; \
     yum copr enable -y tchaikov/python3-asyncssh ; \
   fi ' && \
+systemctl stop packagekit.service ; \
+rm -f /sysroot/var/lib/rpm/__db00* ; \
 yum install -y --setopt=install_weak_deps=False __CEPH_BASE_PACKAGES__


tchaikov (approved):

dsavineau:
systemd isn't running within the container so the `systemctl stop packagekit.service` command doesn't make sense in the Dockerfile.
You can see this from the logs:

```console
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

Also the `/sysroot` directory doesn't exist with the container image so the `rm -f` command fails silently due to the force option

```console
STEP 3: RUN rm /sysroot/var/lib/rpm/__db00*
rm: cannot remove '/sysroot/var/lib/rpm/__db00*': No such file or directory
```

Even with this change, the db5 error is still present in the jenkins log https://jenkins.ceph.com/job/ceph-dev-container-only/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos8,DIST=centos8,MACHINE_SIZE=gigantic/10/console

```console
error: db5 error(-30969) from dbenv->open: BDB0091 DB_VERSION_MISMATCH: Database environment version mismatch
error: cannot open Packages index using db5 -  (-30969)
error: cannot open Packages database in /var/lib/rpm
```

Because this is coming just after the update command while the change is done later.

Also note that this db5 error isn't blocking the container build.

djgalloway:
https://github.com/ceph/ceph-container/pull/2059

djgalloway:
Shoot, I don't even know how `/sysroot` got in there.

If that error isn't fatal though, we can revert this anyway.

PR #2085, opened 2023-01-24
Do Not Merge: IBM Ceph changes

ktdreyer:
I'm testing this change to the IBM Ceph `Dockerfile`.

### DIFF
--- a/ceph-releases/ALL/ubi8/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/ubi8/daemon-base/__DOCKERFILE_INSTALL__
@@ -1,2 +1,14 @@
 microdnf update -y --setopt=install_weak_deps=0 --nodocs && \
-microdnf install -y --setopt=install_weak_deps=0 --nodocs wget unzip util-linux python3-saml python3-setuptools udev device-mapper __CEPH_BASE_PACKAGES__
+microdnf install -y --setopt=install_weak_deps=0 --nodocs wget unzip util-linux python3-saml python3-setuptools udev device-mapper __CEPH_BASE_PACKAGES__ && \
+ln -s /usr/share/ceph/mgr/dashboard/frontend/dist-ibm /usr/share/ceph/mgr/dashboard/frontend/dist && \
+sed -i \
+  -e "s|registry.redhat.io/rhceph/rhceph-5-rhel8:latest|icr.io/ibm-ceph/ceph-5-rhel8|" \
+  -e "s|registry.redhat.io/openshift4/ose-prometheus:v4.10|icr.io/ibm-ceph/prometheus:v4.10|" \
+  -e "s|registry.redhat.io/openshift4/ose-prometheus-node-exporter:v4.10|icr.io/ibm-ceph/prometheus-node-exporter:v4.10|" \
+  -e "s|registry.redhat.io/rhceph/rhceph-5-dashboard-rhel8:latest|icr.io/ibm-ceph/ceph-5-dashboard-rhel8:latest|" \
+  -e "s|registry.redhat.io/openshift4/ose-prometheus-alertmanager:v4.10|icr.io/ibm-ceph/prometheus-alertmanager:v4.10|" \
+  -e "s|registry.redhat.io/rhceph/rhceph-haproxy-rhel8:latest|icr.io/ibm-ceph/haproxy-rhel8:latest|" \
+  -e "s|registry.redhat.io/rhceph/keepalived-rhel8:latest|icr.io/ibm-ceph/keepalived-rhel8:latest|" \
+  -e "s|registry.redhat.io/rhceph/snmp-notifier-rhel8:latest|icr.io/ibm-ceph/snmp-notifier-rhel8:latest|" \
+  -e "s|default='registry.redhat.io'|default='icr.io'|" \
+  /usr/share/ceph/mgr/cephadm/module.py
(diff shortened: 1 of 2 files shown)


ktdreyer:
I've built a container with this change, and I confirmed with `grep -R registry.redhat.io /usr/share/ceph/mgr/` that there are no more references to Red Hat's registry.

The next step with this PR is to decide how we want to merge this into this repository. At first I considered simply copying the entire `ubi8` directory to something like `ibm-ubi8`, but that would introduce more duplication. For simplicity I've just made my IBM changes directly to the `ubi8` files so that it's easy to review the diff. Of course we cannot merge it like this because it would overwrite the things we want to keep for Red Hat.

ceph-releases/ALL/ubi8/daemon-base/__DOCKERFILE_INSTALL__
```diff
@@ -1,2 +1,14 @@
...
+  -e "s|registry.redhat.io/rhceph/rhceph-5-rhel8:latest|icr.io/ibm-ceph/ceph-5-rhel8|" \
+  -e "s|registry.redhat.io/openshift4/ose-prometheus:v4.10|icr.io/ibm-ceph/prometheus:v4.10|" \
+  -e "s|registry.redhat.io/openshift4/ose-prometheus-node-exporter:v4.10|icr.io/ibm-ceph/prometheus-node-exporter:v4.10|" \
+  -e "s|registry.redhat.io/rhceph/rhceph-5-dashboard-rhel8:latest|icr.io/ibm-ceph/ceph-5-dashboard-rhel8:latest|" \
+  -e "s|registry.redhat.io/openshift4/ose-prometheus-alertmanager:v4.10|icr.io/ibm-ceph/prometheus-alertmanager:v4.10|" \
+  -e "s|registry.redhat.io/rhceph/rhceph-haproxy-rhel8:latest|icr.io/ibm-ceph/haproxy-rhel8:latest|" \
+  -e "s|registry.redhat.io/rhceph/keepalived-rhel8:latest|icr.io/ibm-ceph/keepalived-rhel8:latest|" \
+  -e "s|registry.redhat.io/rhceph/snmp-notifier-rhel8:latest|icr.io/ibm-ceph/snmp-notifier-rhel8:latest|" \
+  -e "s|default='registry.redhat.io'|default='icr.io'|" \
+  /usr/share/ceph/mgr/cephadm/module.py
```
adk3798, line 14: are we going to do this for the cephadm binary as well (`/usr/sbin/cephadm`)? It's probably not strictly necessary for things to work, but the version in the container will be getting deployed on the host by the mgr module.

ktdreyer: Yes, we edit the `/usr/sbin/cephadm` binary separately (during the `cephadm` RPM build). The reason we do that before we install the `cephadm` RPM into the container is that since we must distribute that `cephadm` RPM separately for users to bootstrap a cluster.

(Incidentally this last `-e "s|default='registry.redhat.io'|default='icr.io'|"` substitution is slightly different for `module.py` compared to `cephadm`, where we have to do `-e "s|DEFAULT_REGISTRY = 'registry.redhat.io'|DEFAULT_REGISTRY = 'icr.io'|"`)

adk3798 (approved):
The parts I understand (editing the cephadm mgr module's default images mostly) LGTM at least.

ceph-releases/ALL/ubi8/daemon-base/__DOCKERFILE_PREINSTALL__
```diff
@@ -1,17 +1,14 @@
...
 # Build specific labels
-LABEL com.redhat.component="rhceph-container"
-LABEL name="rhceph"
-LABEL description="Red Hat Ceph Storage 5"
-LABEL summary="Provides the latest Red Hat Ceph Storage 5 on RHEL 8 in a fully featured and supported base image."
-LABEL io.k8s.display-name="Red Hat Ceph Storage 5 on RHEL 8"
-LABEL io.openshift.tags="rhceph ceph"
+LABEL com.redhat.component="ibm-ceph-container"
+LABEL name="ibm-ceph"
+LABEL description="IBM Ceph Storage 5"
```
nizamial09, line 11: shouldn't it be `IBM Storage Ceph 5`?

nizamial09 (approved):

ceph-releases/ALL/ubi8/daemon-base/__DOCKERFILE_PREINSTALL__
ktdreyer: I'll drop `Storage`, because I think it's just named "IBM Ceph"

ktdreyer: Or, actually you would know better than me! What are you using in the dashboard?

nizamial09: We use IBM Storage Ceph actually. It was the agreed product name according to our designs..

guits:
@ktdreyer what is the status of this PR?

the target branch is stable-6.0 by the way

ktdreyer:
We still need to land IBM's changes in this repository. I am not sure what files you'd like to use for this.

In the meantime I've rebased this onto `main` in PR #2090 , so I can close this one and we can continue there.

PR #2094, opened 2023-02-02
compose-rhcs: various fixes

guits:
- revert daemon image removal
- update daemon Dockerfile
- rsync stable-6.0 / ubi8 content to the branch 'main'


ktdreyer (changes requested):
I ran `VERSION=5 ./contrib/compose-rhcs.sh` with this PR.

Here's the diff I saw: https://pastebin.mozilla.org/8rs5LBcb

- [ ] `FROM` is wrong
- [ ] Many `ENV` settings are gone. Is this ok to drop things like `CEPH_VERSION pacific`, `CEPH_DEVEL false`,
`CEPH_REF pacific`, `OSD_FLAVOR default`?
- [ ] `microdnf` steps are deleted

ktdreyer:
Here's the diff I'm seeing for `VERSION=5`: https://pastebin.mozilla.org/XO7fY42q . This looks better to me.

For `VERSION=6` we must fix the following issues (diff at https://pastebin.mozilla.org/NHLJc3gs):

- [ ] We have no `ceph-exporter` package, so `microdnf` will fail to install that.
- [ ] We do not need the `Add ceph-container files` section. Those files do not exist.

PR #2097, opened 2023-02-06
ubi9: do not install ceph-exporter

guits:
this package isn't available in internal/downstream repos.


ktdreyer:
I think we could close this PR without merging. We'll just ship the `ceph-exporter` subpackage in RHCS 6.

PR #2098, opened 2023-02-09
split ubi8 / introduce ibm branding

guits:
The idea is to split the ubi8 image so we can build it with either the IBM or the Red Hat branding.


ktdreyer (approved):
I've run the following:

```
VERSION=5 ./contrib/compose-rhcs.sh
```
This deletes the Sree app and demo script as intended. (We won't ship those for IBM)

```
VERSION=5 BRANDING=redhat ./contrib/compose-rhcs.sh
```
This only results in cosmetic changes to the contents of `ceph-5.3-rhel-8`.

ktdreyer:
I'm requesting the `containers/ibm-ceph` dist-git repo now so we can populate `ibm-ceph-5.3-rhel-8`.

PR #2106, opened 2023-03-23
contrib/rhcs.md improvements

ktdreyer:
* We use the **minimal** `ubi9` image now for the rhcs image. Use that in the documentation.
* The ODCS repository supports multiple architectures. Use the `$basearch` repo so it works on ppc64le and s390x.
* Use a generic ODCS .repo filename so that we do not have to remember to bump this when we ship new versions.


ktdreyer:
CC @tuan-hoang1

guits (approved):

PR #2148, opened 2023-08-29
ceph-releases/main/centos: use V5.5 of nfs-ganesha to build centos 8

ljflores:
Description of your changes:

Following https://github.com/ceph/ceph-build/pull/2161, which changes the nfs-ganesha jenkins job to periodically build V5.5 instead of "next", we can update the container script to build with V5.5.

Checklist:


guits:
I think it makes sense that ceph:main images uses nfs-ganesha@devel (next) builds

PR #2153, opened 2023-09-06
ceph-releases/ALL/centos/daemon-base: change hardcoded nfs-ganesha ve…

tserlin:
…rsion

Change hard-coded nfs-ganesha version V5.3.2 to V5.5.

We are currently pinning the daily nfs-ganesha builds to V5.5 in https://jenkins.ceph.com/job/nfs-ganesha

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -3,7 +3,7 @@ yum install -y jq && \
 bash -c ' \
   if [ -n "__GANESHA_PACKAGES__" ]; then \
     if [[ "${CEPH_VERSION}" == master || "${CEPH_VERSION}" == main ]]; then \
-      curl -s -L "https://shaman.ceph.com/api/repos/nfs-ganesha/V5.3.2/latest/centos/__ENV_[DISTRO_VERSION]__/flavors/ceph_main/repo?arch=$(arch)" -o /etc/yum.repos.d/ganesha.repo ; \
+      curl -s -L "https://shaman.ceph.com/api/repos/nfs-ganesha/V5.5/latest/centos/__ENV_[DISTRO_VERSION]__/flavors/ceph_main/repo?arch=$(arch)" -o /etc/yum.repos.d/ganesha.repo ; \
     elif  [[ "${CEPH_VERSION}" == reef ]]; then \
       echo "[ganesha]" > /etc/yum.repos.d/ganesha.repo ; \
       echo "name=ganesha" >> /etc/yum.repos.d/ganesha.repo ; \


ljflores (commented):
What about reef, quincy, and pacific? We should decide on that too.

Also, @guits had refactord the nfs-ganesha repo with this commit:
https://github.com/ceph/ceph-container/commit/b86898a9623bd93398ee9e4059d08ecda7ab3a4c

But I had to revert it in https://github.com/ceph/ceph-container/pull/2151 since it caused a regression in our container scripts (see more details about the regression in https://pad.ceph.com/p/16.2.14-struggles).

@guits mentioned he wants to reintroduce this change; @guits do you agree it could b handled in a subsequent PR?

ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
```diff
@@ -3,7 +3,7 @@ yum install -y jq && \
 bash -c ' \
   if [ -n "__GANESHA_PACKAGES__" ]; then \
     if [[ "${CEPH_VERSION}" == master || "${CEPH_VERSION}" == main ]]; then \
-      curl -s -L "https://shaman.ceph.com/api/repos/nfs-ganesha/V5.3.2/latest/centos/__ENV_[DISTRO_VERSION]__/flavors/ceph_main/repo?arch=$(arch)" -o /etc/yum.repos.d/ganesha.repo ; \
+      curl -s -L "https://shaman.ceph.com/api/repos/nfs-ganesha/V5.5/latest/centos/__ENV_[DISTRO_VERSION]__/flavors/ceph_main/repo?arch=$(arch)" -o /etc/yum.repos.d/ganesha.repo ; \
```
guits, line 6: @ljflores I think we should test ceph devel builds with nfs-ganesha builds

ljflores: @kalebskeithley thoughts?

@guits in that case, we'd have to add something to the nfs-ganesha Jenkins job so it periodically builds "next" as well as V5.5: https://github.com/ceph/ceph-build/blob/main/nfs-ganesha/config/definitions/nfs-ganesha.yml

kalebskeithley: 'next' is the bleeding edge devel branch. It's not really suitable, IMO, to be used. For anything

If you use a 'next' branch build and something doesn't work we may or may not make it a priority to fix.

I strongly suggest you use only released/tagged VX.Y versions, e.g. V5.5, V5.6, etc.   V5.Y is perfectly fine for pacific, quincy, reef, and beyond.

I don't understand the comment that you need to build 'next' occasionally.  Why?

ljflores: In that case, let's stick with V5.5.

ljflores (commented):
LGTM, except we need to ensure that V5.5 is also being used for reef, quincy, and pacific. I'm not sure if further changes need to be made to the script for that to happen.

@tserlin

tserlin:
> LGTM, except we need to ensure that V5.5 is also being used for reef, quincy, and pacific. I'm not sure if further changes need to be made to the script for that to happen.
> 
> @tserlin

Are you okay with merging this change first and seeing if we get a good build from it? Then I'll get a PR for the others.

ljflores:
Sure

ljflores (approved):

PR #2164, opened 2023-10-31
base: add iproute package

asm0deuz:
This makes the ip command available.

Fixes:https://bugzilla.redhat.com/show_bug.cgi?id=2245975

### DIFF
--- a/src/daemon-base/__UTIL_PACKAGES__
+++ b/src/daemon-base/__UTIL_PACKAGES__
@@ -1,2 +1,3 @@
 procps-ng \
-        hostname
+hostname \
+iproute


ktdreyer:
Looks like Rook has worked around this with https://github.com/red-hat-storage/rook/pull/533 instead

BlaineEXE:
My experience has been that the upstream `quay.io/ceph/ceph` images already have `ip` installed. It seems like that isn't by intent, given that the package isn't installed. From a container security perspective, it would be best to not include `ip` in the Ceph image because it isn't needed by Ceph. I will make that same change to upstream Rook as well.

ktdreyer:
That's right @BlaineEXE . The `iproute` package in `quay.io/ceph/ceph:v18.2.0` comes preinstalled in the base image, `quay.io/centos/centos:stream8`.

Once Ceph upstream begins to use `quay.io/centos/centos:stream9-minimal` as the base image, the `iproute` package will not be in Ceph. Nothing in Ceph itself uses `/usr/sbin/ip`, so it's there by "accident".

I agree with you that we should minimize our package list to avoid exposure to future vulnerabilities. I checked `iproute-6.4.0-1.fc38` on Fedora, and there are no setuid binaries, but it still helps to keep our package list small if possible.

ktdreyer:
@asm0deuz you can close this PR.

PR #2166, opened 2023-11-08
bump ganesha version in main containers to 5.7

adk3798:
nfs-ganesha 5.7 is now being built and pushed to
https://shaman.ceph.com/builds/nfs-ganesha/, but we're still only using 5.5 builds.
5.7 should also have the new cmount_path feature so let's use that.

### DIFF
--- a/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
+++ b/ceph-releases/ALL/centos/daemon-base/__DOCKERFILE_INSTALL__
@@ -4,7 +4,7 @@ bash -c ' \
   if [ -n "__GANESHA_PACKAGES__" ]; then \
     if [[ "${CEPH_VERSION}" == master || "${CEPH_VERSION}" == main ]]; then \
       ARCH=$(arch); if [[ "${ARCH}" == "aarch64" ]]; then ARCH="arm64"; fi ; \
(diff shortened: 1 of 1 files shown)


guits:
This is one of the reasons why I was -1 on pinning to a version on main builds, we now have to maintain this.

guits (approved):

guits:
jenkins test sloppy bash

guits:
jenkins lint

dparmar18:
> This is one of the reasons why I was -1 on pinning to a version on main builds, we now have to maintain this.

there is a `V5-stable` branch, which can be used to prevent any conscious efforts to maintain the version.

dparmar18:
> > This is one of the reasons why I was -1 on pinning to a version on main builds, we now have to maintain this.
> 
> there is a `V5-stable` branch, which can be used to prevent any conscious efforts to maintain the version.

PS: this was discussed in cephfs channel upstream https://ceph-storage.slack.com/archives/C04LVQMHM9B/p1699485905106709?thread_ts=1699420006.629709&cid=C04LVQMHM9B with @ffilz

dparmar18:
@adk3798 @guits I created https://github.com/ceph/ceph-container/pull/2167 that adds `V5-stable`, PTAL.

PR #2168, opened 2023-11-09
reef: pin nfs-ganesha to 5.5-1

guits:
As 5.5-2 introduced a regression, we need to pin to this specific version for reef.


adk3798:
NOTE: regression in question https://tracker.ceph.com/issues/63151. Also keep in mind this only is an issue with reef. Main ceph seems to work okay with newer ganesha version (e.g. 5.7)

adk3798 (approved):

BlaineEXE:
FWIW, Matt Benjamin mentioned that the RGW-side issue is on QA here: https://github.com/ceph/ceph/pull/54297#pullrequestreview-1709007226

PR #2176, opened 2023-12-06
ubi9-ibm: unify ADD steps

ktdreyer:
The `quincy` and `reef` templates override the contents of the `ALL` template. As a result, prior to this change, we missed an `ADD` command in the generated `Dockerfile`.

Unify and simplify the instructions so we include the proper steps for `quincy` and `reef`.


bigjust:
Ran compose-rhcs with this and it matched the downstream Dockerfile that had both ADD lines, LGTM

bigjust (approved):

guits (approved):

PR #2185, opened 2024-02-13
add ceph-node-proxy package to ubi9-based IBM images

guits:
This adds the ceph-node-proxy RPM package to the ubi9-based IBM Ceph downstream image.

### DIFF
--- a/ceph-releases/ALL/ubi9-ibm/daemon-base/__CEPH_NODE_PROXY_PACKAGE__
+++ b/ceph-releases/ALL/ubi9-ibm/daemon-base/__CEPH_NODE_PROXY_PACKAGE__
@@ -1,0 +1,1 @@
+ceph-node-proxy__ENV_[CEPH_POINT_RELEASE]__


guits:
@ktdreyer @bigjust @jmolmo

bigjust:
I regenerated the Dockerfile using `make stage` and didn't see the ceph-node-proxy package being added to the reef-ubi9-ibm Dockerfile, which is what I would have expected. Is that how this should be tested?

bigjust:
> @guits I regenerated the Dockerfile using `make stage` and didn't see the ceph-node-proxy package being added to the reef-ubi9-ibm Dockerfile, which is what I would have expected. Is that how this should be tested?

Its possible that I need to test this with https://pkgs.devel.redhat.com/cgit/rpms/ceph/commit/?h=ceph-7.1-rhel-9&id=9ff0aa8b9f9d7c254978d8593c50e22359d04eb1 which I will attempt, but please correct me if this is wrong

guits:
> @guits I regenerated the Dockerfile using `make stage` and didn't see the ceph-node-proxy package being added to the reef-ubi9-ibm Dockerfile, which is what I would have expected. Is that how this should be tested?

I used the script at `./contrib/compose-rhcs.sh` to check that @bigjust

PR #2194, opened 2024-03-21
set the max locked-in to 204800 for root and ceph users

guits:
This sets the maximum locked-in memory to ~200Mb for both root and ceph users.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2266529

### DIFF
--- a/ceph-releases/ALL/__QAT_ACCELERATION__
+++ b/ceph-releases/ALL/__QAT_ACCELERATION__
@@ -1,0 +1,2 @@
+echo '@ceph - memlock 204800' >> /etc/security/limits.conf && \
+    echo '@root - memlock 204800' >> /etc/security/limits.conf && \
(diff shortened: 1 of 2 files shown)


mkogan1 (approved):
LGTM

ceph-releases/ALL/__QAT_ACCELERATION__
```diff
@@ -0,0 +1,2 @@
+echo '@ceph - memlock 204800' >> /etc/security/limits.conf && \
+    echo '@root - memlock 204800' >> /etc/security/limits.conf && \
```
mkogan1, line 2:
```suggestion
echo '@root - memlock 204800' >> /etc/security/limits.conf && \
```

guits: it is actually needed otherwise the rendered Dockerfile has some extra spaces

PR #2196, opened 2024-03-27
Code Cleanup: Move cephfs-top package to CEPHFS package list

jmundack:
Currently `cephfs-top` is a part of the ceph base packages. This is to move it to be alongside the rest of the cephfs packages. This will allow for changing cephfs packages (enabling and disabling as well) based on release/distribution/etc


jmundack:
@guits  - mind taking a look?
https://github.com/ceph/ceph-container/pull/2114 is the PR that introduced this change

PR #2197, opened 2024-03-28
edit /etc/security/limits.conf after installing ceph

ktdreyer:
The `pam` RPM (and `/etc/security` directory) is not present in the ubi9-minimal base images. For minimal images, `pam` is only available after we install ceph.

Edit the `/etc/security/limits.conf` file after ceph (and pam) is installed.

This fixes this error introduced in https://github.com/ceph/ceph-container/pull/2194:

```
/bin/sh: line 1: /etc/security/limits.conf: No such file or directory
```

### DIFF
--- a/src/daemon-base/Dockerfile
+++ b/src/daemon-base/Dockerfile
@@ -31,8 +31,8 @@ RUN \
     # Typical workflow: add new repos; refresh repos; install packages; package-manager clean;
     #   download and install packages from web, cleaning any files as you go.
     # Installs should support install of ganesha for luminous
-    __QAT_ACCELERATION__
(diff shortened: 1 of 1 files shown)


tserlin:
This just affected us downstream in our container builds, with the same error, as the dist-git "containers/rhceph" repo was updated a few hours ago.

guits (approved):

tserlin:
@guits Can you sync the change with downstream? I tried a 7.1 scratch container build downstream with the change manually applied and it does build for me :+1:

ktdreyer:
We'll need this in rhceph and ibm-ceph downstream.

PR #2204, opened 2024-05-01
Don't delete files in /usr/lib64/systemd for centos 9 images

jagombar:
When starting up a cluster from a centos 9 based container image, the following entries are logged multiple times:
  ceph-volume.log:[2024-05-01 15:28:51,612][ceph_volume.process][INFO  ] stderr /usr/sbin/udevadm: error while loading shared libraries: libsystemd-shared-252.so: cannot open shared object file: No such file or directory

Further investigation showed that the Dockerfile for centos 9 was deleting everything in /usr/lib64/systemd/ because it is built from the default /src/__DOCKERFILE_CLEAN_COMMON__.  I think that centos 9 should be keeping the files in this directory in the same way that centos 8 does.

Description of your changes:

Moved /8/__DOCKERFILE_CLEAN_COMMON__ into its parent directory.  Considered copying the file into the /9/ subdirectory, but that seemed unnecessary.

Resolves #

Checklist:


dmick (changes requested):
This looks exactly right, but also needs to be done in the centos-arm64 directory

jagombar:
ah yes, I completely missed that.  Fixed now.

PR #2214, opened 2024-05-29
contrib: add extra identifier tag to c8 containers instead of c9

ljflores:
In response to c8 going EOL (https://lists.ceph.io/hyperkitty/list/dev@ceph.io/thread/H7I4Q4RAIT6UZQNPPZ5O3YB6AUXLLAFI/), we will need to have the orch and upgrade suites run with c9 containers instead of c8.

Description of your changes:

Resolves #

Checklist:

### DIFF
--- a/contrib/build-push-ceph-container-imgs.sh
+++ b/contrib/build-push-ceph-container-imgs.sh
@@ -303,11 +303,14 @@ function push_ceph_imgs_latest {
     full_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${RELEASE}-${distro}-stream${distro_release}-${HOST_ARCH}-devel
     branch_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${BRANCH}
     sha1_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}
-    # for centos9, while we're still building centos8, add -centos9 to branch and sha1 tags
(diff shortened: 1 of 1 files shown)


ljflores:
Big **DNM** until we understand what other working parts are needed.

The current logic in ceph-build is to build c8 and c9 for quincy, reef, squid, and main. This PR will make it so c8 containers are still available for as long as we are able to build them, but they will be named differently (with a `-centos8` identifier at the end).

If people still want to test on c8 containers, they will have to pass in an override yaml, something like this:
```
overrides:
  cephadm:
    image: 'quay-quay-quay.apps.os.sepia.ceph.com/ceph-ci/ceph:7c8f650b36e258f639fa4a83becade57cbfd2009-centos8'
```

dmick (approved):
seems sensible.  of course this impacts teuthology testing as soon as it merges

ljflores:
@dmick there is also a PR open to make teuthology ensure c9 packages have been built before people can run tests. I think this should be merged first: https://github.com/ceph/teuthology/pull/1944

ktdreyer (approved):

contrib/build-push-ceph-container-imgs.sh
```diff
@@ -305,9 +305,9 @@ function push_ceph_imgs_latest {
     sha1_repo_tag=${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}/ceph:${SHA1}
     # for centos9, while we're still building centos8, add -centos9 to branch and sha1 tags
```
ktdreyer, line 306: I think you need to update this comment

ljflores: Will do

ljflores:
https://github.com/ceph/teuthology/pull/1944 has been merged.

dmick (approved):

PR #2219, opened 2024-06-04
ubi9: exclude ppc64le for squid

ktdreyer:
We cannot build squid on RHEL 9 on ppc64le yet due to https://tracker.ceph.com/issues/66306

Revert this change when that issue is resolved and we can build squid on ppc64le.


tserlin (approved):
LGTM

PR #2220, opened 2024-06-05
Makefile: update FLAVORS, ALL_BUILDABLE_FLAVORS

guits:
This changes the default images built (el8 -> el9)


guits:
sloppy tests are broken because they are running on el8 jenkins workers

PR #2227, opened 2024-06-05
ubi9: add aarch64 content sets for squid

ktdreyer:
We're adding support for aarch64 in a future release. Define the content sets for OSBS here.


tserlin (approved):
LGTM

PR #2231, opened 2024-06-09
Support building a Debian 12 Reef container image

thomasgoirand:
This patch adds the necessary ceph-releases/ALL/debian/*/__OS_CODENAME__ files needed to know the Debian release names, so the base container can be built. It also configures the correct package list, as it varies a little bit from the one under CentOS. Finally, the ceph-releases/ALL/debian/daemon-base/__DOCKERFILE_INSTALL__ has been fixed so it uses the minimal Bookworm (currently unofficial, but this will change) repository.

I tested this, and I was able to build a Debian image with it. Comments welcome.


guits (approved):

guits (changes requested):
@thomasgoirand please, add a commit message and sign-off your commit.

thanks!

guits (approved):

PR #2233, opened 2024-07-12
build squid release by default

guits:
These changes will make the project build Ceph Squid by default in addition to quincy, reef and main


adk3798 (approved):

guits:
sloppy bash and python jobs won't pass until ceph/ceph-build/pull/2267 is merged

PR #2238, opened 2024-09-16
squid: add gmonitoring package

guits:
nfs-ganesha 6 introduced a new RPM `gmonitoring`.
This commit includes this new package into the Ceph Squid container image.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -43,11 +43,7 @@ jobs:
         ceph_release: [quincy, reef, squid, main]
         exclude:
           - os: 9
-            ceph_release: squid
-          - os: 9
             ceph_release: quincy
-          - os: 9
-            ceph_release: reef
     runs-on: ubuntu-latest
     steps:
       - name: checkout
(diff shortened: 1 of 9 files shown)


ktdreyer:
I tested this and it will not impact ubi9, because ubi9 uses `src/daemon-base/__GANESHA_PACKAGES__`.

What if we delete `ceph-releases/squid/centos/__GANESHA_PACKAGES__` and just use `src/daemon-base/__GANESHA_PACKAGES__`?

It would simplify things to avoid hard-coding the ganesha version numbers in each individual package entry.

guits:
> I tested this and it will not impact ubi9, because ubi9 uses `src/daemon-base/__GANESHA_PACKAGES__`.

yeh, that was exactly what I suspected.

> What if we delete `ceph-releases/squid/centos/__GANESHA_PACKAGES__` and just use `src/daemon-base/__GANESHA_PACKAGES__`?
> 
> It would simplify things to avoid hard-coding the ganesha version numbers in each individual package entry.

That should do the job, indeed.

tserlin (approved):
LGTM

guits:
@ktdreyer is this package built for arm64 ?

guits:
@ktdreyer while working on this PR, I noticed we have no RPM built here: https://download.ceph.com/rpm-quincy/el9/aarch64/

ktdreyer:
> I noticed we have no RPM built here: https://download.ceph.com/rpm-quincy/el9/aarch64/

That's an interesting data point. I'm not involved with building those, and I think it means we don't support that upstream.

ktdreyer (approved):
I've checked the following:

```
BRANDING=redhat VERSION=7 ./contrib/compose-rhcs.sh
BRANDING=redhat VERSION=8 ./contrib/compose-rhcs.sh
BRANDING=ibm VERSION=8 ./contrib/compose-rhcs.sh
```

The output looks like I expect. Thanks!

Issue #2240, opened 2024-09-17
include libcephfs-daemon for IBM squid release


andrewschoen:
Include `libcephfs-daemon` in squid package install. 

The upstream work is still in process for the squid release: https://github.com/ceph/ceph/pull/58376

This work was already ported downstream for early testing: https://gitlab.cee.redhat.com/ceph/ceph/-/merge_requests/728

Can we please include `libcephfs-daemon` for IBM builds until the upstream work is merged?

PR #2241, opened 2024-09-17
squid/ubi9-ibm: install libcephfs-daemon

andrewschoen:
Fixes: #2240

### DIFF
--- a/ceph-releases/squid/ubi9-ibm/daemon/__CEPH_IBM_PACKAGES__
+++ b/ceph-releases/squid/ubi9-ibm/daemon/__CEPH_IBM_PACKAGES__
@@ -1,0 +1,1 @@
+ceph-mgr-callhome libcephfs-daemon


andrewschoen:
@guits I saw that `ceph-mgr-callhome` is included here already https://github.com/ceph/ceph-container/blob/main/ceph-releases/ALL/ubi9-ibm/daemon-base/__CEPH_IBM_PACKAGES__#L1

I wasn't sure if that would be inherited in the squid specific __CEPH_IBM_PACKAGES__ file though. Do I need to duplicate that install or can I just inlude `libcephs-daemon` here for squid?

phlogistonjohn:
This change is currently only for ibm ceph. Ideally, we will get it upstream too but it will take longer (AFAIK). So assume this is ibm only.

andrewschoen:
> @andrewschoen is this package supposed to be included only for ceph ibm? if yes, then that change looks good to me

yes, for now this is IBM only.

ktdreyer:
@phlogistonjohn there's a PR for this in upstream ceph, right? Could you link to that here?

andrewschoen:
> @phlogistonjohn there's a PR for this in upstream ceph, right? Could you link to that here?

@ktdreyer I included some info about the upstream work in the issue I created: https://github.com/ceph/ceph-container/issues/2240

I think this is the upstream PR that includes `libcephfs-daemon`: https://github.com/ceph/ceph/pull/58376

PR #2244, opened 2024-09-25
Revert ganesha6

guits:
The recent bump to nfs-ganesha v6 broke squid deployments.
Let's revert these changes until we figure out why it's failing.


ktdreyer:
CC @@kalebskeithley

If nfs-ganesha V6.0 is broken, we need to fix that.

Is there a log file we can pass to the nfs-ganesha team that shows the problem?

guits:
@ktdreyer it means quay.io/ceph/ceph:v19.2.0 is broken

travisn:
Rook CI seems to have hit this a couple days ago, see the details of the failure here: https://github.com/rook/rook/issues/14771

